PR #26, opened 2018-12-23
Pass pagination row to inventory so it can use pagination in future

karelhala:
No UI changes yet, can be merged before https://github.com/RedHatInsights/insights-frontend-components/pull/168 and once that PR is merged and new version is applied to chrome inventory will get automatically pagination.


karelhala:
ping @dLobatog

PR #64, opened 2019-02-05
Fixes RHICOMPL-34: Redirect to login on 401's

akofink:
IMO if a user is denied access to a given resource for RBAC reasons, we should be returning 403 rather than 401, as discussed [here](http://www.dirv.me/blog/2011/07/18/understanding-403-forbidden/index.html).


akofink:
I think 401 is the correct code here. If the user isn't authenticated properly (like if the cookie is missing or expired), it should be 401. 403 should be used for authorization issues, not authentication.

PR #81, opened 2019-02-15
Use tab from frontend-components

dLobatog:


dLobatog:
Merge after https://github.com/RedHatInsights/insights-frontend-components/pull/300 is put into a release, alongside a package.json change

dLobatog:
aaand it's merged! We should be able to merge this as soon as it's released

PR #91, opened 2019-03-04
Openshift scanning first wizard step

dLobatog:
We may have this in beta, step by step, and only promote to stable by cherrypicking, thoughts @akofink? This will be a very large change if I have to include everything at once:

screenshot from 2019-03-04 18-07-34
screenshot from 2019-03-05 09-51-46
screenshot from 2019-03-05 09-51-56
screenshot from 2019-03-05 09-52-07

@katierik

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9627,6 +9627,11 @@
         "next-tick": "1"
       }
     },
+    "es6-error": {
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/es6-error/-/es6-error-4.1.1.tgz",
+      "integrity": "sha512-Um/+FxMr9CISWh0bi5Zv0iOD+4cFh5qLeks1qhAopKVAJw3drgKbKySikp7wGhDL0HPeaja0P5ULZrxLkniUVg=="
+    },
     "es6-iterator": {
       "version": "2.0.3",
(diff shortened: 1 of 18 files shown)


katierik:
Empty state: https://redhat.invisionapp.com/d/main#/console/16876256/349896817/preview
Connecting: https://redhat.invisionapp.com/d/main#/console/16876256/349896816/preview

Right now in the mocks, the connection is configured via a modal. 

There is also a step to associate a policy/profile with the imagestreams.
[Comment away on the mocks!]

katierik:
Based on feedback from you, @akofink and Mohit, [the designs changed a bit](https://redhat.invisionapp.com/share/T4QSNP12WXK#/349896816_Add_An_Image_Stream). I'll let Mohit make the call on when that change should be built out. 

Re: the tab being visible to customers outside Red Hat: This worries me a bit, since we're not sure that we want to end up with a separate tab for image streams. 
Beyond that, the empty state should align closest with the ["not entitled" empty state](https://redhat.invisionapp.com/share/V8PLHRR6FAY#/355825641_Empty_State_-_Not_Entitled), but with some different copy. I don't see a screenshot of that here and since I've got a redhat email not sure how to see it - so not sure what is there now.

dLobatog:
@akofink according to PM's feedback on the Slack channel, I'll hide the tab for the time being

akofink:
In that case, there's not much danger in merging! :)

PR #92, opened 2019-03-07
Fixes RHICOMPL-102: Add widget-id to policy cards

akofink:
This aids front end test automation.

### DIFF
--- a/src/SmartComponents/CompliancePolicyCard/CompliancePolicyCard.js
+++ b/src/SmartComponents/CompliancePolicyCard/CompliancePolicyCard.js
@@ -71,7 +71,7 @@ class CompliancePolicyCard extends React.Component {
         );
 
         return (
-            <Card>
+            <Card widget-id={this.policy.ref_id}>
                 <CardBody>
                     <Text style={{ fontWeight: '500', color: '#bbb' }} component={TextVariants.small}>External Policy</Text>
                     <TextContent>


src/SmartComponents/CompliancePolicyCard/CompliancePolicyCard.js
```diff
@@ -71,7 +71,7 @@ class CompliancePolicyCard extends React.Component {
         );
 
         return (
-            <Card>
+            <Card widget-id={this.policy.id}>
```
quarckster: Thanks! That exactly what I asked!

quarckster: What kind of id is stored in `this.policy.id`?

akofink: It looks like this: `5ace0d15-178b-4624-93ea-306aec38f24e`. I think it's the UUID stored in our database.

quarckster: Hm, is it possible to put a value that would be related to name of the card?

akofink: The id needs to be unique, right? The policy name isn't unique here. Here's an example of what we have access to:

image

dLobatog: `ref_id` is unique in the db too, for each account

akofink: Here's an example of duplicate policies in the UI. I'm not sure if this is a bug or if it's intentional.

image

quarckster: For me looks like a bug

akofink: One of these systems is RHEL7; the other is CentOS7. Perhaps that's causing it?

akofink: @quarckster would you like me to use `ref_id`? I think that's a good idea.

quarckster: Probably, and I think it's quite important questions, should we handle centos reports at all?

akofink: The ref_id is different for the two policy cards shown in the screenshot above. rhel7 shows ref_id of xccdf_org.ssgproject.content_profile_standard. centos7 shows ref_id of standard.

akofink: No, we don't officially support centos or fedora, but they will still work, I believe.

akofink:
Updated to use ref_id!

PR #95, opened 2019-03-14
Fixes RHICOMPL-106: Toolbar should not be nested in the Table

akofink:
Before:

image

After:

image

### DIFF
--- a/src/SmartComponents/DownloadTableButton/DownloadTableButton.js
+++ b/src/SmartComponents/DownloadTableButton/DownloadTableButton.js
@@ -2,8 +2,7 @@ import React from 'react';
 import propTypes from 'prop-types';
 import { COMPLIANCE_API_ROOT } from '../../constants';
 import { connect } from 'react-redux';
-import { DownloadIcon } from '@patternfly/react-icons';
-import { Dropdown, DropdownToggle, DropdownItem, Tooltip } from '@patternfly/react-core';
+import { Dropdown, KebabToggle, DropdownItem, Tooltip } from '@patternfly/react-core';
 
 class DownloadTableButton extends React.Component {
     constructor(props) {
@@ -41,17 +40,17 @@ class DownloadTableButton extends React.Component {
         const { isOpen } = this.state;
         const dropdownItems = [
             <DropdownItem key='csv' href={this.downloadLink('csv')}>
-                CSV
+                Export as CSV
             </DropdownItem>,
             <DropdownItem target='_blank' key='json' href={this.downloadLink('json')}>
-                JSON
+                Export as JSON
             </DropdownItem>
         ];
+
         return (
             <React.Fragment>
                 <Tooltip
                     enableFlip={ true }
-                    position="left"
                     content={
                         <div translate>
                             More actions
@@ -60,7 +59,10 @@ class DownloadTableButton extends React.Component {
                 >
                     <Dropdown
                         onSelect={this.onSelect}
-                        toggle={<DropdownToggle onToggle={this.onToggle}><DownloadIcon/></DropdownToggle>}
+                        toggle={<KebabToggle
+                            onToggle={this.onToggle}
+                            style={{ color: 'var(--pf-global--icon--Color--light)' }}
+                        />}
                         isOpen={isOpen}
                         dropdownItems={dropdownItems}
                     />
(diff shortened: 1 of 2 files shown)


quarckster:
@dLobatog you're right, it's not exactly what I asked. 
screen

katierik:
I notated the correct patternfly spacers for the toolbar below. Also please note that the export has been moved into a kebab (UX cross-platform decision) and it is to the right of the "Remediate with Ansible" button. 
The design also include the filters, but the space on the far L&R should be "pf-global-spacer-xl" either way.

Xnip2019-03-19_09-26-05

akofink:
I'm not sure I can modify all the spacing as requested (or I'm not sure how to proceed with the components given). Here's what I have right now. I know it's not exactly what you showed, @katierik, but the table you show is for rules, not systems. idk what the filter is supposed to do, and I haven't figured out the kebab yet.

image

katierik:
For the right side and the item count, the pagination component should be able to handle it 
     See the responsive behavior in the demo
     https://v2.patternfly.org/documentation/core/components/pagination

Filter should be the "Faceted filter" that's documented here: 
     https://docs.google.com/document/d/1G3r9dJFQtLxc5it4FfM6Jj6PlXtzGcAVndfqGPGvWvE/edit?usp=sharing (not on the website yet :/)

Xnip2019-03-20_09-38-52

akofink:
Here's where I'm at now. Looks like the filter components aren't available yet in pf-react:

image

dLobatog:
@akofink I think we're waiting on insights-frontend-components for the filtering components, @karelhala that was sometime ago in the toolbar, is it expected to come back before summit?

karelhala:
@dLobatog yes, you are correct. Filtering was removed from inventory table because we are not able to do multi app cross filtering as of yet, so that won't be part of the basic inventory table for some time. However there is one component made in Advisor that is doing esentialy the same thing [1], I heard from others that they'd like to use this component as well so we should try to move that component out of Advisor repo and provide it to others. As of for now I would recommend you to merge this even without this filering possible and wait for either Allen or someone else to move it and you can use it afterwards.

[1] https://github.com/RedHatInsights/insights-advisor-frontend/pull/180

karelhala:
> This removes TableToolbar though, isn't this something supposed to be the standard component for this kind of toolbar over the inventory table 

Table toolbar is used inside inventory table directly. No need to use it once more, just pass your stuff to inventory and it will be rendered in tableToolbar.

src/SmartComponents/DownloadTableButton/DownloadTableButton.js
```diff
@@ -41,10 +40,10 @@ class DownloadTableButton extends React.Component {
         const { isOpen } = this.state;
         const dropdownItems = [
             <DropdownItem key='csv' href={this.downloadLink('csv')}>
-                CSV
```
karelhala: Consider using download button instead of custom button [RedHatInsights/insights-frontend-components/tree/master/src/PresentationalComponents/DownloadButton](https://github.com/RedHatInsights/insights-frontend-components/tree/master/src/PresentationalComponents/DownloadButton)

akofink: @katierik Is this ok for the download button? If not, @karelhala there doesn't seem to be a way to pass in a custom download button icon, is there?

image

 The mockup provided uses the pf kebab, like this:

image

akofink: Same for the text for CSV and JSON. They don't match the mockup:

image

Mockup:

image

akofink: I attempted to add some more customization to the `DownloadButton` but didn't get very far. Overriding all props like that probably isn't the way to do it.

akofink:
Updated with better spacing. Here's how it looks now:

image

src/SmartComponents/DownloadTableButton/DownloadTableButton.js
katierik: As per UXD conversations, we're going to use the kebab. This hasn't been implemented (or I guess reported) to the inventory group yet.

katierik:
Xnip2019-03-29_08-49-55

A few comments in the image!

For the kebab icon in the toolbar, use the light icon color: "global_icon_Color_light", "--pf-global--icon--Color--light"

quarckster:
> Xnip2019-03-29_08-49-55
> 
> A few comments in the image!
> 
> For the kebab icon in the toolbar, use the light icon color: "global_icon_Color_light", "--pf-global--icon--Color--light"

@katierik you're doing awesome work with finding UI issues. Let's create separate issues for all them to be consistent.

akofink:
@katierik thanks for the feedback! The only thing I can fix here is the kebab color. The rest of the issues are coming from the inventory table and must be fixed there.

akofink:
Updated with kebab color:

image

PR #96, opened 2019-03-14
Update routing to match new version of cloud

karelhala:
You might gonna have to run `npm run build` if you see some errors with static files from `insights/plaform`

### DIFF
--- a/config/spandx.config.js
+++ b/config/spandx.config.js
@@ -1,0 +1,7 @@
+/* global exports */
+
+exports.routes = {
+    '/rhcs/compliance': { host: 'http://localhost:8002' },
+    '/apps/compliance': { host: 'http://localhost:8002' },
+    '/apps/chrome': { host: 'https://ci.cloud.paas.upshift.redhat.com' }
+};
--- a/config/spandx.config.js
+++ b/config/spandx.config.js
@@ -3,6 +3,8 @@
 // Hack so that Mac OSX docker can sub in host.docker.internal instead of localhost
 // see https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
 const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
+const host = 'https://insights-inventory-platform-ci.5a9f.insights-dev.openshiftapps.com';
+
 module.exports = {
     routes: {
         '/apps/compliance': { host: `http://${localhost}:8002` },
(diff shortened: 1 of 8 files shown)


karelhala:
ping @dLobatog

config/spandx.config.js
```diff
@@ -0,0 +1,7 @@
+/* global exports */
+
+exports.routes = {
+    '/rhcs/compliance': { host: 'http://localhost:8002' },
+    '/apps/compliance': { host: 'http://localhost:8002' },
```
akofink: What about rails on port 3000?

karelhala: You can extend this config as much as you want. This is just initial settings to make your local changes apear after cloud is up and running.

akofink: Ok, so what is supposed to be running on port 8002?

karelhala: Webpack dev server, when running local build [config/dev.webpack.config.js#L11](https://github.com/RedHatInsights/compliance-frontend/blob/master/config/dev.webpack.config.js#L11)

config/spandx.config.js
```diff
@@ -0,0 +1,13 @@
+/*global module, process*/
+
+// Hack so that Mac OSX docker can sub in host.docker.internal instead of localhost
+// see https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
+module.exports = {
+    routes: {
+        '/apps/compliance': { host: `http://${localhost}:8002` },
+        '/rhcs/compliance': { host: `http://${localhost}:8002` },
```
dLobatog: If you don't mind could you add

```
         '/api/compliance': { host: `http://${localhost}:3000` }, 
```

so that our backend works right away? Thanks!

config/spandx.config.js
```diff
@@ -0,0 +1,13 @@
+/*global module, process*/
+
+// Hack so that Mac OSX docker can sub in host.docker.internal instead of localhost
+// see https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
+module.exports = {
+    routes: {
```
dLobatog: If you don't mind could you add

```
         '/api/compliance': { host: `http://${localhost}:3000` }, 
```

so that our backend works right away? Thanks!

karelhala: Sure thing!

karelhala:
>  but I couldn't test the pages where the Inventory component is used

This is tricky to do because API is not yet moved to cloud services. I will add path to inventory CI env to allow at least some testing.

PR #97, opened 2019-03-19
Updates to handle cloud.redhat.com

dLobatog:
I am unsure of this, but I think we should be also changing where the backend app is listening for requests in order to be able to merge this:

https://github.com/RedHatInsights/compliance-backend/blob/master/config/routes.rb#L4

### DIFF
--- a/src/entry.js
+++ b/src/entry.js
@@ -19,7 +19,7 @@ const client = new ApolloClient({
 window.insights.chrome.auth.getUser().then(() => {
     ReactDOM.render(
         <Provider store={ init().getStore() }>
-            <Router basename={ `/${RELEASE}/platform/compliance` }>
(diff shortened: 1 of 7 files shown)


src/entry.js
```diff
@@ -19,7 +19,7 @@ const client = new ApolloClient({
 window.insights.chrome.auth.getUser().then(() => {
     ReactDOM.render(
         <Provider store={ init().getStore() }>
-            <Router basename={ `/${RELEASE}/platform/compliance` }>
+            <Router basename={ `${pathName[1] === 'beta' ? pathName[2] : pathName[1]}/compliance` }>
```
dLobatog: Shouldn't we expose `pathname` instead of `RELEASE` here?

PR #110, opened 2019-04-01
RHICOMPL-125 - Fix Compliance Score column

dLobatog:
This PR is meant to merge the current "score" and "compliant" columns into "Compliance score" as per https://redhat.invisionapp.com/share/V8PLHRR6FAY#/screens/336685940_Compliance_2

Screenshot from 2019-04-01 11-04-37
Screenshot from 2019-04-01 11-04-24

@katierik


katierik:
Looks good!

I see issues with the rules failed link missing on the bottom, and then sentence case in the column headers (assuming these are being handled separately, but mentioning it just in case)

dLobatog:
@akofink The rules failed link seems to be a bug, as I see that sometimes, and sometimes it just disappears...

PR #111, opened 2019-04-01
RHICOMPL-125 - Add relative time to 'last scanned' column

dLobatog:
Screenshot from 2019-04-01 11-33-14
Screenshot from 2019-04-01 11-29-54

cc @katierik

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10713,12 +10713,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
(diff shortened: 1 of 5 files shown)


akofink:
This also needs to be on the system detail policies card:

image

dLobatog:
@akofink I'll address both here - thanks for the review!

dLobatog:
Or.. actually just the links issue. The other one is part of the inventory in insights-frontend-components.

PR #112, opened 2019-04-01
RHICOMPL-125 - Fix compliance threshold format

dLobatog:
Screenshot from 2019-04-01 12-24-19
Screenshot from 2019-04-01 12-23-12

Things that changed - the icon is now the outlined circle instead of the filled one, the tooltip is square rather than long, and there's some space below the number itself. 

cc @katierik


src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -176,30 +177,28 @@ const PolicyDetailsQuery = ({ policyId, onNavigateWithProps }) => (
...
+                                                each policy.
+                                                This defines the percentage of passed rules that must be met in order
+                                                for a system to be determined &quot;compliant&quot;.
+                                            </div>
+                                        }
+                                    >
+                                        <Text style={{ fontWeight: 'bold' }} component={ TextVariants.p }>
+                                            Minimum threshold for compliance <OutlinedQuestionCircleIcon/>
+                                        </Text>
+                                        <Text id='threshold-tooltip' component={TextVariants.p}>
```
akofink: I've been putting this kind of one-line style change inline. Would you prefer ids + scss files? I don't really have a preference or know the convention.

It's not clear, but with react, it's like this:

```jsx
<Text style={{ marginBottom: 'var(--pf-global--spacer--md)' }} component={TextVariants.p}>
```

src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -176,30 +177,28 @@ const PolicyDetailsQuery = ({ policyId, onNavigateWithProps }) => (
                             <GridItem span={6}>
                                 <TextContent>
                                     <Text style={{ fontWeight: 'bold' }} component={TextVariants.p}>Description</Text>
-                                    <Text className="policy-description" component={TextVariants.p}>
```
akofink: Is @quarckster using this class for testing?

src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -176,30 +177,28 @@ const PolicyDetailsQuery = ({ policyId, onNavigateWithProps }) => (
...
+                                            <div>
+                                                The threshold for compliance is a value set by your organization for
+                                                each policy.
+                                                This defines the percentage of passed rules that must be met in order
+                                                for a system to be determined &quot;compliant&quot;.
+                                            </div>
+                                        }
+                                    >
+                                        <Text style={{ fontWeight: 'bold' }} component={ TextVariants.p }>
+                                            Minimum threshold for compliance <OutlinedQuestionCircleIcon/>
```
akofink: The icon should be gray. Looks like it's `var(--pf-global--Color--200)`

Mockup:

image

src/SmartComponents/PolicyDetails/PolicyDetails.js
dLobatog: I would say class + scss is definitely preferred, it promotes DRY and it's easier to test. There are many instances where we just use inline styles. Initially I just used them to "mark" these instances as 'hacks' but it's probably better to explain the purpose of a CSS change in the SCSS file if it's supposed to be a hacky workaround.

src/SmartComponents/PolicyDetails/PolicyDetails.js
quarckster: I use this `<span widget-type="InsightsTruncateBlock">` in current version it's nested in this element.

PR #115, opened 2019-04-01
Remove invalid DOM nesting inside a <p>

akofink:
If we want to use `TextVariant.blockquote` or `<div>`, we cannot put it inside an `<EmptyStateBody>`, as it generates the following warning because [it's essentially a `<p>` tag](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/EmptyState/EmptyStateBody.js#L7):

```
Warning: validateDOMNesting(...): <blockquote> cannot appear as a descendant of <p>.
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
```

This does not change how the empty state page looks. It merely replaces the `<p>` with a `<span>`

### DIFF
--- a/src/SmartComponents/CompliancePoliciesCards/CompliancePoliciesCards.js
+++ b/src/SmartComponents/CompliancePoliciesCards/CompliancePoliciesCards.js
@@ -9,9 +9,9 @@ import {
     Button,
     Bullseye,
     EmptyState,
-    EmptyStateIcon,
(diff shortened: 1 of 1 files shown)


akofink:
Disclaimer: this is only _one_ way to fix this. I think there's a more HTML way to have
`<p><blockquote></blockquote></p>`

PR #117, opened 2019-04-05
Request specific columns for CSV export

dLobatog:
This depends on https://github.com/RedHatInsights/compliance-backend/pull/103


akofink:
@dLobatog Just a thought (we're probably too far in at this point, but...) @johnsonm325 exports directly from the redux state. This seems like a pretty good idea to me - it guarantees the user gets the data they see in the UI, and it saves a trip to the backend for the export. When you first implemented export, did you consider doing it that way?

dLobatog:
@akofink I didn't consider that, but it definitely looks to me like the superior way. Perhaps we can move to that in some time, this PR is just meant to address the current approach.

dLobatog:
Irrelevant after https://github.com/RedHatInsights/compliance-frontend/commit/f2747e7c073b08bbe6227f15c21ad75336f3511a

PR #121, opened 2019-04-08
InventoryTable state needs items, not rows

akofink:
Fixes https://projects.engineering.redhat.com/browse/RHICOMPL-137

This took _way_ too long to figure out.


dLobatog:
@akofink  I need to double check the versions but using the code in this PR I see the following things:
* state.items is ignored. 
* state.rows contains all of the hosts I have in the inventory, no compliance data (that's fine)
* Since state.rows is what the table uses to render the rows apparently, the profiles/compliant columns show up empty.

PR #122, opened 2019-04-10
Export from redux store

akofink:


src/store/Reducers/SystemStore.js
```diff
@@ -79,6 +84,40 @@ export const systemsToInventoryEntities = (systems, entities) =>
...
+
+const csvFromState = (state) => {
+    if (state.rows) {
+        const CELL_DELIMITER = ',';
+        let csvRows = [state.columns.map((column) => column.title).join(CELL_DELIMITER)];
+        csvRows = csvRows.concat(state.rows.map((row) => {
+            return state.columns.map((column) => {
+                let cell = get(row, column.key);
+                if (typeof(cell) === 'object') {
+                    cell = get(row, column.key + 'Text');
```
akofink: So, to make this function more reusable, I've added this convention here that if a column is an object, try to get the column key with 'Text' appended. i.e. `facts.compliance.compliant` is a react component object, so we look for the plain-text version at `facts.compliance.compliantText`. This is fairly trivial to add to a reducer, as I've done on line 78.

src/store/Reducers/SystemStore.js
```diff
@@ -69,16 +77,54 @@ export const systemsToInventoryEntities = (systems, entities) =>
...
+
+const csvFromState = (state) => {
+    if (state.rows) {
+        const CELL_DELIMITER = ',';
+        let csvRows = [state.columns.map((column) => column.title).join(CELL_DELIMITER)];
+        csvRows = csvRows.concat(state.rows.map((row) => {
+            return state.columns.map((column) => {
+                let cell = get(row, column.key);
+                if (typeof(cell) === 'object') {
+                    cell = get(row, column.key + '_text');
```
akofink: So, to make this function more reusable, I've added this convention here that if a column is an object, try to get the column key with '_text' appended. i.e. `facts.compliance.compliant` is a react component object, so we look for the plain-text version at `facts.compliance.compliant_text`. This is fairly trivial to add to a reducer, as I've done on line 84-86.

src/store/Reducers/SystemStore.js
```diff
@@ -69,16 +77,54 @@ export const systemsToInventoryEntities = (systems, entities) =>
...
                             }
                         }}>{system.rules_failed}</Link>,
+                        rules_failed_text: system.rules_failed,
                         compliance_score: complianceScore(system),
+                        compliance_score_text: complianceScoreString(system),
                         compliant: compliantIcon(system),
-                        last_scanned: <FormattedRelative value={Date.parse(system.last_scanned)} />
+                        compliant_text: system.compliant.toString(),
+                        last_scanned: <FormattedRelative value={Date.parse(system.last_scanned)} />,
+                        last_scanned_text: system.last_scanned
```
akofink: I know the original bug was that the export didn't look exactly like the html table; however, we definitely want parts of it to be different, like here: a relative date in a csv is pretty much useless IMO. Best to use the full date/time format with timezone to be clear.

src/store/Reducers/SystemStore.js
```diff
@@ -69,16 +77,54 @@ export const systemsToInventoryEntities = (systems, entities) =>
...
                     }
                 }
                 /* eslint-enable camelcase */
             };
         }
     ).filter(value => value !== undefined);
 
+const linkAndDownload = (data, filename) => {
```
dLobatog: Ideally all of these methods would live in `src/Utilities/CsvExport` or somewhere else in `src/Utilities`

src/store/Reducers/SystemStore.js
```diff
@@ -69,16 +77,54 @@ export const systemsToInventoryEntities = (systems, entities) =>
...
+};
+
+const csvFromState = (state) => {
+    if (state.rows) {
+        const CELL_DELIMITER = ',';
+        let csvRows = [state.columns.map((column) => column.title).join(CELL_DELIMITER)];
+        csvRows = csvRows.concat(state.rows.map((row) => {
+            return state.columns.map((column) => {
+                let cell = get(row, column.key);
+                if (typeof(cell) === 'object') {
```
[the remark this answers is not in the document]
akofink: It looks like it's pretty standard practice to wrap those fields in quotes: `foo, bar, "PCI for RHEL 7, Standard for RHEL 7", baz`. When I open such a csv in libreoffice, it works with default import settings. What do you think? I'd prefer this to changing the field data itself.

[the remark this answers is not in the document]
dLobatog: That sounds good, anything that makes it clear that's a column would work

akofink:
Updated - also, we need to handle JSON too :crying_cat_face:

PR #133, opened 2019-05-03
Fix all systems pagination to show all systems instead of limited amount

karelhala:
### Description
When fetching all systems we have to get amount of systems in DB by getting host count in all profiles. Laer we have to fetch current page from graphql and once that is done call host fetch to get inventory info. Once inventory fetches these data we have to map our specific information to it as well.

### Changes
* Call graphql with `allProfiles` query and take `total_host_count`. Once that is fullfilled pass number of systems to Systems table as total
* Fetch `name`, `profile_names` and `compliant` values for each page in systems
* Pass callback to systems to reducer so we can access previously fetched systems (from compliance DB)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11003,12 +11003,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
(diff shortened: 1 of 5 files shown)


karelhala:
ping @dLobatog

PR #134, opened 2019-05-05
Move fetching profileDetails systems to systemsTable component

dLobatog:
This allows us to easily fetch systems on a per_page basis on the
ProfileDetails page, following the same logic as we follow in the
Systems page


dLobatog:
This should allow us to display profiles where you may have 900 hosts. Before this PR it tries to fetch all 900 compliance/rules_passed/failed from the beginning, and it won't even load the page.

PR #144, opened 2019-05-15
Use new frontend components monerepo

karelhala:
This PR changes imports of all things from insights frontend components to new version of monorepo components. https://github.com/RedHatInsights/frontend-components


karelhala:
We are planning to copy files between old and new repository for some time, so no need to rush with this PR. But in few days we are going to close old repository and use only the new one.

karelhala:
I checked the bundle size, because main reason for this is to reduce the size of bundles and it reduced the size from 6.5MB to 3.2MB and build time was reduced from 4 minutes to 2 minutes.

akofink:
Is there any danger in going ahead and merging this? Why still WIP?

karelhala:
With the recent changes it does not need to be WIP anymore. We are still pushing to ci beta on master merges so it's good to go. We have enabled travis builds in the monorepo now and we will wait for auto releasing once most apps sart to use it so we don't release versions that could be broken.

PR #145, opened 2019-05-15
Update spandx to match the sample app example

akofink:
Modeled after https://github.com/RedHatInsights/insights-frontend-starter-app/blob/master/profiles/local-frontend.js


config/spandx.config.js
```diff
@@ -1,17 +1,17 @@
...
-    routes: {
-        '/apps/compliance': { host: `http://${localhost}:8002` },
-        '/rhcs/compliance': { host: `http://${localhost}:8002` },
-        '/r/insights/platform': { host: 'http://access.ci.cloud.paas.upshift.redhat.com' },
-        '/apps/chrome': { host: 'https://ci.cloud.paas.upshift.redhat.com' },
-        '/api/compliance': { host: `http://${localhost}:3000` },
-        '/r/insights/platform/inventory/': { host } // for CI env
-    }
-};
+
```
dLobatog: Apparently eslint is making the PR fail because of this line

PR #146, opened 2019-05-20
Request for all host remediations at once

dLobatog:
Before this change, to display the remediations modal we had
to fetch all remediations available per system. Now we request
all remediations using only one request.

### DIFF
--- a/src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
+++ b/src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
@@ -34,12 +34,10 @@ class ComplianceRemediationButton extends React.Component {
     }
 
     /* eslint-disable camelcase */
-    formatRule = ({ title, ref_id }, profile, system) => ({
+    formatRule = ({ title, ref_id }, profile, systems) => ({
         id: `ssg:rhel7|${profile}|${ref_id}`,
         description: title,
-        systems: [
-            system
-        ]
+        systems
     })
 
     findRule = (rules, ref_id) => {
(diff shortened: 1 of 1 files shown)


src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -77,23 +75,38 @@ class ComplianceRemediationButton extends React.Component {
...
-            }
-        });
+
+        allSystems.forEach(system => { result.systems.push(system.id); });
+
+        if (selectedRules) {
+            result.issues.push(selectedRules.map(rule => this.formatRule(rule, [allSystems[0].id])));
+        } else {
+            const rules = flatten(allSystems.map(system => system.rule_objects_failed));
+            const rulesPerSystem = rules.reduce((acc, rule) => {
```
dLobatog: This reduce function is basically the main difference. Before, I was iterating through all systems and requesting remediations to tell me which `rule_objects_failed` have a valid remediation. In this function we make a mapping where we can find the system IDs applicable for each rule. 
Later on we use it to query the system IDs like `rulePerSystem[rule_ref_id]` and that returns the system IDs.

src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -34,12 +34,10 @@ class ComplianceRemediationButton extends React.Component {
...
 
     /* eslint-disable camelcase */
-    formatRule = ({ title, ref_id }, profile, system) => ({
+    formatRule = ({ title, ref_id }, profile, systems) => ({
         id: `ssg:rhel7|${profile}|${ref_id}`,
         description: title,
-        systems: [
-            system
-        ]
+        systems
```
dLobatog: Before, we only handled formatting for 1 system at a time.

akofink:
Are you sure this is working as expected? I'm still seeing two requests (with two systems) to `remediations/v1/resolutions` when I click 'Remediate with Ansible':

image

PR #156, opened 2019-06-19
Add docker compose setup for development

bastilian:
This adds a Docker image and docker compose setup to eventually only have to use `docker-compose build` & `docker-compose up` to get the frontend and insights-proxy up and running for development.

This is based on the setup used for https://github.com/RedHatInsights/notifications-frontend/.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,15 @@
 FROM fedora:29
 RUN dnf install -y npm git
 
-COPY . /frontend
+RUN mkdir /frontend
 WORKDIR /frontend
 
 COPY package*.json /frontend/
 RUN npm install
 
-CMD ["/usr/bin/npm", "run", "start"]
+COPY ./entrypoint.sh /
+ENTRYPOINT ["/entrypoint.sh"]
+
+CMD ["npm", "run", "start"]
 
 EXPOSE 8002
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM fedora:29
+RUN dnf install -y npm git
+
+COPY . /frontend
+WORKDIR /frontend
+
+RUN rm -rf node_modules/*
+
+RUN npm install
+
+CMD ["/usr/bin/npm", "run", "start"]
+
+EXPOSE 8002
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,7 @@ RUN dnf install -y npm git
 COPY . /frontend
 WORKDIR /frontend
 
-RUN rm -rf node_modules/*
-
+COPY package*.json /frontend/
 RUN npm install
 
 CMD ["/usr/bin/npm", "run", "start"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,7 +22,7 @@ services:
           - PLATFORM=linux
           - PORT=8002
           # Comment out to use a local backend
-          - LOCAL_API=false
+          # - LOCAL_API=false
           # - LOCAL_API_SCHEME=http
           # - LOCAL_API_PORT=3000
           - UI_HOST=frontend
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,38 +1,18 @@
 version: '3'
 services:
     frontend:
+        env_file: .env
         build: .
         ports:
             - 8002:8002
-        environment:
-            - HOST=0.0.0.0
         volumes:
             - .:/frontend:z
-            - node_modules:/frontend/node_modules
+            - /frontend/node_modules
 
     proxy:
+        env_file: .env
         image: docker.io/redhatinsights/insights-proxy
         ports:
-          - "1337:1337"
-        depends_on:
-          - frontend
-        links:
-          - frontend
-        environment:
-          - PLATFORM=linux
-          - PORT=8002
-          # Comment out to use a local backend
-          # - LOCAL_API=false
-          # - LOCAL_API_SCHEME=http
-          # - LOCAL_API_PORT=3000
-          - UI_HOST=frontend
+          - 1337:1337
         volumes:
           - ./config/spandx.config.js:/config/spandx.config.js:z
-
-networks:
-    default:
-        external:
-            name: docker_default # should match insights-upload's network name
-
-volumes:
-    node_modules:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,34 +2,8 @@ version: '3'
 services:
   frontend:
     build: .
-    network_mode: "host"
     ports:
       - "8002:8002"
     volumes:
       - .:/frontend/:z
       - /frontend/node_modules/
-    extra_hosts:
-      - prod.foo.redhat.com:127.0.0.1
-      - stage.foo.redhat.com:127.0.0.1
-      - qa.foo.redhat.com:127.0.0.1
-      - ci.foo.redhat.com:127.0.0.1
-  proxy:
-    image: docker.io/redhatinsights/insights-proxy
-    network_mode: "host"
-    ports:
-      - "1337:1337"
-    extra_hosts:
-      - prod.foo.redhat.com:127.0.0.1
-      - stage.foo.redhat.com:127.0.0.1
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM fedora:29
```
akofink: Why not use the nodejs ubi?

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN dnf install -y npm git
+
+COPY . /frontend
+WORKDIR /frontend
+
+RUN rm -rf node_modules/*
+
+RUN npm install
```
akofink: IMO this should copy package.json and package-lock.json right before running npm install to take advantage of incremental image builds. There's no reason to copy the whole repo in, as we'll most likely want to volume mount the repo in development anyway.

docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      - "1337:1337"
+    extra_hosts:
+      - prod.foo.redhat.com:127.0.0.1
+      - stage.foo.redhat.com:127.0.0.1
+      - qa.foo.redhat.com:127.0.0.1
+      - ci.foo.redhat.com:127.0.0.1
+    environment:
+      - PLATFORM=linux
+      - PORT=8002
+      - LOCAL_CHROME=true
```
akofink: :thinking: I don't understand how this works without insights-chrome being cloned somewhere

docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      - LOCAL_API=true
+      - LOCAL_API_SCHEME=http
+      - LOCAL_API_PORT=3000
+    volumes:
+      - ${LOCAL_CHROME_PATH}:/chrome:Z
```
akofink: I think we want `:z` instead of `:Z`. `:z` allows us to mount this directory in multiple different containers at the same time.

Dockerfile
bastilian: I tried ubi, but ran into some issue with npm and permissions. I'll keep investigating to make it work.

Dockerfile
bastilian: Done. Only package.json and package-lock.json are coppied in.

docker-compose.yml
bastilian: Yeah, it is needed to have insight-chrome cloned and built somewhere. This is helpful when chroming inside the insights-proxy is outdated, which might be less the case now. I found it helpful also in cases where you need to hack the chroming. With `notifications` which, that now is `hooks`, we needed to run a modified version of chroming, since we had no place in the navigation yet.

Dockerfile
akofink: If it's too hard to figure out, we can always change it in the future. This is just for development after all.

docker-compose.yml
akofink: Could we make it configurable with a .env variable? i.e. `LOCAL_CHROME=${LOCAL_CHROME}`

Dockerfile
bastilian: I left it `fedora` for now, but we can revist it later.

bastilian:
We could mount a directory that contains packages that should be linked and runs `npm link` for each of them in an entrypoint script.

I'll look into that and see that this is supported.

bastilian:
atm. hot loading needs to be disabled for the UI to load correctly.

The issue seems to be that the proxy runs with HTTPS and the websocket connection isn't SSL. This could be fixed by mounting the proxies SSL certificates and configure hot loading with SSL and certificates.

akofink:
I still don't really understand how I am supposed to `npm link ../frontend-components/packages/*` with this docker environment.

bastilian:
@akofink the idea with the entrypoint is that there will be a LINKED env variable, which will be used in the volumes definition to mount a directory that contains the packages that should be linked. When the container starts the entrypoint will check the mounted directory and run `npm link` for each package. That would require the packages to all be in that directory. Would that work?

akofink:
I think that's fine. We also need to run `npm run watch` in the frontend-components directory after we've linked the packages. It would be helpful to update the Readme with full instructions on how to get setup with all of this.

akofink:
@bastilian there's no risk in merging this. I'd rather get it merged and used than sit here forever.

bastilian:
@akofink It's rebased now upon latest master and made sure everything is working as described.

There is still an issue with running tests within the container. For some reason it can not find the tests. I suspect it is permission issues, which I tried to fix, but nothing helped and it still threw:

```
[root@b3de3ab0fb60 frontend]# npm test

> compliance-frontend@1.1.0 test /frontend
> jest --passWithNoTests

events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: spawn find ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
    at Function.Module.runMain (internal/modules/cjs/loader.js:834:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
Emitted 'error' event at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:246:12)
    at onErrorNT (internal/child_process.js:415:16)
    [... lines matching original stack trace ...]
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
npm ERR! Test failed.  See above for more details.
[root@b3de3ab0fb60 frontend]#
```

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '3'
+services:
+  frontend:
+    build: .
+    ports:
+      - 8002:8002
+    volumes:
+      - .:/frontend
+      - node_modules:/frontend/node_modules
```
akofink: Maybe the issue is here? With selinux enforcing, I think we need to include the `:z` at the end.

bastilian: I tried that, it still had issues. :/ It can be a local issue as well, so if it works with `:z` for you we can add it and I'll try to figure out what is going on on my machine.

akofink: It's also not working for me. It seems weird to mount node_modules, as it's already mounted with the cwd. Removing the second volume mount doesn't solve the test problem, but it does allow me to run `npm ci` on the container without errors.

bastilian: I'll try to figure this out over the weekend.

The reason for putting `node_modules` into a volume is to avoid issues with binaries for development environments where the Docker host is not the same architecture as the developing machine. For example docker on macOS runs within a linux VM.

akofink: Are binaries stored in `node_modules`? I thought it was all JS.

akofink: Also, I'm personally not super interested in supporting development on top of OS X ;)

akofink:
Closing this as stale. We can reopen if anyone chooses to revisit it.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+
+networks:
+  default:
+    external:
+      name: docker_default # should match insights-upload's network name
+
```
akofink, line 15: We probably don't need this - the frontend doesn't need direct access to any backend service, right?

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+services:
+    frontend:
+        build: .
+        ports:
+            - 8002:8002
+        environment:
+            - HOST=0.0.0.0
+        volumes:
+            - .:/frontend:z
+            - node_modules:/frontend/node_modules
```
akofink, line 11: We should remove this volume - it's part of the volume on line 10, right?

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+          # - LOCAL_API_SCHEME=http
+          # - LOCAL_API_PORT=3000
+          - UI_HOST=frontend
+        volumes:
+          - ./config/spandx.config.js:/config/spandx.config.js:z
+
+networks:
+    default:
+        external:
+            name: docker_default # should match insights-upload's network name
```
akofink, line 35: We can remove this as well

PR #160, opened 2019-07-02
RHICOMPL-145 - Commands to copy should have white background

dLobatog:
Screenshot from 2019-07-02 18-37-36

The reason why the TextInput was grey is that disabled text inputs are grey by default, the color comes from the system, not from Patternfly. If we set the `isReadOnly` option, the text input to copy becomes grey instead of white.

### DIFF
--- a/src/SmartComponents/CompliancePoliciesCards/CompliancePoliciesCards.js
+++ b/src/SmartComponents/CompliancePoliciesCards/CompliancePoliciesCards.js
@@ -101,12 +101,14 @@ const CompliancePoliciesCards = () => (
                                 about your system&#39;s compliance to policies.
                                 <br/>
                                 Generate a report with OpenSCAP with the following command:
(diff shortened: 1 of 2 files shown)


src/SmartComponents/CompliancePoliciesCards/CompliancePoliciesCards.js
```diff
@@ -101,12 +101,14 @@ const CompliancePoliciesCards = () => (
                                 about your system&#39;s compliance to policies.
                                 <br/>
                                 Generate a report with OpenSCAP with the following command:
-                                <ClipboardCopy isReadOnly variant={ClipboardCopyVariant.expansion}>
+                                <ClipboardCopy style={{ background: 'white' }}
```
akofink: Should we make a css class?

dLobatog: Good point, updated

PR #163, opened 2019-07-09
RHICOMPL-133 - Update notifications when a new remediation is created

dLobatog:
This also includes adaptions to ensure the new GraphQL API works with the remediate with Ansible button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@patternfly/react-table": "2.1.6",
     "@redhat-cloud-services/frontend-components": "0.0.5",
     "@redhat-cloud-services/frontend-components-inventory-compliance": "0.0.10",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "@patternfly/react-table": "2.1.6",
     "@redhat-cloud-services/frontend-components": "0.0.5",
     "@redhat-cloud-services/frontend-components-inventory-compliance": "0.0.9",
-    "@redhat-cloud-services/frontend-components-notifications": "0.0.3",
+    "@redhat-cloud-services/frontend-components-notifications": "0.0.6",
```
dLobatog: This is really the key - 0.0.3 does not display any notifications currently, not sure why

PR #171, opened 2019-07-19
RHICOMPL-250 - Select business objective should not show vertical scroll

dLobatog:
Screenshot from 2019-07-19 11-17-28

In addition to the main change, I'm using a version of react-select here that's meant to be integrated with pf4, and updating the react-core.


akofink:
Agreed - eventually it will need to scroll if there are many BOs or a very tiny screen.

PR #172, opened 2019-07-19
RHICOMPL 251 - Create business objective automatically

dLobatog:
Do NOT merge yet.

### DIFF
--- a/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
+++ b/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
@@ -6,23 +6,13 @@ import { ReduxFormCreatableSelectInput } from '../ReduxFormWrappers/ReduxFormWra
 import gql from 'graphql-tag';
 import { compose, withApollo } from 'react-apollo';
 import { connect } from 'react-redux';
+import debounce from 'lodash/debounce';
 
 const GET_BUSINESS_OBJECTIVES = gql`
 query businessObjectives {
     businessObjectives {
         id
         title
-    }
-}
-`;
-
-const CREATE_BUSINESS_OBJECTIVE = gql`
-mutation createBusinessObjective($input: createBusinessObjectiveInput!) {
-    createBusinessObjective(input: $input) {
-        businessObjective {
-            id
-            title
-        }
     }
 }
 `;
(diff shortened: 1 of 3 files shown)


src/SmartComponents/EditPolicy/BusinessObjectiveField.js
```diff
@@ -60,6 +61,10 @@ class BusinessObjectiveField extends React.Component {
         value: businessObjective.id
     });
 
+    handleInputChange = debounce(value => {
```
dLobatog: This is a valid solution IMO - if the user stops typing for 500ms, we can go ahead and create the business objective. 
However, it looks like this will pollute the DB with tons of business objectives that were not assigned. Any thoughts on how to deal with that @akofink ?

akofink: I don't understand why we would create the BO before the user clicks the 'Save' button. Why not just update a prop somewhere?

dLobatog: We can update it in Redux by dispatching the `redux-form` event. But whenever the user clicks on submit, we want to have an ID for that ready. 
I'll check if we can delay the creation and fake the ID until the very last moment.

akofink: > But whenever the user clicks on submit, we want to have an ID for that ready.

I guess I don't understand why we can't just update the shown BO in the UI at the same time as we post it to the backend. We need the ID so we can make the request again for the profile, right? Why do we even need to make this request?

src/SmartComponents/EditPolicy/BusinessObjectiveField.js
```diff
@@ -60,34 +52,37 @@ class BusinessObjectiveField extends React.Component {
...
+        this.handleCreate(value);
+    }, 500)
+
     handleChange = (newValue) => {
         this.setState({ selected: newValue });
     };
 
     handleCreate = (inputValue) => {
-        const { client, options } = this.state;
+        if (inputValue.length === 0) {
```
dLobatog: This is what was happening on your screenshot - empty options were being created

src/SmartComponents/EditPolicy/BusinessObjectiveField.js
```diff
@@ -60,34 +52,37 @@ class BusinessObjectiveField extends React.Component {
...
+            type: '@@redux-form/CHANGE',
+            meta: {
+                field: 'businessObjectiveTitle',
+                form: 'editPolicy'
+            },
+            payload: newOption
+        });
+        this.setState({
+            isLoading: false,
+            options: [newOption, ...originalOptions],
```
dLobatog: To avoid multiple new options being created, we save the original options in the state, and add the newly created one every time.

PR #174, opened 2019-07-25
RHICOMPL-269 - Allow to select 'None' business objective

dLobatog:
Screenshot from 2019-07-25 10-01-23

@katierik - thoughts about this? The text fields font-weight do not exactly match what I see on Marvelapp, to do so I would have to override default Patternfly styles, so I wasn't sure whether we should or not. 

The form field labels are in *bold* in Patternfly 4 (https://www.patternfly.org/v4/documentation/react/components/form/) 
And the titles are not in bold - https://www.patternfly.org/v4/documentation/react/components/title/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -18,7 +18,3 @@ section.pf-l-page-header.pf-l-page__main-section, section.pf-c-page-header.pf-c-
 .policy-details-dropdown {
     float: right
 }
-
-[field-id="edit-policy-business-objective"] {
-    height: 170px
-}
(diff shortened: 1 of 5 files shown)


katierik:
Match PF. I'll update the mocks to fix.

katierik:
Match PF. I'll update the mocks to fix.
Put out a request to PF to fix the styles in the sketch library to match what is actually existing to avoid this in the future

src/SmartComponents/EditPolicy/insertfile
```diff
@@ -0,0 +1,127 @@
+: INSERT INTO "rule_references" ("id","href","label") VALUES
```
akofink: What's this for?

PR #188, opened 2019-08-27
Systems table counts systems, not profiles

akofink:
To reproduce:

Upload 2 policies for 1 system, and look at the total displayed by the systems table.


bastilian:
Tested & verified. ✅

PR #196, opened 2019-09-20
Use remediationAvailable to avoid remediations API call

dLobatog:
This little thing will save us the API call to remediations API every time we hit the remediation button to fix a system (so this works on PolicyDetails or SystemDetails).

### DIFF
--- a/src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
+++ b/src/SmartComponents/ComplianceRemediationButton/ComplianceRemediationButton.js
@@ -17,7 +17,8 @@ query FailedRulesForSystem($systemIdsQuery: String!){
             profiles {
                 refId
             }
-            title
+            title,
+            remediationAvailable
         }
     }
 }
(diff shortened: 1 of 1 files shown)


akofink:
@dLobatog I now realize https://github.com/RedHatInsights/frontend-components/pull/153 goes with this PR to completely remove the remediations request, right? I'll re-test with that.

akofink:
I'm getting the following in the JS console when trying to click 'Remediation with Ansible':

```
Uncaught (in promise) TypeError: Issues array must not be empty
    at s (frontend-components-remediations.6f8b7adb.js:3)
    at t.default (frontend-components-remediations.6f8b7adb.js:3)
    at Object.t.openWizard (frontend-components-remediations.6f8b7adb.js:3)
    at SystemDetails.js:303009
```

Other than that, the PRs seem to work as expected.

akofink:
Yes, I no longer see the remediations calls when the page loads or when checking boxes. I was wondering: is the system details remediate button working for you with this change? For me, it works before this change but not after, throwing the JS error pasted above.

akofink:
> This little thing will save us the API call to remediations API every time we hit the remediation button to fix a system (so this works on PolicyDetails or SystemDetails).

> This change is visible on the Systems list, and PolicyDetails page.

Wait, so this doesn't actually affect SystemDetails?

PR #197, opened 2019-09-24
RHICOMPL-354 - Create and list OpenSCAP policies

dLobatog:
Screenshot from 2019-09-24 08-44-55
Screenshot from 2019-09-24 08-44-39

This is meant to be the UI for creating and listing policies. For now it's a draft PR as there is no data model to save this in the db.

Requires https://github.com/RedHatInsights/frontend-components/pull/177


dLobatog:
Latest update ensures data is preserved on redux between pages, and fetched in the review page.

akofink:
Looking pretty good! Is there some backend PR to go with this?

dLobatog:
Peek 2019-11-04 10-48

This is the latest update - the only thing missing is actually submitting the request to create the object.

src/SmartComponents/CreatePolicy/validate.js
```diff
@@ -0,0 +1,15 @@
+export const validateFirstPage = (benchmark, profile) => {
+    if (!benchmark || !profile) {
+        return false;
+    } else {
+        return true;
+    }
+};
+
+export const validateSecondPage = (name, refId) => {
+    if (!name || !refId) {
```
dLobatog: This is probably not enough to validate - we should make sure there are no duplicate profiles in the db, so a call to backend is in order here.

src/SmartComponents/PoliciesTable/PoliciesTable.test.js
```diff
@@ -0,0 +1,41 @@
+import toJson from 'enzyme-to-json';
+import { useQuery } from '@apollo/react-hooks';
+import PoliciesTable from './PoliciesTable.js';
+
+jest.mock('@apollo/react-hooks');
+jest.mock('../CreatePolicy/EditPolicyRules', () => {
+    return <p>Rules table</p>;
+});
+
+describe('PoliciesTable', () => {
```
dLobatog: This is missing tests for:
- Search
- Pagination

dLobatog: Solved :+1: ! Learned a bit about testing `withRouter` components in the process...

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -0,0 +1,114 @@
...
+                                                    type='radio'
+                                                    value={id}
+                                                    id={id}
+                                                />
+                                                {` ${title} - ${version}`}
+                                            </Text>
+                                        );
+                                    })}
+                                </FormGroup>
+                                <FormGroup label="Profile type" isRequired fieldId="profile-type">
```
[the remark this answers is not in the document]
katierik: You're right - this needs a proper empty state. Will update when posted

dLobatog:
@akofink I should've put this somewhere - you need https://github.com/RedHatInsights/frontend-components/pull/177 in order to go past Step 2, as the rules table needs to be modified a bit

src/SmartComponents/CreatePolicy/ProfileTypeSelect.js
```diff
@@ -0,0 +1,58 @@
...
+import Truncate from 'react-truncate';
+import {
+    Grid,
+    GridItem,
+    TextContent,
+    TextVariants,
+    Text
+} from '@patternfly/react-core';
+
+class ProfileTypeSelect extends React.Component {
```
bastilian: This is a good candidate to be in `Presentational Components` and be turned into a functional compnent.

src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -5,13 +5,14 @@ import {
...
     Form,
+    Title,
     Button
 } from '@patternfly/react-core';
 import React, { Component } from 'react';
 import propTypes from 'prop-types';
 import UpdateProfileButton from './UpdateProfileButton';
-import ProfileThresholdField from './ProfileThresholdField';
-import BusinessObjectiveField from './BusinessObjectiveField';
+import ProfileThresholdField from '../ProfileThresholdField/ProfileThresholdField';
```
bastilian: If you add these to `SmartComponents/index.js` you can use `import { ProfileThresholdField, BusinessObjectiveField } from 'SmartComponents'`

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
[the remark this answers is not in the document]
katierik: Based on the original design for the empty state (here: https://marvelapp.com/c2bcd9h/screen/61668472) I added a new screen to the mock deck to reflect the latest. 
Latest here: https://marvelapp.com/c2bcd9h/screen/61843314

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -0,0 +1,114 @@
...
+                                    <ProfileTypeSelect profiles={selectedBenchmark && selectedBenchmark.profiles} />
```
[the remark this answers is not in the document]
katierik: Is there some way we can consolidate these security guides? 
It looks like (and please correct me if I am wrong) that there is a security guide for RHEL 6, and that security guide has many versions. I am not clear if the security guide version directly corresponds to a RHEL version (e.g. 6.2), or if they are backwards compatible. Also, I'm not sure about the use case someone would have choosing an older version of the security guide. 

Going to loop in Mohit for visibility & discussion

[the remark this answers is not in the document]
akofink: There are various SCAP Security Guide (SSG) versions which contain all the benchmarks for all software available to scan against (i.e. RHEL6, RHEL7, Debian 9, firefox, etc.). So an SSG version has many benchmarks. Displayed above is `<Benchmark name> - <SSG version>`. We could very well allow users to select an SSG version first, then select from available benchmarks. Or, more useful perhaps, we could assume users always want the latest SSG version and don't display or bother users with the SSG version; in that case, we'd just display available benchmarks.

To add to the confusion though, systems are scanned against whatever SSG version is installed by the latest `scap-security-guide` package currently installed on that system, which is different for RHEL6, 7, and 8. We still haven't designed around this; We could, for example, have the client download the correct most recent SSG from the compliance app and scan clients using that, but there are issues with this approach.

src/SmartComponents/CompliancePolicies/CompliancePolicies.test.js
```diff
@@ -1,27 +1,28 @@
 import toJson from 'enzyme-to-json';
 import { useQuery } from '@apollo/react-hooks';
-
 import { CompliancePolicies } from './CompliancePolicies.js';
 
 jest.mock('@apollo/react-hooks');
+jest.mock('../CreatePolicy/EditPolicyRules', () => {
```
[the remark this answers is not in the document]
dLobatog: Yeah, if you do not mock it, it will try to query the list of rules and fail to do so. The PR depends on https://github.com/RedHatInsights/frontend-components/pull/177

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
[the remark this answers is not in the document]
katierik: @akofink This is good information.

This seems to me like it needs a little discussion. I'm thinking about the SSG version updating (do we update the policies? does the user need to manually go and do that?) and what the differences between SSG versions might have on the reporting. AFAIK there haven't been any discussions around this and the effect on the user.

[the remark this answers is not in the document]
akofink: We have talked a bit with the remediations service about the SSG versions, as they need to keep up to date to generate remediations. We've also raised it with the SCAP SMEs and the openscap team, and they recommend to try and always use the latest SSG version. Again, this isn't what's currently happening, as scans happen against the packaged scap-security-guide with RHEL. We should discuss this again with Mohit and see if we should plan some tasks around it.

We run a task every Sunday at midnight to check for a new SSG version and import it, along with its benchmark, profiles, rules, etc. There is no updating of previous records (rules, profiles, etc.); they are all new and not associated with any test results. Test results must point back to valid historical rule/profile data to properly display results and generate remediations. If users want to use the new SSG version, they can download it to their clients themselves and set their scans to reference it (I can elaborate here, if needed). There is currently no support to do this automatically via the insights-client or other mechanism.

If users do update their own SSG, we have this case: Users can also upload reports generated from SSG versions and/or benchmarks the compliance app hasn't previously imported. In this case, we import the SSG and its entire benchmark (and associated profiles, rules, etc.) based on the user's uploaded test result report (test results files also contain their entire benchmark data). I'm not super comfortable with this, as it's allowing a user with untrusted input to affect other users in different accounts. We've only just started talking about how to mitigate this, but we should also loop in PM.

bastilian:
> Merging #197 into master will decrease coverage by 6.72%.

Just a friendly reminder not to forget to add a few more tests. :)

dLobatog:
@bastilian notice that codecov is actually from a pretty old version of the PR - I'm trying to retrigger it (https://codecov.io/gh/RedHatInsights/compliance-frontend/pull/197/diff)

dLobatog:
https://codecov.io/gh/RedHatInsights/compliance-frontend/commit/bbbd1569c2bc8b522a762ff57dd0d66704fc9219 let's see how long it takes to finish :smile:  had to trigger this through the CLI

src/SmartComponents/CreatePolicy/FinishedCreatePolicy.test.js
```diff
@@ -0,0 +1,30 @@
...
+        /* eslint-disable react/display-name */
+        FinishedCreatePolicyWrapper = (props) => (
+            <FinishedCreatePolicy {...props} store={store} />
+        );
+        /* eslint-enable react/display-name */
+        FinishedCreatePolicyComponent = shallow(<FinishedCreatePolicyWrapper />).dive();
+    });
+
+    it('expect to render without error', () => {
+        expect(toJson(FinishedCreatePolicyComponent)).toMatchSnapshot();
```
bastilian: This snapshot is only of the first component in the tree, in this case in Apollo-wrapped `FinishedCreatePolicy`. A snapshot of what `FinishedCreatePolicy` would result in with the values that are currently in `store` would be more desirable.

There are a couple of ways to do that the simplest would probably be to use `react-test-renderer`, like in the `ReviewCreatedPolicy` tests. Another way would be to use `mount` instead of shallow rendering. Or stick with `shallow` and test `FinishedCreatePolicy` directly, unwrapped.

src/SmartComponents/CreatePolicy/EditPolicyRules.js
```diff
@@ -0,0 +1,79 @@
...
+
+const columns = [
+    { title: 'Rule', transforms: [sortable] },
+    { title: 'Policy', transforms: [sortable] },
+    { title: 'Severity', transforms: [sortable] },
+    { title: <React.Fragment>{ ANSIBLE_ICON } Ansible</React.Fragment>, transforms: [sortable], original: 'Ansible' }
+];
+
+const EditPolicyRules = ({ profileId }) => (
+    <Query query={ QUERY } variables={{ profileId }}>
```
bastilian: Prefer `useQuery` in functional components. In general prefer hooks, if available, in functional components.

src/SmartComponents/CreatePolicy/EditPolicySystems.test.js
```diff
@@ -0,0 +1,29 @@
...
+        /* eslint-disable react/display-name */
+        EditPolicySystemsWrapper = (props) => (
+            <EditPolicySystems {...props} store={store} />
+        );
+        /* eslint-enable react/display-name */
+        EditPolicySystemsComponent = shallow(<EditPolicySystemsWrapper />).dive();
+    });
+
+    it('expect to render without error', () => {
+        expect(toJson(EditPolicySystemsComponent)).toMatchSnapshot();
```
bastilian: SImilar to `FinishedCreatePolicy` snapshot. In this case the snapshot should allow verifying that for example the `columns` are passed to `SystemsTable`.

src/SmartComponents/CreatePolicy/CreatePolicy.test.js
```diff
@@ -0,0 +1,38 @@
...
+        expect(component.toJSON()).toMatchSnapshot();
+    });
+
+    it('expect to render the wizard', async () => {
+        const wrapper = shallow(
+            <Provider store={store}>
+                <CreatePolicy isOpen />
+            </Provider>
+        );
+        expect(toJson(wrapper.dive().dive())).toMatchSnapshot();
```
bastilian: Oh, this is also a way to get a child component with shallow, but it will require to adjust in case we change the `CreatePolicy` component and match the depth in which the component in question resides. Better would be to use `find` to get a specific component and avoid having to `dive`.

src/SmartComponents/CreatePolicy/FinishedCreatePolicy.test.js
dLobatog: I realize that the snap is not the best, for some reason with `react-test-renderer` I get `TypeError: Cannot read property '0' of undefined`.  I'll check how can I make a better snap here

bastilian: I'm not sure, but I suspect that it has something to do with apollo. The component doesn't use apollo/react-hooks, but only hooks is mocked, which could mess up the runtime.
Using react-test-renderer or `mount` instead of `shallow` probably requyires the apollo mockprovider to be set up and wrap the component in as well.

bastilian: In case adding the apollo mock provider does not work, please, do consider doing the tests with an unwrapped exported dumb component and use shallow rendering. For now this is enough. Our components sometimes differ in how they are built, some rely on only on props, some invoke Apollo and or redux, some use Apollo hooks, some the HOCs... To setup each test with all it needs to fully operate and be able to render, might take some effort for some components. Testing the component dumb will only require to pass in the proper props and mock calls to outside from within the component itself.

Testing with the redux mockstore and apollos mockprovider, does make the test more comprehensive, but also increases the test complexity to also test the integration with everything around. This only makes sense for a few components or to write in addition to a "dumb test". 

In this case the test with the component exported could look something like this and would get us a useful:
```
import toJson from 'enzyme-to-json';
import { FinishedCreatePolicy } from './FinishedCreatePolicy.js';

describe('FinishedCreatePolicy', () => {
    it('expect to render without error', () => {
        const props = {
            mutate: jest.fn(() => Promise.resolve({})),
            benchmarkId: '1',
            cloneFromProfileId: '2',
            onClose: jest.fn(),
            refId: '3',
            name: 'NAME',
            description: 'DESC',
            systemIds: [1,2,3],
            complianceThreshold: 10
        }
        const wrapper = shallow(<FinishedCreatePolicy { ...props } />)
    });
});

```

(The mutate would need to be properly mocked)

We can add tests that allow full rendering of a component in the future, but for now let's just cover the basics and make sure code within the components works as expected and produce a snapshot to verify and assert.

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -0,0 +1,128 @@
...
+class CreatePolicy extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            isOpen: this.props.isOpen,
+            stepIdReached: 1
+        };
+
+        this.toggleOpen = () => {
```
bastilian: Do `toggleOpen` and `onNext` need to be in the constructor?

src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
```diff
@@ -0,0 +1,133 @@
...
+            id
+        }
+    }
+}
+`;
+
+class FinishedCreatePolicy extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = { percent: 0 };
```
bastilian: This and similar constructors that init state (from props) could also be written as a class propery[1].

[1] https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions

src/SmartComponents/ComplianceReports/ComplianceReports.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    const policies = data.allProfiles.filter((profile) => profile.totalHostCount > 0);
+
+    let policyCards = [];
+    let pageHeader;
+    if (policies.length) {
+        pageHeader = <PageHeader>
+            <PageHeaderTitle title="Compliance" />
+            <ComplianceTabs/>
```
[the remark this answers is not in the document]
akofink: @dLobatog This is still relevant

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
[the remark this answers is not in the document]
akofink: @dLobatog The premature wrapping in the screenshot above is still an issue for Profile type.

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -0,0 +1,123 @@
...
+                    <Wizard
+                        isOpen={isOpen}
+                        onClose={this.toggleOpen}
+                        isFullWidth
+                        isFullHeight
+                        title="Create SCAP policy"
+                        description="Create a new policy for managing SCAP compliance"
+                        steps={steps}
+                        onNext={this.onNext}
+                    />
```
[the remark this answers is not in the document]
akofink: This is weird, but I wouldn't block the PR on it.

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -0,0 +1,123 @@
...
+                component: <FinishedCreatePolicy onClose={this.toggleOpen}/>,
+                isFinishedStep: true,
+                canJumpTo: stepIdReached >= 6
+            }
+        ];
+
+        return (
+            <React.Fragment>
+                <Button variant="primary" onClick={this.toggleOpen}>
+                    Create new policy
```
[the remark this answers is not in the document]
akofink: This is weird looking on a skinny monitor, but I wouldn't block the PR on it.

bastilian:
@dLobatog Nice work! However, there are few functional issues.

* When the Wizard is finished, it requires a manual refresh in order to get the just created policy on screen.
* When reopening the Wizard (after a complete and successful previous creation) it still has the values prefilled from the before
* It doesn't seem to correctly save the BusinessObjective

There are also a few issues that aren't always reproducable: 
* It occasionally throws an error in the console with `Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state. in EditPolicySystems (created by Form(EditPolicySystems))`
 * Sometimes when passing on from the "Systems" tab in the wizard:`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in SystemsTable (created by Context.Consumer)`
* Another occasional warning is `Warning: TextInput contains an input of type number with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://fb.me/react-controlled-components` when the form opens.
* When finishing the Wizard it sometimes also just throws `Error:` and the progress is stuck 0%

These errors sometimes and sometimes don't break the wizard to complete properly. 🤷‍♂ 

Aside from these issues the code looks good now, with a few minor comments, that would be nice if addressed, but we can also clean up later on:

* The empty `constructor` that only call super aren't really doing anything
	* They don't hurt, but don't have any purpose either
* Some components could be simpler and use less returns
	* Similar to changes in `CompliancePolicies.js`
* <Query> could be replaced with `useQuery` to introduce consistency
* Not sure if safer or any good, but EditPolicySystems.js `if (selectedSystemsIds)..` could be wrapped in an `useEffect`
* EditPolicyDetails.js snapshot is still odd snapshotting just `ReduxForm`
* Why `JSON.parse` redux state values?

dLobatog:
@bastilian Thanks! I'm currently addressing these one by one, but just for now, the last question `Why JSON.parse redux state values?` - you cannot save JS objects on redux-form store, so the recommended way is to save them as a JSON string and parse whenever needed (or just don't save objects). The one object I'm saving is the profile so that I can query different attributes about it at any stage of the form.

dLobatog:
Updated with the right EditPolicyDetails snapshot (for an unconnected component)

PR #200, opened 2019-10-11
Ui review fixes

bastilian:
This addresses minor UI issues identified during a UX review

### DIFF
--- a/src/SmartComponents/EditPolicy/ProfileThresholdField.js
+++ b/src/SmartComponents/EditPolicy/ProfileThresholdField.js
@@ -32,7 +32,7 @@ class ProfileThresholdField extends React.Component {
                     isValid={validThreshold}
                     helperTextInvalid='Threshold has to be a number between 0 and 100'
                     helperText="A value of 95% or higher is recommended"
-                    label="Compliance threshold (%):">
+                    label="Compliance threshold (%)">
                     <Field name='complianceThreshold' id='complianceThreshold' isRequired={true}
                         onChange={this.handleThresholdChange}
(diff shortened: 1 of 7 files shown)


src/SmartComponents/EditPolicy/ProfileThresholdField.js
```diff
@@ -38,7 +38,8 @@ class ProfileThresholdField extends React.Component {
                         isValid={validThreshold}
                         defaultValue={threshold}
                         aria-label="compliance threshold"
-                        component={ReduxFormTextInput} type='number' />
+                        component={ReduxFormTextInput} type='number'
+                        style={ { width: '60%', display: 'block' } } />
```
akofink: Inline styles are supposed to be avoided, but I don't know why :man_shrugging:

src/Utilities/TextHelper.js
```diff
@@ -1,3 +1,11 @@
...
     return (value).toFixed(fixed) + (withPercent ? '%' : '');
 };
+
+export const pluralize = (value, singular, plural) => {
+    if (!plural) {
+        plural = singular + 's';
+    }
+
+    return (value > 1 || value === 0 ? plural : singular);
+};
```
akofink: We are [at least the second team](https://github.com/RedHatInsights/frontend-components/blob/1dba13ec39aa533d94fef9d60de7311b5b30f238/packages/remediations/src/steps/ResolutionModeStep.js#L14) requiring something like this. Maybe we could plan to move it to frontend-components or ask the UI folks for guidance?

PR #202, opened 2019-10-21
RHICOMPL-384 - Papercuts on Policy list

dLobatog:
"Systems meet compliance threshold" doesn't need to wrap
Tooltips on charts should be the black ones
  - I needed to update Patternfly-charts to the latest version, which actually allows us to remove a lot of code. 
 - Removed a semicolon in the loading state

Screenshot from 2019-10-21 13-00-02

Screenshot from 2019-10-21 12-59-49

cc @katierik

### DIFF
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -87,7 +87,7 @@ const CompliancePolicies = () => {
                 <Main>
                     <div className="policies-donuts">
                         <Grid gutter='md'>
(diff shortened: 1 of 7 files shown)


src/SmartComponents/CompliancePolicies/CompliancePolicies.js
```diff
@@ -87,7 +87,7 @@ const CompliancePolicies = () => {
                 <Main>
                     <div className="policies-donuts">
                         <Grid gutter='md'>
-                            <LoadingComplianceCards/>;
```
akofink: Aha! :laughing:

PR #203, opened 2019-10-23
RHICOMPL-376 - Fix reusing BusinessObjective

dLobatog:
Peek 2019-10-23 10-17

### DIFF
--- a/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
+++ b/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
@@ -32,6 +32,9 @@ class BusinessObjectiveField extends React.Component {
             isLoading: true
         };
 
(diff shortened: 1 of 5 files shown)


src/SmartComponents/EditPolicy/BusinessObjectiveField.js
```diff
@@ -62,20 +65,22 @@ class BusinessObjectiveField extends React.Component {
     };
 
     handleCreate = (inputValue) => {
-        if (inputValue.length === 0) {
+        const { originalOptions } = this.state;
+
+        if (inputValue.length === 0 || originalOptions.map(option => option.label).indexOf(inputValue) !== -1) {
```
dLobatog: @akofink This ensures we don't create the same option twice, if it's found, we just don't create it again

PR #207, opened 2019-11-02
Correct usage and behavior of BusinessObjective select

bastilian:
This changes the select props to not missuse the placeholder to
show the currently selected business-objective, but correctly
set the value and allow it to be unset correctly.

**Before:**

bo_notsofixed

**After:**

bo_fixed


dLobatog (approved):
Nice, this makes it much more intuitive to set a BO to none. Thanks @bastilian ! Merging to CI/QA...

PR #210, opened 2019-11-05
Update Breadcrumbs to use patternfly component

bastilian:
This replaces the `Breadcrumbs` component from `frontend-components` with the Patternfly 4 component. As the Patternfly component had no equivilant to `onNavigate`, it is using `onClick` instead and required a minor fix.

Visibly nothing changes, they are identical.


bastilian:
This doesn't depend on #209, but let's wait till that is merged, and update snapshots after it is rebased on top of it.

akofink:
It looks like there's a failing `PolicyDetails.test.js`

bastilian:
@akofink The snapshots were taken with updated packages from #209, on CI it uses older packages. Once the package PR is merged, I'll rebase this one and the snapshots should pass.

PR #213, opened 2019-11-06
Refactoring notes

bastilian:
This PR should serve as discussion ground for refactoring.

**Goals for the refactoring:**

* Remove unused code
	* Identify code that is not used and remove it
* Move "Presentational Components"
	* Move Components that are (mostly) controlled via props
	* Add a snapshot test, or two
	* Bonus: refactor component to be as simple as possible, while keeping snapshot output the same.
* Extract "Presentational Components"
	* Extract code from Smart Components, which could live in it's own component. Examples would be EmptyStates, common or similar codethat could be abstracted and used instead.
	* Add a snapshot test
* Remove Components that shouldn't exist
	* There are seem to be some "wrapper" components that seem to only pass on props
	* This adds unneccessary levels
* Add Basic tests for remaining smart components

An example for refactoring the frontend is #212 (atm. this is also included here. disregard that. the only important commit is https://github.com/RedHatInsights/compliance-frontend/commit/5767eb15d8df59680a4bc9b61a1468eabd0a7b38)


akofink:
I think we can safely close this now, right @bastilian?

bastilian:
@akofink yes. with #256 open most of what I noted here should be addressed.

PR #215, opened 2019-11-14
Simplify BusinessObjectiveField and fix behaviour

bastilian:
This simplifies the `BusinessObjectiveField` and makes more use of [react-select](https://react-select.com/props) features and fixes the issues with setting initial options.

**tl;dr:**
 * A `loadOptions` function is passed to `Select` instead of handling options in `BusinessObjectiveField`
 * `BusinessObjectiveField` is now disconnected from redux-form and `dispatch` passed
 * `handleCreate` and `handleChange` are simplified
 * `handleChange` dispatches the redux-form change
 * `UpdateProfileButton` has now a very verbose `handleBusinessObjective`

**What to test:** (Checklist for manual and jest tests)

 * [x] Create a new BusinessObjective
 * [x] Remove the BusinessObjective and ensure it is unset
 * [x] Assign an existing BusinessObjective
 * [x] Create a new BusinessObjective while one is already assigned
 * [x] Change to a different BusinessObjective
 * [x] Don't change the BusinessObjective and make sure it is **not** unset
 * [x] Try saving without any BusinessObjective assigned

### DIFF
--- a/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
+++ b/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
@@ -24,10 +24,12 @@ class BusinessObjectiveField extends React.Component {
         };
     }
 
-    loadOptions = () => {
+    loadOptions = (title) => {
+        const variables = title ? { title } : {};
         this.props.client.cache.reset();
         return this.props.client.query({
-            query: GET_BUSINESS_OBJECTIVES
+            query: GET_BUSINESS_OBJECTIVES,
+            variables
         }).then((items) => {
             return items.data.businessObjectives.map(businessObjective => (
                 this.createOption(businessObjective)
--- a/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
+++ b/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
@@ -25,6 +25,7 @@ class BusinessObjectiveField extends React.Component {
     }
 
     loadOptions = () => {
+        this.props.client.cache.reset();
         return this.props.client.query({
             query: GET_BUSINESS_OBJECTIVES
         }).then((items) => {
--- a/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
+++ b/src/SmartComponents/EditPolicy/BusinessObjectiveField.js
@@ -25,7 +25,6 @@ class BusinessObjectiveField extends React.Component {
     }
 
     loadOptions = () => {
-        this.props.client.cache.reset();
         return this.props.client.query({
             query: GET_BUSINESS_OBJECTIVES
         }).then((items) => {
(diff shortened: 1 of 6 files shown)


src/SmartComponents/EditPolicy/BusinessObjectiveField.js
```diff
@@ -22,30 +20,18 @@ class BusinessObjectiveField extends React.Component {
...
+            selected: props.businessObjective ? this.createOption(props.businessObjective) : null
         };
+    }
 
-        // Resetting the cache is necessary to reload the list of Business Objectives
-        // after removing or adding any.
-        props.client.cache.reset();
-        props.client.query({
+    loadOptions = () => {
+        this.props.client.cache.reset();
```
bastilian: I tried without this, and it seemed to work fine. I'll test it again and see if we can get rid of it, unless someone has a reason we should have this.

src/SmartComponents/EditPolicy/UpdateProfile.js
```diff
@@ -28,36 +28,66 @@ mutation createBusinessObjective($input: createBusinessObjectiveInput!) {
...
 class UpdateProfileButton extends React.Component {
-    onClick = () => {
-        const { mutate, policyId, threshold, businessObjective, onClick } = this.props;
-        let businessObjectivePromise;
-        if (businessObjective && businessObjective.create) {
-            businessObjectivePromise = mutate({
+    handleBusinessObjective = () => {
+        const { businessObjective, editPolicyBusinessObjective, mutate } = this.props;
+
+        // nothing changed
```
bastilian: These comments will become tests for this function.

src/SmartComponents/EditPolicy/BusinessObjectiveField.js
bastilian: I tried it again without it and all worked as expected -> removing it. :)

bastilian: It's back in again. While further testing it was unrelyable cleared.

src/SmartComponents/EditPolicy/UpdateProfileButton.js
```diff
@@ -0,0 +1,100 @@
...
+        businessObjective {
+            id
+            title
+        }
+    }
+}
+`;
+
+export class UpdateProfileButton extends React.Component {
+    handleBusinessObjective = () => {
```
bastilian: This is verbose on purpose, but should be considered subject to refactor in the future.

bastilian:
There were a few issues still, but these are now addressed. Ready for review. :)

bastilian:
@akofink Thanks for testing and reviewing it! There was one issue still with filtering properly, which I found while investigating the issues you found. This is now fixed

Regarding the first issue with the field not selecting the creation explicitly, this could be fixed, but we should consult with UX before we implement it. We can check `onBlur` if there is a text value and trigger creation, but how it behaves now could be expected behaviour and require the explicit clicking or confirming the create option.

For avoiding the issue with creating the same BusinessObjective twice, should first be addressed with a validation in the backend. At the moment there only seems to be a presence validation for the title. Once the validations are in place, the UI should be made to react to it.
We'll also need to

PR #225, opened 2019-11-18
Get System results properly for hosts without policies

dLobatog:


src/store/Reducers/SystemStore.js
```diff
@@ -42,13 +42,18 @@ const lastScanned = (system) => {
     return result;
 };
 
+const rulesCount = (system, rulesMethod) => {
+    const rulesCount = system.profiles.map((profile) => profile[rulesMethod]);
+    return rulesCount.length > 0 && rulesCount.reduce((acc, curr) => acc + curr);
```
dLobatog: This is the key line - without it, we were passing `[]` to `.reduce` and that threw an error `TypeError: reduce of empty array with no initial value`

src/store/Reducers/SystemStore.js
```diff
@@ -34,21 +34,26 @@ const complianceScore = (system) => (
     </React.Fragment>
 );
 
-const lastScanned = (system) => {
+export const lastScanned = (system) => {
     const dates = system.profiles.map((profile) => new Date(profile.lastScanned));
-    const last = new Date(Math.max.apply(null, dates));
+    const last = new Date(Math.max.apply(null, dates.filter((date) => date.getTime() == date.getTime())));
```
dLobatog: I had to change this one to make sure when you had 2 profiles, one with `lastScanned: 'Never'` and another with a valid date, we would get the valid date as the last scanned time in general for the system. 

To be more specific, `date.getTime()` returns `NaN` for invalid dates, and NaN is always != to NaN

src/store/Reducers/SystemStore.js
```diff
@@ -34,21 +34,26 @@ const complianceScore = (system) => (
     </React.Fragment>
 );
 
-const lastScanned = (system) => {
+export const lastScanned = (system) => {
     const dates = system.profiles.map((profile) => new Date(profile.lastScanned));
-    const last = new Date(Math.max.apply(null, dates));
+    const last = new Date(Math.max.apply(null, dates.filter((date) => date.getTime() === date.getTime())));
```
akofink: @dLobatog I'm confused. It looks like you're filtering with a tautology, `date.getTime() === date.getTime()`. Am I missing something?

PR #234, opened 2019-11-26
RHICOMPL-429 - Fix color of Compliant icon for 100%

dLobatog:


src/store/Reducers/SystemStore.js
```diff
@@ -7,12 +7,12 @@ import { EXPORT_TO_CSV } from '../ActionTypes';
...
 import { fixedPercentage } from '../../Utilities/TextHelper';
 
-const compliantIcon = (system) => (
+export const compliantIcon = (system) => (
     <React.Fragment>
         {
             ((system.rulesPassed + system.rulesFailed) === 0) ?
                 <QuestionCircleIcon style={{ color: 'var(--pf-global--disabled-color--100)' }}/> :
-                system.compliant ?
+                system.profiles.every(profile => profile.compliant === true) ?
```
dLobatog: Notice we have to rely on `compliant` - computing compliance based on 100% `rulesPassed - rulesFailed == 0` is not good enough if you have set compliance thresholds

src/store/Reducers/SystemStore.test.js
```diff
@@ -43,4 +45,76 @@ describe('SystemsStore:', () => {
...
+        const system = {
+            rulesPassed: 30,
+            rulesFailed: 300,
+            profiles: [
+                { compliant: false },
+                { compliant: false }
+            ]
+        };
+
+        expect(compliantIcon(system)).toEqual(
```
bastilian: These could be replaced with taking a snapshot.

PR #279, opened 2020-01-10
RHICOMPL-449 - Round compliance thresholds to 1st decimal

dLobatog:


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -14,6 +14,7 @@ import ProfileThresholdField from './ProfileThresholdField';
 import BusinessObjectiveField from './BusinessObjectiveField';
 import { formValueSelector } from 'redux-form';
 import { connect } from 'react-redux';
+import round from 'lodash/round';
```
akofink: We should just use `toFixed` built in to Javascript rather than lodash. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

dLobatog: Unfortunately toFixed returns a string - which is not what we want to send to GraphQL

PR #285, opened 2020-01-17
Added OUIA action for modal

psav:


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -28,6 +28,7 @@ export class EditPolicy extends Component {
     }
 
     handleModalToggle = () => {
+        insights.chrome.appAction(!isModalOpen ? 'edit-business-objective' : '');
         this.setState(({ isModalOpen }) => ({
```
dLobatog: This works, you need to find the `isModalOpen` variable first, though, which is in `this.state`. Notice you can also remove the `isModalOpen` from line 32 as we don't need to find the variable twice.

```
        const { isModalOpen } = this.state;
        insights.chrome.appAction(!isModalOpen ? 'edit-business-objective' : '');
        this.setState({
            isModalOpen: !isModalOpen,
            businessObjective: this.props.businessObjective
        });
```

^ I tested the above and it works, I don't see anything happening in the console though, but the insights.chrome.appAction method is called and the action is triggered

psav: Yeh you won't see anything in the console - but if you search the DOM for edit-business-objective - you should find it attached to an element now

PR #305, opened 2020-01-24
Do not display title twice on Threshold/BO fields

dLobatog:
Screenshot from 2020-01-24 08-11-13

This PR fixes the problem of repeated titles on the EditPolicy modal.


src/SmartComponents/EditPolicy/__snapshots__/EditPolicy.test.js.snap
```diff
@@ -60,13 +60,6 @@ exports[`EditPolicy expect to render without error 1`] = `
     title="Edit policy details"
   >
     <Form>
-      <Title
```
dLobatog: Note these Title fields are already rendered on the child component so they are not rendered in the `shallow` snap

PR #319, opened 2020-01-30
Poor man's solution to SystemsTable loading state

dLobatog:
Instead of letting the inventory think "loaded" is `true` when it's requesting just `[]`, this PR makes sure the GraphQL request has finished before setting `loaded` to `true`.  

Peek 2020-01-30 21-01

### DIFF
--- a/src/store/Reducers/SystemStore.js
+++ b/src/store/Reducers/SystemStore.js
@@ -91,9 +91,13 @@ export const systemsToInventoryEntities = (systems, entities) =>
         }
     ).filter(value => value !== undefined);
 
-export const entitiesReducer = (INVENTORY_ACTION, systems, columns) => applyReducerHash(
+export const entitiesReducer = (INVENTORY_ACTION, systems, columns, isGraphqlFinished) => applyReducerHash(
     {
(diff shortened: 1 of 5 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -91,9 +91,13 @@ export const systemsToInventoryEntities = (systems, entities) =>
         }
     ).filter(value => value !== undefined);
 
-export const entitiesReducer = (INVENTORY_ACTION, systems, columns) => applyReducerHash(
+export const entitiesReducer = (INVENTORY_ACTION, systems, columns, isGraphqlFinished) => applyReducerHash(
     {
         [INVENTORY_ACTION.LOAD_ENTITIES_FULFILLED]: (state) => {
+            if (!isGraphqlFinished()) {
```
dLobatog: This pattern looks strange but the very same authors of the Inventory use it on the vulnerabilities systems table:

https://github.com/RedHatInsights/vulnerability-ui/blob/b4fa8c50b6f4ad1892eadf1a78e31cac106511cb/src/Components/SmartComponents/SystemsExposedTable/SystemsExposedTable.js#L129

PR #320, opened 2020-01-31
Add system count to Policies table

dLobatog:
Adding system count to the table, to match https://marvelapp.com/c2bcd9h/screen/62294966 a bit more.
Screenshot from 2020-01-31 10-33-17

### DIFF
--- a/src/SmartComponents/PoliciesTable/PoliciesTable.js
+++ b/src/SmartComponents/PoliciesTable/PoliciesTable.js
@@ -48,6 +48,7 @@ export class PoliciesTable extends React.Component {
             columns: [
                 { title: 'Policy name' },
                 { title: 'Policy type' },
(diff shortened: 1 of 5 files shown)


src/SmartComponents/PoliciesTable/PoliciesTable.js
```diff
@@ -48,6 +48,7 @@ export class PoliciesTable extends React.Component {
             columns: [
                 { title: 'Policy name' },
                 { title: 'Policy type' },
+                { title: 'Systems count' },
```
akofink: The mock just says 'Systems' - did we receive other guidance here from UX?

dLobatog: Nope, I'll fix that

PR #321, opened 2020-02-03
Delete policy action on PoliciesTable

dLobatog:
Needs https://github.com/RedHatInsights/compliance-backend/pull/303 to work
Peek 2020-02-04 18-01

### DIFF
--- a/src/SmartComponents/DeletePolicy/DeletePolicy.js
+++ b/src/SmartComponents/DeletePolicy/DeletePolicy.js
@@ -7,40 +7,53 @@ import propTypes from 'prop-types';
 import React from 'react';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_PROFILE } from '../../Utilities/graphql/mutations';
+import { addNotification } from '@redhat-cloud-services/frontend-components-notifications';
+import { dispatchAction } from '../../Utilities/Dispatcher';
 
 const DeletePolicy = ({ isModalOpen, policy, toggle, onDelete }) => {
-    const [deletePolicy] = useMutation(DELETE_PROFILE);
+    const [deletePolicy] = useMutation(DELETE_PROFILE, {
+        onCompleted: () => {
+            dispatchAction(addNotification({
+                variant: 'success',
+                title: `Removed policy ${policy.name}`
+            }));
+            onDelete();
+            toggle();
+        },
+        onError: (error) => {
+            dispatchAction(addNotification({
+                variant: 'danger',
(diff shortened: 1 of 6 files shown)


src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -0,0 +1,47 @@
...
+import { useMutation } from '@apollo/react-hooks';
+import { DELETE_PROFILE } from '../../Utilities/graphql/mutations';
+
+const DeletePolicy = ({ name, id }) => {
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    const [deletePolicy] = useMutation(DELETE_PROFILE);
+
+    return (
+        <React.Fragment>
+            <div onClick={() => setIsModalOpen(true)}>Delete policy</div>
```
dLobatog: Still a WIP because I cannot get this hook to change the state somehow :thinking:

bastilian: Hm... this looks like it should work. Maybe it is updating state, but not rerendering?

dLobatog: Seems that the problem is that the Dropdown closes the component immediately when clicking on the item or outside :/ 
I was able to prevent that once, but not twice

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -0,0 +1,63 @@
...
+import React, { useState } from 'react';
+import { useMutation } from '@apollo/react-hooks';
+import { DELETE_PROFILE } from '../../Utilities/graphql/mutations';
+
+const DeletePolicy = ({ name, id }) => {
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    const [deletePolicy] = useMutation(DELETE_PROFILE);
+
+    const toggle = (event) => {
+        event.preventDefault();
```
dLobatog: This works and only `stopPropagation` is needed

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -0,0 +1,63 @@
...
+        event.stopPropagation();
+        setIsModalOpen(true);
+    };
+
+    const submitMutation = (event) => {
+        event.preventDefault();
+        event.stopPropagation();
```
dLobatog: This doesn't, because the onClick event is not even captured here :/

PR #347, opened 2020-02-11
RHICOMPL-468 - Update Policies and Reports pages

dLobatog:
:warning: This is just not going to go in a single PR. I'm keeping this open just to have a place where we can get all of the changes at once and see how the whole thing is supposed to look.

Screenshot from 2020-02-11 11-32-25
Screenshot from 2020-02-11 11-32-12
Screenshot from 2020-02-11 11-31-51
Screenshot from 2020-02-11 11-31-36

TODO before merging:
 - [ ] Merge https://github.com/RedHatInsights/frontend-components/pull/351
 - [ ] Review tests and coverage remain OK
- This is likely going into 3 separate PRs

PR #363, opened 2020-02-13
RHICOMPL-468 - Updated report cards

dLobatog:
I'm leaning towards splitting https://github.com/RedHatInsights/compliance-frontend/pull/347 into multiple PRs. This is one of them that is probably easy to separate. This just changes the look and feel of the cards, and renames them, as in the new UI, these are Report cards, not Policy cards (policies are just shown on a table).

The link to `View report` is linking to the same as `View policy` for now, I'm going to make another PR with the new report details page and another with the new policy details page.  

Screenshot from 2020-02-13 14-51-18


src/PresentationalComponents/ReportCard/ReportCard.js
```diff
@@ -23,23 +24,29 @@ import {
 import '../../Charts.scss';
 import { fixedPercentage } from '../../Utilities/TextHelper';
 
-class CompliancePolicyCard extends React.Component {
+class ReportCard extends React.Component {
     policy = this.props.policy;
     state = {
-        cardTitle: <Truncate lines={1}>{this.policy.name}</Truncate>
+        cardTitle: <Truncate lines={1}>{this.policy.name}&nbsp;<PolicyPopover policy={this.policy} /></Truncate>
```
bastilian: While we do keep some inventory components in states, it does feel like something to avoid.

It would be much nicer if the `cardTitle` would be something like `cardTitleTruncated` and have something like this in render:

```js
const cardTitle = `${this.policy.name}&nbsp;` + <PolicyPopover policy={this.policy} />

...

{ cardTitleTruncated ? <Truncate lines={1}>{ cardTitle }</Truncate> : cardTitle }

```

dLobatog: Fair enough, fix incoming!

bastilian:
@dLobatog LGTM! Two minor minor things, notes if will.

Per mockups, the two links should be flipped. "View report" should be first, followed by "VIew policy". The other minor thing, is that when hovering over `OutlinedQuestionCircleIcon` it would be good if it showed a proper `pointer` cursor to indicate it is clickable.

dLobatog:
Thanks @bastilian , I'm addressing that currently on another PR :smile:

PR #371, opened 2020-02-17
RHICOMPL-292 - Assign policies to UI

dLobatog:
Peek 2020-02-17 18-57

This is just a draft for now to show how to add actions to the current systems table in case it's useful until I finish it.

### DIFF
--- a/src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
+++ b/src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
@@ -22,7 +22,7 @@ const CheckboxFieldArray = ({ input, options }) => {
         <React.Fragment>
             {
                 options.map(({ label, value, defaultChecked }, index) => (
-                    <Checkbox key={label} id={label} label={label}
+                    <Checkbox key={value} id={value} label={label}
                         name={`${name}[${index}]`} value={value}
                         isChecked={inputValue === '' ? defaultChecked : inputValue.includes(value)}
                         onChange={(checked) => handleChange(checked, value)}
--- a/src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
+++ b/src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { Field } from 'redux-form';
+import propTypes from 'prop-types';
+import { Checkbox } from '@patternfly/react-core';
+
+const CheckboxFieldArray = ({ input, options }) => {
+    const { name, onChange } = input;
+    const inputValue = input.value;
+
+    const handleChange = (checked, value) => {
+        const arr = [...new Set(inputValue)];
+        if (checked) {
+            arr.push(value);
+        } else {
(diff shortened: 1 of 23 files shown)


src/PresentationalComponents/ReduxFormWrappers/ReduxFormWrappers.js
```diff
@@ -18,6 +18,7 @@ export const ReduxFormTextArea = field => (
 
 export const ReduxFormCheckboxInput = field => (
     <Checkbox
+        checked={field.isChecked}
```
bastilian: Maybe it's better to use `isChecked` looking at the source of `Checkbox`[1] it seems like `isChecked` overrides `checked` when an `onChange` is set.

[1] https://github.com/patternfly/patternfly-react/blame/master/packages/patternfly-4/react-core/src/components/Checkbox/Checkbox.tsx#L64-L72

src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
```diff
@@ -0,0 +1,60 @@
...
+        const handleChange = (checked) => {
+            const arr = [...new Set(inputValue)];
+            if (checked) {
+                arr.push(value);
+            } else {
+                arr.splice(arr.indexOf(value), 1);
+            }
+
+            return onChange(arr);
+        };
```
bastilian: The change handler would be better outside the mapping with the value passed.

src/PresentationalComponents/CheckboxGroup/CheckboxGroup.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { Field } from 'redux-form';
+import propTypes from 'prop-types';
+import { Checkbox } from '@patternfly/react-core';
+
+const CheckboxFieldArray = ({ input, options }) => {
+    const { name, onChange } = input;
+    const inputValue = input.value;
+
+    const checkboxes = options.map(({ label, value, defaultChecked }, index) => {
```
bastilian: This could also go directly into the return to render like this:
```jsx
{ 
  options.map(({ label, value, defaultChecked }, index) => {
    const checked = inputValue === '' ? defaultChecked : inputValue.includes(value)
    <Checkbox key={label} id={label} label={label} 
              name={`${name}[${index}]`} value={value} isChecked={checked} 
              onChange={(checked) => handleChange(checked, value)} />
  })
}
```

akofink:
@katierik what do you think? The words in the modal seem a bit squished to me.

image

bastilian:
@dLobatog Codewise it looks good to me, but there is something funky going on:

Kapture 2020-02-19 at 13 29 26

katierik:
> @katierik what do you think? The words in the modal seem a bit squished to me.
> 
> image

They are! I noticed this (and missed the comment) and opened a JIRA bug for it. 
There should be a medium spacer between the instructions and the checkboxes

dLobatog:
Also updated all buttons to be left-aligned as per https://projects.engineering.redhat.com/browse/RHICOMPL-489

PR #372, opened 2020-02-18
Add bool allSystems option for SystemsTable

akofink:
allSystems=true for host-policy wizard
allSystems=false for systems tab (aka reports by system)

### DIFF
--- a/src/store/Reducers/SystemStore.js
+++ b/src/store/Reducers/SystemStore.js
@@ -10,6 +10,8 @@ import {
 } from '../../PresentationalComponents';
 
 export const lastScanned = (system) => {
+    if (system.profiles === undefined) { return 'Never'; }
+
     const dates = system.profiles.map((profile) => new Date(profile.lastScanned));
     const last = new Date(Math.max.apply(null, dates.filter((date) => isFinite(date))));
     const result = (last instanceof Date && isFinite(last)) ? last : 'Never';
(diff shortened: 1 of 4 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -22,83 +22,81 @@ export const rulesCount = (system, rulesMethod) => {
...
-            if (matchingEntity === undefined) {
-                return;
-            }
+export const systemsToInventoryEntities = (systems, entities, allSystems) =>
+    entities.map(entity => {
+        // This should compare the inventory ID instead with
+        // the ID in compliance
+        if (entity.facts === undefined) { entity.facts = {}; }
+        let matchingSystem = systems.find((system) => {
+            return entity.id === system.id;
```
dLobatog: It looks like this should be `system.node.id` isn't it? Otherwise id is indefined from what I'm seeing

src/store/Reducers/SystemStore.js
```diff
@@ -22,83 +22,81 @@ export const rulesCount = (system, rulesMethod) => {
...
+        }) || {'profiles': []};
+        if (matchingSystem === {}) {
+            if (!allSystems) { return; }
+        } else {
+            matchingSystem.profileNames = matchingSystem.profiles.map((profile) => profile.name).join(', ');
```
dLobatog: Looks like `matchingsystem` here would also need `.node` to get any attributes :confused:

src/store/Reducers/SystemStore.js
```diff
@@ -22,83 +22,82 @@ export const rulesCount = (system, rulesMethod) => {
...
-                        compliance_score_text: complianceScoreString(system),
-                        last_scanned: (system.lastScanned instanceof Date) ?
-                            { title: <FormattedRelative value={Date.parse(system.lastScanned)} /> } :
-                            system.lastScanned,
-                        last_scanned_text: system.lastScanned
-                    }
-                }
-                /* eslint-enable camelcase */
-            };
+        let matchingSystem = systems.map((s) => s.node).find((system) => {
```
akofink: Updated to call `.node`

akofink:
@dLobatog Ok, tests should be good. We still need to add a filter to toggle this allSystems prop, but maybe we can add that in a future PR.

akofink:
Updated to hide the toolbar if allSystems

PR #373, opened 2020-02-18
Configure filters and chips for InventoryTable

bastilian:
This removes our custom components for the inventory table toolbar in `SystemsTable` and instead makes use of the configurable filters.

Screenshot 2020-02-18 at 17 12 15

**Notes:**
Here be dragons. it works, but with kinks, known issues that should be fixed:
 * [x] The `ConditionalFilter` component is missing a feature to only show the filter icon and hide the label, unless toggled open[1]
 * [x] Clearing a filter will clear the whole "category"
 * [ ] When clearing a search term the value lingers around in the input.*
 * [x] On loading the export toggle has a loading state that rips apart the layout[2]
 * [x] Filter updates don't reset the page to the first yet
 * [ ] Exporting CSV and JSON works, but selectedEntities aren't recognized. (This seems already broken in `master`.)
 * [ ] FilterChips show the field value atm. the onChange only returns the value and there is no proper matching to an appropriate label yet.
 * [ ] SystemsTable has become complexer and should be cleaned up

[1] https://github.com/RedHatInsights/frontend-components/pull/367
[2] https://github.com/RedHatInsights/frontend-components/pull/368
[*] These issues need to be addressed in `frontend-components`


bastilian:
Thanks for taking a look, @akofink.

Regarding your notes:

> Search by name doesn't happen on typing - you have to press enter

This is on purpose, for now, I'll investigate, but having it `onChange` breaks the page (temporarily).

> Lots of failed prop type errors in the JS console

One of them is likely, because this uses a not yet released new prop for the `ConditionalFilter` to hide the label, as in the mock[1].

> Compliance and Non-compliant can both be active, which I think is fine, ...Same with compliance score

The filtering is a bit confusing, I tried to keep the same functionallity. 
We could make it a radio group, meaning you can only select either compliant or non-comopliant and not both. 
For the score it makes senst to be able and select multiple ranges.

> ...  but clicking the X on one clears both. ...

Yeah, this is not yet implemented. I noted it in the PR description.

> The text is obscured for name or reference search

Will be fixed. :)

[1] https://github.com/RedHatInsights/frontend-components/pull/367

bastilian:
Functionally it should now be ready to test further. 
The query text in the filter chips is fixed, as well as the filter clearing odities.

There is also now a PR open to fix the issue with the "export button" breaking the table toolbar layout when loading.[1]

[1] RedHatInsights/frontend-components#368

bastilian:
This is now ready for a proper testing and reviewing. 

The things that are still open from my notes: 
> When clearing a search term the value lingers around in the input.

The search term gets properly removed, but somewhere the value persists staying around, but I believe it is more an issue to be addressed in the frontend-components.

> Exporting CSV and JSON works, but selectedEntities aren't recognized.

This seems to be broken already, as mentioned. On CI the export does as well contain all items, even if some are just selected.

> FilterChips show the field value atm. the onChange only returns the value and there is no proper matching to an appropriate label yet. & SystemsTable has become complexer and should be cleaned up

Over all I'm not 100% happy with the code. `updateFilterChips` is an eysore, which is where the mapping to nice labels could happen, but I'm out of ideas at the moment how to get to the point to implement it nicely.

For now, it is working as it should, but should be improved at some point. 
The nicest would be if there'd be a framework to work with the toolbar that helps manage filters like that and coordinates filters and filterchips and would only require passing a config.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -135,6 +114,134 @@ class SystemsTable extends React.Component {
...
+
+    updateSearchFilter = debounce((_event, selectedValue) => {
+        this.setState({
+            ...this.state,
+            page: 1,
+            search: selectedValue
+        }, this.filterUpdate);
+    }, 500)
+
+    updateCompliancFilter = debounce((_event, selectedValues, value) => {
```
dLobatog: Typo, `updateComplianceFilter`

src/SmartComponents/SystemsTable/SystemsTable.test.js
```diff
@@ -7,6 +7,15 @@ import { Provider } from 'react-redux';
 jest.mock('@redhat-cloud-services/frontend-components-inventory-compliance', () =>
     () => 'Mocked ComplianceSystemDetails'
 );
+const mockExportJsonFunction = jest.fn();
+jest.mock('Utilities/Export', () => ({
+    exportToJson: () => mockExportJsonFunction()
+}));
+
+// We mock the debounce function otherwise we'd have to deal with time.
+import debouce from 'lodash/debounce';
```
dLobatog: typo, debounce

src/Utilities/Export.js
```diff
@@ -38,11 +40,21 @@ export const csvFromState = (state) => {
...
+        linkAndDownload(csv, filename());
     }
 };
+
+export const exportToJson = (selectedEntities) => {
+    const url = (selectedEntities.length > 0) ?
+        COMPLIANCE_API_ROOT + '/systems.json' + '?search=(id ^ (' + selectedEntities.join(',') + '))' :
+        COMPLIANCE_API_ROOT + '/systems.json';
+
+    linkAndDownload(url, filename('json'));
```
dLobatog: Note, this would fix https://projects.engineering.redhat.com/browse/RHICOMPL-117

src/SmartComponents/SystemsTable/filterConfig.js
```diff
@@ -0,0 +1,68 @@
...
+    ''
+);
+
+export const labelForValue = (value, category) => {
+    try {
+        return noOpDefaultConfig.items.filter((item) => (item.label === category))[0]
+        .filterValues.items.filter((item) => (item.value === value))[0].label;
+    }
+    catch (_) {
+        console.info('No label found for ' + value + ' in ' + category);
```
dLobatog: This won't pass eslint

src/Utilities/FilterBuilder.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    if (search.length > 0) {
+        additionalFilter.push(`name ~ ${search}`);
+    }
+
+    return [additionalFilter.join(' and ')].filter((f) => (f.length > 0));
+};
+
+export const buildFilterString = (state) => {
```
dLobatog: Can we get a test for this one for sample inputs/outputs?

src/SmartComponents/SystemsTable/filterConfig.js
```diff
@@ -0,0 +1,69 @@
...
+    updateSearchFilter,
+    updateCompliancFilter,
+    { complianceScores, complianceStates },
+    search
+) => ({
+    hideLabel: true,
+    items: [
+        {
+            type: conditionalFilterType.text,
+            label: 'Name or reference',
```
dLobatog: Just `Name` is fine, name or reference is probably coming from the rules table, there's no system reference

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -6,7 +6,7 @@ import {
     TextContent,
     TextVariants
 } from '@patternfly/react-core';
-import SystemsTable from '../SystemsTable/SystemsTable';
+import { SystemsTable } from 'SmartComponents';
```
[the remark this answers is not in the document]
bastilian: Thanks for the hint! I was lost at first with this one. What SystemsTable exported changed as well, so just fixing that was not enough. there is now an export of the SystemsTable just with Apollo and no redux, that seemed to have caused issues as well. I also discovered and issue where it would fire the change always, now it only does when the systemIds change. The form should work now again as expected.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -135,6 +114,134 @@ class SystemsTable extends React.Component {
         this.state.loaded
     )
 
+    onExportSelect = (_, format) => {
+        if (format === 'csv') {
```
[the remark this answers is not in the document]
bastilian: Done.

src/SmartComponents/SystemsTable/filterConfig.js
bastilian: I disabled the rule for the this line.

src/Utilities/FilterBuilder.js
bastilian: Added a few tests now.

src/SmartComponents/SystemsTable/filterConfig.js
bastilian: Changed.

bastilian:
@dLobatog Thanks for reviewing and testing! Everything that can be addressed should be done now.
 
* Filters are built properly now as they were before in master, and has tests to show input and output for the `FilterBuilder` are added.
* The misplacement issue on loeading should be fixed with thin the next frontend-components update. The hidden label should already be in the package version we use, so updating the packages when testing should reflect this already.
* The order of export button and remidiations button is tricky. 
It is possible to put the button right after the filter as it is on the mocks, but we'd need to build the export button again ourselfs or and work around it somehow. 
We should be able to add actions on that position, but we'll need to change the Primary Toolbar in frontend-components to be able to do that properly. It shouldn't be hard. I'll look into that.

src/SmartComponents/SystemsTable/filterConfig.js
```diff
@@ -0,0 +1,69 @@
+import {
+    conditionalFilterType
+} from '@redhat-cloud-services/frontend-components';
+
+export const defaultFilterConfig = (
+    updateSearchFilter,
+    updateCompliancFilter,
```
dLobatog: Typo, updateComplianceFilter

bastilian:
@dLobatog Thanks for testing and reviewing! I took a look at the issues you found.

The filter changes now trigger a loading state. There are still occasions where updates can trip over each other, while one request is loading and another happens to be done sooner it'll get in a weird state where the UI stalls and shows "no hosts" found. I tried to minimize the chances of that happing, but quick fingers and slow responses could still see an undesired state.

The filters now also survive pagination interactions. Overall it should be much better now.
This also now contains a `FilterConfigBuilder`, to get this out of `SystemsTable` and make it more generic to use with other tables/toolbars/etc. that take such a filterConfig.

The `FilterBuilder` could be more generic and use `FilterConfigBuilder` to make it reusable elsewhere and the chip-management could use a brush up as well, and there is still the issue with the input text hangin' around, but that's for another day, unless you insist.

I'm unsure what to do about the "N/A" hosts, should we consider them "uncompliant" and see how we can hide them when the filter is to be compliant?

bastilian:
@dLobatog It's now rebased upon master. I tested it and filters and all still work, including assigning, after rebasing and fixing, but please, have another look.

PR #389, opened 2020-02-21
 RHICOMPL-469 - Compliance Reports should have a By Systems tab 

dLobatog:
The "By systems" tab is now the old systems table, which only contains hosts with compliance policies assigned:  
Screenshot from 2020-02-21 17-11-31

And the /systems page is now the whole inventory with the actions to assign the policies:
Screenshot from 2020-02-21 17-11-09

### DIFF
--- a/src/SmartComponents/ReportsSystems/ReportsSystems.js
+++ b/src/SmartComponents/ReportsSystems/ReportsSystems.js
@@ -1,13 +1,13 @@
 import React from 'react';
 import SystemsTable from '../SystemsTable/SystemsTable';
-import { ComplianceTabs } from 'PresentationalComponents';
+import { ComplianceTabs, ReportTabs } from 'PresentationalComponents';
 import { PageHeader, PageHeaderTitle, Main } from '@redhat-cloud-services/frontend-components';
 
-const ComplianceSystems = () => {
+const ReportsSystems = () => {
     const columns = [{
         composed: ['facts.os_release', 'display_name'],
         key: 'display_name',
-        title: 'Name',
+        title: 'System name',
         props: {
             width: 40
         }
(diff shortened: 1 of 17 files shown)


src/SmartComponents/ReportsSystems/ReportsSystems.js
```diff
@@ -0,0 +1,48 @@
...
+            width: 10
+        }
+    }, {
+        key: 'facts.compliance.last_scanned',
+        title: 'Last scanned',
+        props: {
+            width: 10
+        }
+    }];
+    const beta = window.location.pathname.split('/')[1] === 'beta';
```
bastilian: There is `window.insights.chrome.isBeta()` to check this.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -300,13 +300,13 @@ class SystemsTable extends React.Component {
...
-            total={ totalCount }
+            total={ allSystems ? undefined : totalCount}
             perPage={ perPage }
             variant={ compact ? pfReactTable.TableVariant.compact : null }
             items={ allSystems ? undefined : items.map((edge) => edge.node.id) }
-            filterConfig={ filterConfig }
+            filterConfig={ allSystems ? undefined : filterConfig }
             exportConfig={ exportConfig }
-            activeFiltersConfig={{
+            activeFiltersConfig={allSystems ? undefined : {
```
bastilian: Just a note: This makes me feel iffy. I think the proper way would be to define an object before that either contains or does not contain props. Something like this should only put the values in the object when they are defined:

```js
const filterConfig = allSystems ? undefined : filterConfig;
const total= allSystems ? undefined : totalCount;
----
const inventoryTableProps = {
    filterConfig
}
----
<InventoryCmp
   { ...ineventoryTableProps }
/>
```

There is already a warning in master that is moaning about props being `undefined` but required, so we should fix this up at some point.

dLobatog: Fully onboard with that, I'm updating that now

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -268,19 +268,18 @@ class SystemsTable extends React.Component {
     }
 
     render() {
-        const { remediationsEnabled, compact, enableExport } = this.props;
+        const { remediationsEnabled, compact, enableExport, allSystems } = this.props;
```
dLobatog: Notice `allSystems` should come from props, not state, so this was broken before

PR #392, opened 2020-02-24
Add checkbox to optionally remove all test results aka. reports

bastilian:
Adds a checkbox to optionally remove reports for a given policy.
Screenshot 2020-02-24 at 09 55 36

Accompanying backend PR: https://github.com/RedHatInsights/compliance-backend/pull/326

### DIFF
--- a/src/SmartComponents/DeletePolicy/DeletePolicy.js
+++ b/src/SmartComponents/DeletePolicy/DeletePolicy.js
@@ -1,16 +1,18 @@
 import {
     Modal,
     TextContent,
-    Button
+    Button,
+    Checkbox
 } from '@patternfly/react-core';
 import propTypes from 'prop-types';
-import React from 'react';
+import React, { useState, useEffect } from 'react';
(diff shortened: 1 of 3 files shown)


src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -31,6 +33,16 @@ const DeletePolicy = ({ isModalOpen, policy, toggle, onDelete }) => {
...
     });
     const { name, id } = policy;
+    const variables = {
+        input: {
+            id,
+            deleteAllTestResults
+        }
+    };
+
+    useEffect(() => {
```
bastilian: I'm not sure why, yet, but for some reason the `deleteAllTestResults` state is retained throughout all checkboxes.
Even with different `id` props they still stay selected when checked and the modal closed and a different delete modal is opened. this effect just makes sure that when the policy (aka. the modal opens) the state is set to the default state. (which is here still false, but should maybe be true)

PR #395, opened 2020-02-25
Create host if not exists on host-profile association

akofink:
Requires https://github.com/RedHatInsights/compliance-backend/pull/330

### DIFF
--- a/src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
+++ b/src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
@@ -15,21 +15,21 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 const QUERY = gql`
-query systemWithProfiles($id: String!){
-    system(id: $id) {
-        id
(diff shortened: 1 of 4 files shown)


src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
```diff
@@ -15,8 +15,8 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 const QUERY = gql`
-query systemWithProfiles($id: String!){
-    system(id: $id) {
+query systemWithProfiles($search: String!){
+    allSystems(search: $search) {
```
dLobatog: I know I haven't marked this endpoint as deprecated (we should have a system for this that's better than just commenting) but this endpoint should be deprecated in favor of `systems` which matches GraphQL-Relay format & is easier to maintain

akofink: Updated to use systems instead of allSystems.

PR #396, opened 2020-02-26
Only use @apollo/react-hooks & @apollo/react-hoc

bastilian:
This changes apollo related modules to be used from `@apollo/react-hooks` and `@apollo/react-hoc` and removes the duplicate `ApolloProvider` in entry(-dev).js

This also updates packages that are not monitored by our bot to the latest versions and removes the (unused) packages `classnames` & `create-react-class`.


src/entry.js
```diff
@@ -28,11 +27,9 @@ window.insights.chrome.auth.getUser().then(() => {
         <Provider store={ init().getStore() }>
             <IntlProvider locale={navigator.language}>
                 <Router basename={`${release}${pathName[0]}/${pathName[1]}`}>
-                    <ApolloProviderHooks client= {client}>
-                        <ApolloProvider client={client}>
-                            <App />
-                        </ApolloProvider>
-                    </ApolloProviderHooks>
+                    <ApolloProvider client={client}>
```
dLobatog: Looks like this is the main point of the PR, nice   :smile:

PR #399, opened 2020-02-27
Use DateFormat component to get dates with tooltips

bastilian:
Relative dates should have the full datetime displayed in a tooltip. 
Using DateFormat provides that.

Screenshot 2020-02-27 at 11 34 24


katierik:
@akofink the last thing I heard from Ben is that this was a bug on the inventory side, they are supposed to be using the relative time.

PR #401, opened 2020-02-27
Centralise error handling

bastilian:
Currently, we handle errors per page, inconsistently, this is an attempt to correct this.
It introduces a `StateView` and a `StateViewWithError` based on that. 

These are used to define `StateViewPart`s that contain components to render for certain states.
The `StateViews` have a object passed that container keys matched to the `stateKey` props of each part. If the value for that key is not `undefined` it will render the corresponding part.

The `StateViewWithError` has the `ErrorPage` defined to render in case of an error and only needs to have a `error` property in the states object passed in order to render it.

The `ErrorPage` itself is now also handling `404` errors and will show the `InvalidObject` component:

Screenshot 2020-03-10 at 02 15 44

It also has been changed to look similar to the `InvalidObject` component:

Screenshot 2020-03-10 at 02 16 05

For this it would be good to have feedback from @katierik. 

For now only the details pages for reports, policies and systems have been adapted to use the `StateViewWithError`.


src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -34,45 +37,38 @@ query System($inventoryId: String!){
...
-        <React.Fragment>
+    return <StateViewWithError stateValues={ { error, data, loading } }>
+        <StateViewPart stateKey='data'>
             <PageHeader>
                 <Breadcrumb>
                     <BreadcrumbItem to='/rhel/compliance/systems' onClick={ onClick }>
                         Systems
                     </BreadcrumbItem>
-                    <BreadcrumbItem isActive>{ data.system.name }</BreadcrumbItem>
+                    <BreadcrumbItem isActive>{ systemName }</BreadcrumbItem>
```
bastilian: This caused an interesting issue.

The issue was when `data` was undefined that React still tried to render this component (noted: in `SystemDetails`' components context not `StateView`), even though `StateView`  should prevent this. React still tried to access `system.name` on `data`. This is seems to be an oddness of how React renders and passes on child components and tries to determine props, i think. I would have to investigate deeper on why this happens.

bastilian:
@akofink thanks for checking it out! 

Regarding the 401 handling: at the moment only details pages use the State view and improved error page, lists and other pages will need to be adapted over time. it shouldn't be hard at all, but it would be best to keep those out of this PR. so, in short: the 401 handling should work on any system, report or policy detail.

The error page desgin and copy is basic and should definitelly be nicer with additional feedback from UX and present errors properly either generalised or handled for each kind of error.

akofink:
Yep, okay - I see the 401 handled now with the same error page:

image

I thought, based on the code (`chrome.auth.logout()`), I'd be sent back to the SSO login screen. IMO this isn't an error we need to actually show the user, since we can fix it ourselves by forcing the user to login again.

bastilian:
@akofink The logout should redirect you as well now, the call was missing a "bounce" argument to perform the redirect it does pass it now. I only tested that it calls the logout function, but i've got the jwt object mocked locally, mind testing again and see with your setup (given you are not too mocking  the `jwt` function too :)).

bastilian:
@akofink Thanks for testing it again! It now shouldn't render anything. It's not the best yet, but should be good enough for now.

PR #402, opened 2020-02-27
Call systemFetch on assigning policies

bastilian:


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -288,7 +288,9 @@ class SystemsTable extends React.Component {
                 isModalOpen={isAssignPoliciesModalOpen}
                 id={selectedSystemId}
                 fqdn={selectedSystemFqdn}
-                toggle={() => this.setState((prev) => ({ isAssignPoliciesModalOpen: !prev.isAssignPoliciesModalOpen }))}
+                toggle={() => this.setState((prev) => (
+                    { isAssignPoliciesModalOpen: !prev.isAssignPoliciesModalOpen }
+                ), this.systemFetch)}
```
akofink: An API call every time I toggle the modal? That doesn't seem quite right. I'd think we should check if the list of profiles in the table matches what the modal says and refresh only if they're different.

PR #405, opened 2020-02-28
RHICOMPL-389 - Show rules/lastScan/compliant per profile on systems table

dLobatog:
Before, the systems table displayed the *global* compliance, *global* failed rules count, and *global* last scanned, even when you visit the ReportDetails page, or the PolicyDetails page. 

Now, on the "Reports" "By systems" page, we show the global values, but on every specific "ReportDetails" or "PolicyDetails" page, we only see the values related to that profile.

### DIFF
--- a/src/PresentationalComponents/ComplianceScore/ComplianceScore.js
+++ b/src/PresentationalComponents/ComplianceScore/ComplianceScore.js
@@ -11,7 +11,7 @@ const CompliantIcon = (system) => (
         {
             ((system.rulesPassed + system.rulesFailed) === 0) ?
                 <QuestionCircleIcon style={{ color: 'var(--pf-global--disabled-color--100)' }}/> :
-                system.profiles.every(profile => profile.compliant === true) ?
+                system.compliant ?
                     <CheckCircleIcon style={{ color: 'var(--pf-global--success-color--100)' }}/> :
                     <ExclamationCircleIcon style={{ color: 'var(--pf-global--danger-color--100)' }}/>
         }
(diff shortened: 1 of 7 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -9,24 +9,39 @@ import {
     complianceScoreString
 } from 'PresentationalComponents';
 
-export const lastScanned = (system) => {
-    if (system.profiles === undefined) { return 'Never'; }
+const findProfiles = (system, defaultReturn, profileId) => {
```
dLobatog: The bulk of the meaningful changes are found on this file

src/PresentationalComponents/ComplianceScore/ComplianceScore.js
```diff
@@ -11,7 +11,7 @@ const CompliantIcon = (system) => (
         {
             ((system.rulesPassed + system.rulesFailed) === 0) ?
                 <QuestionCircleIcon style={{ color: 'var(--pf-global--disabled-color--100)' }}/> :
-                system.profiles.every(profile => profile.compliant === true) ?
+                system.compliant ?
```
dLobatog: It's probably a good idea to remove this `.every` from here anyway, it's even more of a "PresentationalComponent" now that it just limits itself to presenting content, not computing it

src/store/Reducers/SystemStore.js
```diff
@@ -9,24 +9,39 @@ import {
...
+    if (system.profiles === undefined) { return defaultReturn; }
 
-    const dates = system.profiles.map((profile) => new Date(profile.lastScanned));
+    let profiles = system.profiles;
+
+    if (profileId !== undefined && profileId !== '') {
+        profiles = profiles.filter((profile) => profile.id === profileId);
```
bastilian: Nice! You could also make the `profileId` an array of ids and do `profileIds.includes(profile.id)` to allow reducing down to multiple profiles.

PR #406, opened 2020-03-02
Fix 'By systems' N/A score

dLobatog:


src/store/Reducers/SystemStore.js
```diff
@@ -9,20 +9,18 @@ import {
     complianceScoreString
 } from 'PresentationalComponents';
 
-const findProfiles = (system, defaultReturn, profileIds) => {
-    if (system.profiles === undefined) { return defaultReturn; }
```
dLobatog: This return is useless, it was never being called, it's just leftovers from when I was working on this PR. Besides if it actually returned 'Never', '0' or such values, the whole reducer would fail because `"Never".map` is undefined

src/store/Reducers/SystemStore.js
```diff
@@ -9,20 +9,18 @@ import {
...
-
+export const findProfiles = (system, profileIds) => {
     let profiles = system.profiles;
 
-    if (profileIds !== undefined && profileIds !== '') {
+    if (profileIds !== undefined && profileIds.toString() !== '') {
```
dLobatog: Perhaps it's a bit strange, but in JS, `[''] !== ['']`, so just to check that the array just contains an empty string, this is safer

PR #413, opened 2020-03-06
 RHICOMPL-502 - Save list of selected rules to tailor

dLobatog:
:warning:  Requires https://github.com/RedHatInsights/frontend-components/pull/392 to work
Screenshot from 2020-03-06 14-01-27

Still in draft until the frontend-components part is merged

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicyRules.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyRules.js
@@ -44,7 +44,7 @@ const columns = [
     { title: <React.Fragment>{ ANSIBLE_ICON } Ansible</React.Fragment>, transforms: [sortable], original: 'Ansible' }
 ];
 
-export const EditPolicyRules = ({ profileId, benchmarkId }) => {
+export const EditPolicyRules = ({ profileId, benchmarkId, dispatch }) => {
     const { data, error, loading } = useQuery(QUERY, { variables: { profileId, benchmarkId } });
 
     if (error) { return error; }
@@ -60,6 +60,16 @@ export const EditPolicyRules = ({ profileId, benchmarkId }) => {
             selectedFilter
             columns={columns}
             loading={loading}
+            handleSelect={((selectedRules) => {
+                dispatch({
+                    type: '@@redux-form/CHANGE',
+                    meta: {
+                        field: 'selectedRules',
+                        form: 'policyForm'
+                    },
+                    payload: selectedRules
+                });
+            })}
             profileRules={ !loading && [{
                 profile: { refId: data.profile.refId, name: data.profile.name },
                 rules: data.benchmark.rules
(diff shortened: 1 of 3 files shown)


src/SmartComponents/CreatePolicy/EditPolicyRules.js
```diff
@@ -10,7 +10,18 @@ import propTypes from 'prop-types';
 import { useQuery } from '@apollo/react-hooks';
 
 const QUERY = gql`
-query Profile($profileId: String!){
+query benchmarkAndProfile($benchmarkId: String!, $profileId: String!){
+    benchmark(id: $benchmarkId) {
```
akofink: If we're passing profile ID, why would we need to also pass benchmark ID?

src/SmartComponents/CreatePolicy/EditPolicyRules.js
```diff
@@ -33,35 +44,43 @@ const columns = [
...
-export const EditPolicyRules = ({ profileId }) => {
-    const { data, error, loading } = useQuery(QUERY, { variables: { profileId } });
+export const EditPolicyRules = ({ profileId, benchmarkId }) => {
+    const { data, error, loading } = useQuery(QUERY, { variables: { profileId, benchmarkId } });
 
     if (error) { return error; }
 
     if (loading) { return <EmptyTable><Spinner/></EmptyTable>; }
 
+    // Figure out pre-selected rules by computing the list of data.benchmark.rules - data.profiles.rules
```
akofink: Where does this happen? I don't see any subtraction in the code. Is this comment still relevant?

src/SmartComponents/CreatePolicy/EditPolicyRules.test.js
```diff
@@ -12,9 +12,16 @@ jest.mock('@redhat-cloud-services/frontend-components-inventory-compliance', ()
 
 describe('EditPolicyRules', () => {
     it('expect to render without error', () => {
```
akofink: How about a test for showing the expected rules?

src/SmartComponents/CreatePolicy/EditPolicyRules.js
dLobatog: It doesn't happen anymore - I'll remove it

src/SmartComponents/CreatePolicy/EditPolicyRules.test.js
dLobatog: There are a few of such tests in https://github.com/RedHatInsights/frontend-components/pull/392

src/SmartComponents/CreatePolicy/EditPolicyRules.js
dLobatog: Because these are different GraphQL queries - the first query just finds the benchmark. I could do 
```
profile {
   benchmark { 
      rules {
      }
   }
}
```
but with 3 levels of nesting, the queries get noticeably slower

akofink: Ok - seems like a bug in our GraphQL API that the response would be slower despite returning the same data, but we can maybe investigate that later.

dLobatog:
From what I see, redux-form is still saving the right values when you uncheck all - but it's not right in the UI, I'll check what's going on

PR #428, opened 2020-03-13
Minor changes and polish ErrorPage

bastilian:
Changes based on feedback from UX (aka. @katierik). Icon subject to change.
Screenshot 2020-03-13 at 16 02 12

### DIFF
--- a/src/PresentationalComponents/ErrorPage/ErrorPage.js
+++ b/src/PresentationalComponents/ErrorPage/ErrorPage.js
@@ -6,11 +6,11 @@ import {
     Text
 } from '@patternfly/react-core';
 import {
-    ErrorCircleOIcon
+    WarningTriangleIcon
 } from '@patternfly/react-icons';
 import {
     InvalidObject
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/ErrorPage/ErrorPage.js
```diff
@@ -6,11 +6,11 @@ import {
     Text
 } from '@patternfly/react-core';
 import {
-    ErrorCircleOIcon
+    WarningTriangleIcon
 } from '@patternfly/react-icons';
 import {
     InvalidObject
-} from '@redhat-cloud-services/frontend-components';
+} from '@redhat-cloud-services/frontend-components/components/InvalidObject';
```
bastilian: Importing components (from frontend-components) like this will help us reduce our bundle size.

bastilian:
@akofink Just wanted to hold off on it for confirmation on the icon from @katierik, that was/is not 100%.

PR #444, opened 2020-03-26
RHICOMPL-292 - Only display systems with reports on Reports by System

dLobatog:
:warning: Depends on https://github.com/RedHatInsights/compliance-backend/pull/365


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -93,10 +93,14 @@ class SystemsTable extends React.Component {
     }
 
     systemFetch = () => {
-        const { client } = this.props;
+        const { client, showOnlySystemsWithTestResults } = this.props;
         const { policyId, perPage, page, activeFilters } = this.state;
         let filter = this.filterBuilder.buildFilterString(activeFilters);
 
+        if (showOnlySystemsWithTestResults) {
+            filter = `has_test_results = true and ${filter}`;
```
akofink: Do we need parentheses? Imagine the case where the filter is currently `FOO or BAR`, then this would result in `has_test_results = true and FOO or BAR`, interpreted as `(has_test_results = true and FOO) or BAR` instead of `has_test_results = true and (FOO or BAR)`

```suggestion
            filter = `has_test_results = true and (${filter})`;
```

src/SmartComponents/SystemsTable/SystemsTable.test.js
```diff
@@ -67,7 +67,7 @@ describe('SystemsTable', () => {
 
     it('expect to not render a loading state', () => {
         const wrapper = shallow(
-            <SystemsTable { ...defaultProps }  items={ items.data.allSystems } systemsCount= { 1 } />
+            <SystemsTable { ...defaultProps }  items={ items.data.systems } systemsCount= { 1 } />
```
akofink: We need to make sure this doesn't land before the backend change that removes allSystems.

dLobatog: This is actually leftover code that's not used anywhere - the SystemsTable component uses the Relay endpoint `systems` - `items.data.allSystems` is nil in here. It's defined in line 21

PR #447, opened 2020-03-30
RHICOMPL-508 - Display special message if no policies are assigned

dLobatog:


src/SmartComponents/AssignPoliciesModal/SubmitPoliciesButton.js
```diff
@@ -7,13 +7,24 @@ import { addNotification } from '@redhat-cloud-services/frontend-components-noti
...
 import { reset } from 'redux-form';
 
-const SubmitPoliciesButton = ({ policyIds, systemId, toggle, dispatch }) => {
+export const completedMessage = (system) => {
+    const profiles = system.profiles.map((profile) => profile.name);
+    let message;
+    if (profiles.length > 0) {
+        message = `Associated ${profiles} to ${system.name}`;
+    } else {
+        message = `${system.name} is now not associated to any policy`;
```
akofink: This phrasing seems awkward. Does it come from the mocks somewhere? @katierik please advise.

katierik: This is telling the user what policies a given system is associated with, right?
What has the user just done?

Assuming that the user has changed the policies and now is getting a confirmation message...
First one - "Associated [profiles] to [system]" is decent
Second one is more awkward. I would suggest "[System] is not associated to any policies."
Top message looks decent - "Associated [profiles] to [system]"
The message for no association is awkward. Propose changed to - [system]

akofink: The user has just edited the associated policies for this host:

Peek 2020-03-30 15-11

These will be displayed in the green popup after clicking save.

katierik: Sounds good.

PR #450, opened 2020-03-30
RHICOMPL-306 - Fix pagination for all inventory hosts

dLobatog:


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -89,7 +89,14 @@ class SystemsTable extends React.Component {
     }
 
     onRefresh = ({ page, per_page: perPage }) => {
-        this.setState({ page, perPage }, this.systemFetch);
+        const { showAllSystems } = this.props;
+        if (showAllSystems && this.inventory && this.inventory.current) {
```
dLobatog, line 93: I tried to do this as this, but `this.inventory` is not available within that context:
```
this.setState(
            { page, perPage },
            () => {
                if (showAllSystems && this.inventory && this.inventory.current) {
                    this.inventory.current.onRefreshData({ page, perPage });
                } else {
                    this.systemFetch();
                }
            }
        );
```

PR #452, opened 2020-03-30
RHICOMPL-525 - Empty state for 'assign policy modal'

dLobatog:
Screenshot from 2020-03-30 10-04-01

The icon should show up all the time, even if the text is truncated


akofink:
@dLobatog your description on this PR looks like it was for a different change

akofink:
Also, this verbiage is grammatically incorrect in that it ends in a preposition.

> Choose which policies rhel7-insights-client.virbr0.akofink-laptop should be a part of. Note: Only policies for this host operating system are shown.

This is how you're supposed to write it:

> Choose policies of which rhel7-insights-client.virbr0.akofink-laptop should be a part. Note: Only policies for this host operating system are shown.

This (or similar) sounds much less awkward:

> Choose policies assigned to rhel7-insights-client.virbr0.akofink-laptop. Note: Only policies for this host operating system are shown.

@katierik please advise.
@dLobatog we can create a new card for this update; this PR is good for merge IMO.

PR #453, opened 2020-03-30
RHICOMPL-468 - Only show Reports as profiles with test results

dLobatog:
:warning:  Requires https://github.com/RedHatInsights/compliance-backend/pull/371


src/SmartComponents/Reports/Reports.js
```diff
@@ -19,19 +19,24 @@ import {
...
-        refId
-        description
-        totalHostCount
-        compliantHostCount
-        majorOsVersion
-        complianceThreshold
-        businessObjective {
-            id
-            title
+    profiles(search: "has_test_results = true", limit: "1000"){
```
akofink: Limit needs to be an integer

```suggestion
    profiles(search: "has_test_results = true", limit: 1000){
```

PR #470, opened 2020-04-02
Revert "Add checkbox to optionally remove all test results aka. repor…

dLobatog:
…ts (#392)"

This reverts commit 2571724bda570c7d3e13be7addf00dcdcb2261e0.

Since soft-deleting policies is something we're not doing for our current goal, we are now just deleting policies *and* test results always. I reverted the commit and included 'deleteAllTestResults: true'.

### DIFF
--- a/src/SmartComponents/DeletePolicy/DeletePolicy.js
+++ b/src/SmartComponents/DeletePolicy/DeletePolicy.js
@@ -1,19 +1,16 @@
 import {
     Modal,
     TextContent,
-    Button,
-    Checkbox
+    Button
 } from '@patternfly/react-core';
 import propTypes from 'prop-types';
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 import { useMutation } from '@apollo/react-hooks';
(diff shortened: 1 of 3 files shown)


src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -50,8 +43,7 @@ const DeletePolicy = ({ isModalOpen, policy, toggle, onDelete }) => {
                 <Button key='destroy'
                     aria-label="delete"
                     variant='danger'
-                    onClick={() => deletePolicy({ variables: { input: { id, deleteAllTestResults } } })}
-                >
+                    onClick={() => deletePolicy({ variables: { input: { id, deleteAllTestResults: true } } })}>
```
bastilian, line 46: Should we revert the change in the backend that went along with this and have this behaviour by default in the backend?

dLobatog: Agreed, I'll do that - we can always get it back by re-reverting

PR #472, opened 2020-04-03
RHICOMPL-461 - Breadcrumbs in beta env stay in beta

dLobatog:
Before this change, clicking on the breadcrumbs will always lead you to
the stable version of the app. However, if you are in -beta, you should
remain in beta when clicking on them.


dLobatog:
@akofink Yup, there was a bug recently opened against insights-chrome because of that, one day we should look into what defines what's highlighted, because it's not looking like it's the URL

PR #477, opened 2020-04-04
Update cache-settings for Reports page

dLobatog:
The Reports page is not updating properly when you do stuff such as
removing a policy, or clicking back and forth between the vertical
navigation elements. Instead, we can use cache-and-network which returns
the cache elements first - and makes a network call in the backend,
which is only used to populate the data if it's different to what was in
the cache.

### DIFF
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -32,7 +32,7 @@ const QUERY = gql`
 `;
 
 export const CompliancePolicies = () => {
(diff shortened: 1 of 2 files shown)


bastilian:
In the longterm, we should look into a more coordinated way of keeping data in the UI up to date. As far as i understand `cache-and-network`[1] increases the requests for these queries. 

`useMutation` does allow providing an update function that can be used to update data of queries in cache and only query if needed and keep requests to the backend to a minimum.[2] 

[1] https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980
[2] https://www.apollographql.com/docs/react/data/mutations/#making-all-other-cache-updates

PR #478, opened 2020-04-04
Style OS selection buttons

dLobatog:
Screenshot from 2020-04-04 21-05-05


bastilian:
@dLobatog Works nicely! There is a bit of styling missing, when selected:
Screenshot 2020-04-06 at 13 25 41

and very minor: no hyphens.

PR #481, opened 2020-04-06
Fix padding issue with ComplianceTabs

bastilian:


bastilian:
@dLobatog The selector is now more specific and should hopefully overrule any other definitions. at least locally it the top rule.

PR #483, opened 2020-04-07
Clear wizard values when closing it

dLobatog:
Screencast_2020年04月07日_09時13分53秒 webm

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreatePolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreatePolicy.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Button, Wizard } from '@patternfly/react-core';
-import { formValueSelector } from 'redux-form';
+import { formValueSelector, destroy } from 'redux-form';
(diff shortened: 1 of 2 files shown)


src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -23,9 +23,10 @@ class CreatePolicy extends React.Component {
 
     toggleOpen = () => {
         const { isOpen } = this.state;
+        const { dispatch } = this.props;
         this.setState({
             isOpen: !isOpen
-        });
+        }, () => dispatch(destroy('policyForm')));
```
dLobatog, line 29: I tried calling `reset('policyForm')` here to no avail, this is equivalent for our purposes

bastilian:
@dLobatog I tested it, but there seems to be something fishy. Some values in state seem to remein and mark steps as "done", but when you go to them they error out: 

Kapture 2020-04-07 at 9 32 19

dLobatog:
Thanks @bastilian  - good catch, I'll fix it up

PR #484, opened 2020-04-07
Fetch Inventory when ComplianceThreshold is changed

dLobatog:
The previous call to `systemFetch` is worthless here, it just makes a call to GraphQL but it doesn't refresh the inventory. To refresh the Compliance Icons (:red_circle: to :green_heart: or viceversa), we need the inventory to be reloaded and trigger LOAD_ENTITIES_FULFILLED in the reducer, which in turns regenerates the icons. 

It'd be a better option to dispatch a new action which explicitly just updates the icons but nothing else.

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -85,7 +85,7 @@ class SystemsTable extends React.Component {
 
     componentDidUpdate = (prevProps) => {
         if (prevProps.complianceThreshold !== this.props.complianceThreshold) {
-            this.systemFetch();
+            this.fetchInventory();
         }
     }
 


bastilian:
@dLobatog where are you seeing this issue? This is **without** this patch on policy details on the systems tab and the icons are updated as expected:

update_systems_table

bastilian:
@akofink it doesn't break anything, but it at least doubles the amount of requests (2x to our graphql endpoint and 4 requests to the inventory API) made as far as i can see.

akofink:
Hmm.. I don't see the quadruple requests to inventory. On the normal compliance/systems page, with or without this change, I see double requests to inventory, though.

image

bastilian:
With this applied this is what i am seeing (notice the requests to the right):

Kapture 2020-04-07 at 19 51 07

It might be hard to identify, but (GH gif file size limit), there are two distinct pairs of requests made to the inventory API, above are two graphql requests.

bastilian:
Can this be closed? The compliance threshold is currently not shown for now.
https://github.com/RedHatInsights/compliance-frontend/pull/565 will improve loading and refreshing of the SystemsTable overall as well.

PR #485, opened 2020-04-07
Allow editing ref_id in policy creation wizard

akofink:
Not to be merged until we're sure we want this.


dLobatog:
:warning: A reminder, if we do this now, the extenral profile won't be able to know which "policy" does it come from

akofink:
Closing for now - we will most likely revisit this in the future.

PR #446, opened 2020-03-28
Use PrimaryToolbar for PoliciesTable

bastilian:
This changes the custom toolbar for the `PoliciesTable` to use the `PrimaryToolbar`.

Screenshot 2020-03-28 at 23 24 05

Screenshot 2020-03-28 at 23 23 52

It also removes the filter utilities to use the modules exported from `frontend-components-inventory-compliance` in https://github.com/RedHatInsights/frontend-components/pull/425

**Note:** Tests will fail until  https://github.com/RedHatInsights/frontend-components/pull/425 is merged and a new package version is released.

PR #487, opened 2020-04-08
RHICOMPL-555 - Show empty state only if no policies were created

dLobatog:
If no policies are available in the account, we should show the empty
state. Otherwise, we should show the empty state for the table if just
the filter makes it show no policies.

Screencast_2020年04月08日_12時25分37秒 webm

### DIFF
--- a/src/SmartComponents/PoliciesTable/__snapshots__/PoliciesTable.test.js.snap
+++ b/src/SmartComponents/PoliciesTable/__snapshots__/PoliciesTable.test.js.snap
@@ -1,5 +1,375 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`PoliciesTable expect not to render emptystate if rows are empty but initial policies were not 1`] = `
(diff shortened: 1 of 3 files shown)


src/SmartComponents/PoliciesTable/__snapshots__/PoliciesTable.test.js.snap
```diff
@@ -1,5 +1,375 @@
...
+      onSetPage={[Function]}
+      page={1}
+      perPage={10}
+      variant="bottom"
+    />
+  </TableToolbar>
+</Fragment>
+`;
+
+exports[`PoliciesTable expect not to render emptystate if rows are empty but initial policies were not 2`] = `
```
dLobatog, line 254: Notice in the 2nd snap, we still show the table, not the whole ComplianceEmptyState

dLobatog:
Closing as https://github.com/RedHatInsights/compliance-frontend/pull/446 fixed it

PR #488, opened 2020-04-08
Sort the benchmarks by RefID before displaying them 

dLobatog:


src/SmartComponents/CreatePolicy/__snapshots__/CreatePolicy.test.js.snap
```diff
@@ -300,19 +300,27 @@ exports[`CreatePolicy expect to render the wizard 1`] = `
                             <button
                               class="pf-c-button pf-m-tertiary wizard-os-button "
                               type="button"
-                            />
+                            >
```
dLobatog, line 303: We were missing the refIds entirely from the fixtures file, so this is :+1:

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -79,7 +79,7 @@ const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
                     isRequired
                     fieldId="benchmark">
                     <br/>
-                    { benchmarks && benchmarks.map((benchmark) => {
+                    { benchmarks && benchmarks.sort((a, b) => a.refId.localeCompare(b.refId)).map((benchmark) => {
```
akofink, line 82: This is _probably_ safe. It'd only take a small change to future refId formats to cause a bug. IMO it'd be better to do something like `.match(/\d/)[0]`

PR #497, opened 2020-04-13
RHICOMPL-562 - View in inventory leads to 'insights/inventory'

dLobatog:


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -243,7 +243,7 @@ class SystemsTable extends React.Component {
                     title: 'View in inventory',
                     onClick: (_event, _index, { id }) => {
                         const beta = window.location.pathname.split('/')[1] === 'beta';
-                        window.location.href = `${window.location.origin}${beta ? '/beta' : ''}/rhel/inventory/${id}`;
+                        window.location.href = `${window.location.origin}${beta ? '/beta' : ''}/insights/inventory/${id}`;
```
akofink, line 246: This should be in constants IMO. Also related: https://github.com/RedHatInsights/compliance-frontend/pull/501

bastilian: Another idea would be to have a function that gives back the correct path, similar to the `..._path` helpers in Rails. There we can prepend the propper prefix(es) for all routes.

Something like `path_for('ROUTEKEY', id)`, where `ROUTEKEY` is the object prop from `paths` in `routes.js` and the following arguments replace an `:id` in the path string.

dLobatog:
I'm preparing a PR to move 'insights' to a constant

PR #498, opened 2020-04-13
RHICOMPL-568 - Pass only selected hosts to Remediations

dLobatog:
Screencast_2020年04月13日_11時11分33秒 webm

Before, we were passing literally all hosts, regardless of selections to the button (see the JIRA card)

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -208,7 +208,7 @@ class SystemsTable extends React.Component {
     }
 
     render() {
-        const { remediationsEnabled, compact, enableExport, showAllSystems, showActions } = this.props;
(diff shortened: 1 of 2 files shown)


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -268,7 +268,7 @@ class SystemsTable extends React.Component {
                 { remediationsEnabled &&
                     <reactCore.ToolbarItem style={{ marginLeft: 'var(--pf-global--spacer--lg)' }}>
                         <ComplianceRemediationButton
-                            allSystems={ items.map((edge) => edge.node) }
+                            allSystems={ items.map((edge) => edge.node).filter(system => selectedEntities.includes(system.id)) }
```
dLobatog, line 271: Notice before, we were sending *all* of the systems to the button, we should only be sending the selected ones, like we do for the JSON/CSV button

PR #504, opened 2020-04-14
RHICOMPL-571 - Show tooltip close to threshold icon

dLobatog:
Screenshot from 2020-04-14 09-14-40

### DIFF
--- a/src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
+++ b/src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
@@ -23,33 +23,31 @@ const PolicyDetailsDescription = ({ policy }) => (
         </CardHeader>
         <CardBody>
             <TextContent>
-                <Tooltip
-                    position='right'
-                    content={
+                <Text component={TextVariants.h5}>
+                    Compliance threshold
+                    <Tooltip
+                        position='right'
+                        content={
+                            <span>
+                                The threshold for compliance is a value set by your organization for
+                                each policy.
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
```diff
@@ -23,33 +23,31 @@ const PolicyDetailsDescription = ({ policy }) => (
...
-                        </Text>
-                    </span>
-                </Tooltip>
-                <React.Fragment>
-                    <Text component={TextVariants.h5}>Business objective</Text>
-                    <Text component={TextVariants.p}>
-                        { policy.businessObjective && policy.businessObjective.title || '-' }
-                    </Text>
-                </React.Fragment>
+                        </span>
```
dLobatog, line 41: I know the span tags may seem strange but `Tooltip` doesn't pick the children HTML up unless you wrap it in a real HTML tag (no React.Fragment)

src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
```diff
@@ -23,33 +23,31 @@ const PolicyDetailsDescription = ({ policy }) => (
         </CardHeader>
         <CardBody>
             <TextContent>
-                <Tooltip
-                    position='right'
-                    content={
+                <Text component={TextVariants.h5}>
+                    Compliance threshold
+                    <Tooltip
```
dLobatog, line 28: The only change here is just to be more specific at applying the tooltip specifically to the icon

PR #507, opened 2020-04-14
RHICOMPL-575 - AssignPoliciesModal only lists policies matching OS

dLobatog:
For a host which has system_profile empty:
Screencast_2020年04月14日_21時02分32秒 webm

For a host that actually has the os_release in system_profile: 
Screencast_2020年04月14日_20時59分27秒 webm

### DIFF
--- a/src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
+++ b/src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import propTypes from 'prop-types';
 import { CheckboxGroup, ErrorPage } from 'PresentationalComponents';
 import {
@@ -13,6 +13,7 @@ import gql from 'graphql-tag';
 import { reset, reduxForm, formValueSelector } from 'redux-form';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
+import { getSystemProfile } from '../../Utilities/InventoryApi';
 
 const QUERY = gql`
 query systemWithProfiles($search: String!){
@@ -30,6 +31,7 @@ query systemWithProfiles($search: String!){
         nodes {
             id
             name
+            majorOsVersion
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/SmartComponents/AssignPoliciesModal/__snapshots__/AssignPoliciesModal.test.js.snap
```diff
@@ -1757,3 +1757,920 @@ exports[`AssignPoliciesModal expect to render an open modal without error 1`] =
...
+                                        checked=""
+                                        class="pf-c-check__input"
+                                        id="1"
+                                        name="policies[0]"
+                                        type="checkbox"
+                                        value="1"
+                                      />
+                                      <label
+                                        class="pf-c-check__label"
+                                        for="1"
```
dLobatog, line 2118: :warning: Notice here ONLY profile 1 is actually shown. This is *right* , because profile 2 is from RHEL 7.

src/SmartComponents/AssignPoliciesModal/__snapshots__/AssignPoliciesModal.test.js.snap
```diff
@@ -1757,3 +1757,920 @@ exports[`AssignPoliciesModal expect to render an open modal without error 1`] =
...
+                                </div>
+                                <br />
+                                <div
+                                  class="pf-c-content"
+                                >
+                                  <p
+                                    class=""
+                                    data-pf-content="true"
+                                  >
+                                    No eligible policies for this system.
```
dLobatog, line 2216: :warning: No policies are eligible because in the test, all policies are rhel7 and the system_profile response contains rhel 8

src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.test.js
```diff
@@ -53,30 +69,91 @@ describe('AssignPoliciesModal', () => {
                 profiles: {
                     nodes: [{
                         id: '1',
-                        name: 'profile1'
+                        name: 'profile1',
+                        majorOsVersion: '8'
+                    }, {
```
dLobatog, line 74: Notice I added a profile here with RHEL7 - and the snapshot didn't change, this is correct because we don't want the RHEL7 one to show up for a RHEL8 host.

src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.test.js
```diff
@@ -1,10 +1,23 @@
 import toJson from 'enzyme-to-json';
 import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
+import { act } from 'react-dom/test-utils';
 import { useQuery, useMutation } from '@apollo/react-hooks';
 
+const waitForComponentToPaint = async (wrapper: any) => {
```
dLobatog, line 7: Not particularly proud of this one but the wrapper takes a little bit of time to load, and `act` doesn't play well with enzyme. I tried many options to test it but I couldn't get it to load the component *after* the useEffect.

 https://github.com/enzymejs/enzyme/issues/2073 gave me the idea

src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
```diff
@@ -40,15 +42,33 @@ const isSystemAssignedToProfile = (system, id) => (
...
-    if (error) { return <ErrorPage error={error}/>; }
+    if (error || errorSystemOs) { return <ErrorPage error={error}/>; }
 
-    if (loading) { return <React.Fragment/>; }
+    if (loading || loadingSystemOs) { return <React.Fragment/>; }
 
     const system = data.systems.nodes.length === 1 && data.systems.nodes[0];
-    const options = data.profiles.nodes.map((policy) => (
+    const options = data.profiles.nodes.filter(
+        policy => (!systemOs || policy.majorOsVersion === systemOs)
```
dLobatog, line 70: Some systems don't have a system_profile - in that case, we want not to filter by anything.

src/SmartComponents/AssignPoliciesModal/AssignPoliciesModal.js
```diff
@@ -13,6 +13,7 @@ import gql from 'graphql-tag';
 import { reset, reduxForm, formValueSelector } from 'redux-form';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
+import { getSystemProfile } from '../../Utilities/InventoryApi';
```
bastilian, line 16: There should be no need for the `../../`

PR #508, opened 2020-04-14
Conditionally link to policies if not external

dLobatog:
Screenshot from 2020-04-15 10-12-11
Screenshot from 2020-04-15 10-12-00
Screenshot from 2020-04-15 10-39-09

In addition to the changes in the views:

* Assign Policies modal is now restricted to only policies that are NOT external.
* SCAP policies table is now restricted to only policies that are NOT external


src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -65,11 +66,16 @@ query Profile($policyId: String!){
...
 
 export const PolicyDetailsQuery = ({ policyId, onNavigateWithProps }) => {
-    const { data, error, loading, refetch } = useQuery(QUERY, {
+    let { data, error, loading, refetch } = useQuery(QUERY, {
         variables: { policyId }
     });
     const [activeTab, setActiveTab] = useState(0);
     let policy = data && !loading ? data.profile : {};
+    if (policy.external) {
+        error = { message: 'This is an external policy.' };
```
dLobatog, line 75: @bastilian Is this right? Without setting `loading = true`, the bottom part of the StateView still shows the data from the external policy. It's not like anyone should get to this page though - you can only get here if you get the policy ID and put it in the URL bar directly

bastilian: If you only want to show the error state part then both `data` and `loading` need to be `undefined`, since that is what `useQuery` would return in that case.

dLobatog: Works great, thanks!

src/SmartComponents/CompliancePolicies/CompliancePolicies.js
```diff
@@ -15,7 +15,7 @@ import PoliciesTable from '../PoliciesTable/PoliciesTable';
 
 const QUERY = gql`
 {
-    profiles {
+    profiles(search: "external = false") {
```
akofink, line 18: This should be the default, right? (Or we should make it the default)

dLobatog: It's not the default currently

PR #512, opened 2020-04-16
RHICOMPL-589 - Add tooltip to show SSG version on wizard

dLobatog:
:warning: There's some error when rendering the test currently

Screenshot from 2020-04-16 11-41-41

### DIFF
--- a/src/SmartComponents/CreatePolicy/__snapshots__/CreateSCAPPolicy.test.js.snap
+++ b/src/SmartComponents/CreatePolicy/__snapshots__/CreateSCAPPolicy.test.js.snap
@@ -202,268 +202,1775 @@ Array [
 `;
 
 exports[`CreateSCAPPolicy should render policies from the selected benchmark only 1`] = `
-Array [
-  <div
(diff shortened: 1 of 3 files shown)


bastilian:
The ` Error: Uncaught [Error: tippy() was passed a plain object which is no longer supported as an
    argument.` is really odd. It only seems to be an issue within the tests but not when rendered in the UI/browser.

src/SmartComponents/CreatePolicy/__snapshots__/CreateSCAPPolicy.test.js.snap
```diff
@@ -202,554 +202,485 @@ Array [
...
+    >
+      <small
+        className=""
+        data-pf-content={true}
+      >
+        SCAP Security Guide (SSG): 
+        Guide to the Secure Configuration of Red Hat Enterprise Linux 6
+         - 
+        0.1.28
+        <Tooltip
```
dLobatog, line 420: Here's the key!

PR #513, opened 2020-04-16
RHICOMPL-586 - Do not load rules if there's no remediation button

dLobatog:
This improves the performance on the Systems table (the one in beta with all systems) and the same table in the CreatePolicy wizard. It doesn't make sense to pull the rules, especially when that is an expensive operation, as they are useless in this context without the Remediation button. The table is slow when the Remediation button is used but I'm working on fixing that as well (likely on the backend).

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -55,6 +55,29 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
 }
 `;
 
+export const GET_SYSTEMS_WITHOUT_FAILED_RULES = gql`
+query getSystems($filter: String!, $perPage: Int, $page: Int) {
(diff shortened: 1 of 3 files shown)


dLobatog:
Agreed - I was looking a bit into it but so far this worked

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -55,6 +55,29 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
 }
 `;
 
+export const GET_SYSTEMS_WITHOUT_FAILED_RULES = gql`
```
bastilian, line 58: You could wrap `GET_SYSTEMS` in a function with an `includeRulesFailed` argument and have the `ruleObjectsFailed` part only included if true. It would also allow to pass `remediationsEnabled` directly to `GET_SYSTEMS(remediationsEnabled)` in the query option.

PR #514, opened 2020-04-16
Remove remediate with ansible button from systems tab

dLobatog:


src/SmartComponents/PolicyDetails/PolicySystemsTab.js
```diff
@@ -4,6 +4,7 @@ import propTypes from 'prop-types';
 
 const PolicySystemsTab = ({ policy, complianceThreshold }) => (
     <SystemsTable policyId={policy.id}
+        remediationsEnabled={false}
```
akofink, line 7: I'm confused by this prop - it makes it sound like remediations were already disabled here, but in testing, it works as expected.

dLobatog: It's true by default

akofink: Oh, sorry lol - I saw this as a deletion rather than an addition. Don't mind me! :+1:

PR #519, opened 2020-04-16
Changes to report card to display SSG from policy

dLobatog:
This commit changes the Report card to:

 - display SSG on a separate line
 - show the policy info on the popover
 - link to the policy ID on view policy if availabble

### DIFF
--- a/src/PresentationalComponents/ReportCard/PolicyPopover.js
+++ b/src/PresentationalComponents/ReportCard/PolicyPopover.js
@@ -13,7 +13,7 @@ import {
 import propTypes from 'prop-types';
 
 const PolicyPopover = ({ profile }) => {
-    const { policy, complianceThreshold, majorOsVersion, businessObjective } = profile;
+    const { name, policy, complianceThreshold, majorOsVersion, businessObjective } = profile;
     return (
         <Popover
             headerContent={name}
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/ReportCard/PolicyPopover.js
```diff
@@ -12,13 +12,13 @@ import {
 } from 'react-router-dom';
 import propTypes from 'prop-types';
 
-const PolicyPopover = ({ policy }) => {
-    const { name, id, complianceThreshold, majorOsVersion, businessObjective } = policy;
+const PolicyPopover = ({ profile }) => {
```
bastilian, line 15: Should we rename it to ProfilePopover?

akofink: This popover actually contains the policy info, so arguably, we should keep it `PolicyPopover`

bastilian: Right, then let's keep it `PolicyPopover`. I just wanna try to avoid and resolve any confusion. :)

dLobatog: This one particularly made sense now, because we show the policy info here now - I pass the profile but the attributes are really for the policy via delegation

PR #526, opened 2020-04-17
RHICOMPL-528 - All systems table papercuts

dLobatog:
Screencast_2020年04月17日_09時51分46秒 webm

This matches the mocks more accurately :smile:

### DIFF
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -8,25 +8,19 @@ const ComplianceSystems = () => {
         key: 'display_name',
         title: 'Name',
         props: {
-            width: 40
+            width: 40, isStatic: true
         }
     }, {
         key: 'facts.compliance.policies',
         title: 'Policies',
         props: {
-            width: 40
+            width: 40, isStatic: true
         }
     }, {
-        key: 'facts.compliance.compliance_score',
-        title: 'Compliance score',
+        key: 'facts.compliance.details_link',
+        title: '',
         props: {
-            width: 10
-        }
-    }, {
-        key: 'facts.compliance.last_scanned',
-        title: 'Last scanned',
-        props: {
-            width: 10
+            width: 20, isStatic: true
         }
     }];
 
@@ -37,6 +31,7 @@ const ComplianceSystems = () => {
             <PageHeader className={ beta ? 'beta-page-header' : 'stable-page-header' } >
                 <PageHeaderTitle title="Compliance systems" />
                 { !beta && <ComplianceTabs/> }
+                <br/>
             </PageHeader>
             <Main>
                 <SystemsTable showAllSystems remediationsEnabled={false} columns={columns} />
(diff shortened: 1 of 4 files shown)


src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -8,25 +8,19 @@ const ComplianceSystems = () => {
...
         props: {
-            width: 40
+            width: 40, isStatic: true
         }
     }, {
         key: 'facts.compliance.policies',
         title: 'Policies',
         props: {
-            width: 40
+            width: 40, isStatic: true
```
dLobatog, line 17: This removes the sort icon on the title of the table

src/store/Reducers/SystemStore.js
```diff
@@ -107,7 +108,14 @@ export const systemsToInventoryEntities = (systems, entities, showAllSystems, pr
                     display_name: { title: <Link to={{ pathname: `/systems/${matchingSystem.id}` }}>
                         { entity.display_name || matchingSystem.name }
                     </Link> },
-                    policies: matchingSystem.profileNames,
+                    policies: matchingSystem.profileNames || {
+                        title: <Text className='grey-icon'>No policies</Text>
+                    },
+                    details_link: matchingSystem.profileNames && {
```
dLobatog, line 114: I'm adding the `matchingSystem.profileNames` as otherwise we don't want to show this link

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -8,33 +8,30 @@ const ComplianceSystems = () => {
...
     }];
 
     const beta = window.location.pathname.split('/')[1] === 'beta';
 
     return (
         <React.Fragment>
-            <PageHeader className={ beta ? 'beta-page-header' : 'stable-page-header' } >
+            <PageHeader
+                className={ beta ? 'beta-page-header' : 'stable-page-header' }
+                style={ { paddingBottom: '20px' } }
```
dLobatog, line 33: I'm adding this as style purposefully. We have to submit a PR removing all the `beta/stable` differences soon. This change should include a class for the page-header with 20px of padding-bottom, and we should remove the tabs padding-top.

bastilian: Ok,... in what case should there be a padding at the bottom? This will add a padding always, in both beta and stable, with no difference. and there is currently no difference. Looking at the UI it seems there should be a padding when it `beta` since then there are no tabs. Then this 20px padding bottom should be in the css class definition `beta-page-header`, while `stable-page-header` will keep it at 0.

dLobatog: Only if there are no tabs, there should be padding

bastilian: Then the padding-bottom in the css for `beta-page-header` should be set to `20px` and instead of here. This will set padding in both cases.

bastilian: Since this is going away on Monday already, consider this moot. :) I thought this would still be needed for a few more weeks or even longer.

PR #531, opened 2020-04-17
Remove beta checks for page header

dLobatog:
Notice to promote this to each environment we need:

* https://github.com/RedHatInsights/cloud-services-config/pull/176
* https://github.com/RedHatInsights/cloud-services-config/pull/177
* https://github.com/RedHatInsights/cloud-services-config/pull/178


src/App.scss
```diff
@@ -5,16 +5,9 @@
 
 
 .pf-c-page__main .pf-c-page-header.pf-c-page__main-section {
-  &.stable-page-header {
```
dLobatog, line 8: I think this is basically the change that allows us to remove the `style` we talked about on https://github.com/RedHatInsights/compliance-frontend/pull/526  @bastilian

PR #538, opened 2020-04-21
RHICOMPL-606 - Add policy id to rowdata and use for deletion

bastilian:
When filtering looking up the policy with per page, index, etc. returns the wrong policy to delete.
This adds the policy id to the rowdata to later use it to find the correct policy to delete.

### DIFF
--- a/src/SmartComponents/PoliciesTable/PoliciesTable.js
+++ b/src/SmartComponents/PoliciesTable/PoliciesTable.js
@@ -38,6 +38,7 @@ const emptyRows = [{
 const policiesToRows = (policies) => (
     policies.map((policy) => (
         {
+            policyId: policy.id,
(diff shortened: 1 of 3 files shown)


bastilian:
@dLobatog It's a bit refactored now and has a test to verify the function the sets the policy and opens the modal. Testing if the function via the actionResolver does the same when filtering is something we should do on every table that we use filters on. I'll look into having shared tests that we can add to each filtered table.

For now it is verifyied via snapshots that the `policyId` is present in the `rowData`, that the actions are properly set on the table and that when `setAndDeletePolicy` is called with an `policyId` the policy is set and the modal is opened.

dLobatog:
Was able to verify with 2 pages and search - thanks @bastilian !

PR #539, opened 2020-04-21
RHICOMPL-603 - Make compliant filter optional

bastilian:
We aren't showing compliance status and score at the moment, therefore we also will remove the filters for them for now.


bastilian:
@dLobatog This should be better. I missed that table on ReportDetails. The compliant filters are now enable-able, disabled by default and enabled on ReportDetails.

PR #548, opened 2020-04-22
RHICOMPL-597 - Clear inventory filters when loading systems

bastilian:
This will prevent Inventory filters to persist throughout navigating to other pages.

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -210,7 +210,7 @@ class SystemsTable extends React.Component {
     }, 500)
 
     async fetchInventory() {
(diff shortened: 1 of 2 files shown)


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -223,6 +223,8 @@ class SystemsTable extends React.Component {
             pfReactTable
         });
 
+        this.props.clearInventoryFilter();
```
dLobatog, line 226: Just a nitpick, you can get this function on the this.props assignment in line 213, for clarity so we know which props are used in each method easily

bastilian:
@dLobatog Thanks for reviewing! Updated to assign `clearInventoryFilter`.

PR #553, opened 2020-04-23
Add link within EmptyState OpenSCAP Button

dLobatog:
The Button variant='link' from Patternfly just styles the text, but it needs an anchor tag to actually link to somewhere.

### DIFF
--- a/src/PresentationalComponents/CompliancePoliciesEmptyState/CompliancePoliciesEmptyState.js
+++ b/src/PresentationalComponents/CompliancePoliciesEmptyState/CompliancePoliciesEmptyState.js
@@ -33,11 +33,10 @@ const CompliancePoliciesEmptyState = ({ title }) => (
             </EmptyStateBody>
             <CreatePolicy onWizardFinish={() => { location.reload(); }} />
             <EmptyStateSecondaryActions>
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/CompliancePoliciesEmptyState/CompliancePoliciesEmptyState.js
```diff
@@ -33,11 +33,11 @@ const CompliancePoliciesEmptyState = ({ title }) => (
             </EmptyStateBody>
             <CreatePolicy onWizardFinish={() => { location.reload(); }} />
             <EmptyStateSecondaryActions>
-                <Button
-                    variant="link"
-                    href="https://www.open-scap.org/getting-started/">
-                    Learn about OpenSCAP and Compliance
-                </Button>
+                <a target='_blank' rel='noopener noreferrer' href='https://www.open-scap.org/getting-started/'>
+                    <Button variant='link'>
```
bastilian, line 37: You should be able to set the anchor tag as the component of the button according to PF4 docs: https://www.patternfly.org/v4/documentation/react/components/button#links-as-button

dLobatog: Oh good point, thanks!

PR #558, opened 2020-04-24
Create ruleObjectsFailed from rules/compliant fields

dLobatog:
The response ruleObjectsFailed hasn't been reviewed on the backend for a
long time, and it returns the wrong rules. Currently it looks at
historical RuleResults (all of them!) for the system, and retrieves the
failing ones. This is NOT what we're interested on. We're
interested only on whether the rules for the *latest test results for
the current profiles* are compliant or not.


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -37,19 +38,15 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
                 profiles {
                     id
                     name
-                    rulesPassed
```
dLobatog, line 40: This may seem like a stretch, but these 2 are generating N+1 queries all over, as we have to call a SQL count query for every profile/host - after finding the testResult. We *already* have this data in the frontend by means of `rules/compliant`

src/Utilities/ruleHelpers.js
```diff
@@ -1,9 +1,11 @@
+import flatten from 'lodash/flatten';
+
 export const profilesRulesFailed = (profiles) => (
-    profiles.map(profile => profile.rules.filter(rule => !rule.compliant)).flat()
+    flatten(profiles.map(profile => profile.rules.filter(rule => !rule.compliant)))
 );
 
 export const profilesRulesPassed = (profiles) => (
-    profiles.map(profile => profile.rules.filter(rule => rule.compliant)).flat()
+    flatten(profiles.map(profile => profile.rules.filter(rule => rule.compliant)))
```
bastilian, line 8: Why not use flatMap? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -37,19 +38,16 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
...
                     score
-                }
-                ruleObjectsFailed {
-                    refId
-                    profiles {
+                    rules {
                         refId
+                        title
+                        compliant
+                        remediationAvailable
```
dLobatog, line 49: This is used on the ComplianceRemediationButton to determine whether the rule is available or not

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -37,19 +38,16 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
                 profiles {
                     id
                     name
-                    rulesPassed
-                    rulesFailed
+                    refId
```
dLobatog, line 41: I added refId to the profile query because we should pass it to remediations, see https://github.com/RedHatInsights/frontend-components/pull/480

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -308,7 +304,9 @@ class SystemsTable extends React.Component {
                 { remediationsEnabled &&
                     <reactCore.ToolbarItem style={{ marginLeft: 'var(--pf-global--spacer--lg)' }}>
                         <ComplianceRemediationButton
-                            allSystems={ items.map((edge) => edge.node).filter(system => selectedEntities.includes(system.id)) }
+                            allSystems={ systemsWithRuleObjectsFailed(
+                                items.filter(edge => selectedEntities.includes(edge.node.id)
```
dLobatog, line 308: We only want to compute this when `remediationsEnabled`

PR #559, opened 2020-04-24
Add ID to fields to avoid apollo-gql error

dLobatog:
This is meant to prevent https://github.com/apollographql/apollo-client/issues/2510


src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -45,6 +45,7 @@ query Profile($policyId: String!){
         majorOsVersion
         lastScanned
         policy {
+            id
```
dLobatog, line 48: This is the key one, if I add this the PolicyDetails view is "fixed", but I added others where I saw fit

PR #565, opened 2020-05-01
RHICOMPL-638 - Fix SystemsTable loading behaviour

bastilian:
This puts systems into the inventory store, that is updated with fetching systems.
It is also ensured that systems are fetched and updated prior to `fetchInventory` and is called when filters or paginating changes.

To ensure the table is up to date even when items stay the same there is now an `UPDATE_ROWS` action that will get dispatched and recalculate the rows without triggering a inventory refresh.

It also removes debouncing certain calls, which could lead to states being out of sync. 

The primary objective is to fix the policy creation step to assign systems, but since the `SystemsTable` is used in other places here's a list where to click through:

**On each use of the SystemsTable test:**

  * Filter
    * by Name
      * [Known bug] Subsequent name filter, without clearing filters, won't work
    * by Compliance/Score
  * Paginate
    * and filter
    * change per Page
    * click on same per Page count
  * Edit Policies of a system (where enabled)
    * [Known (UX) Bug] Policies previously assigned will be deselected on selecting any policy
  * View in inventory

**Test on:**
  * SCAP Policies -> Create Policy -> Step 4: Systems
  * Reports -> By policy -> View Report
  * Reports -> By policy -> View policy -> Systems | SCAP Policies -> View Policy -> Systems
  * Reports -> By system
  * Systems

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.js
@@ -6,7 +6,7 @@ import {
     TextContent,
     TextVariants
 } from '@patternfly/react-core';
-import { SystemsTableWithApollo as SystemsTable } from '../SystemsTable/SystemsTable';
+import SystemsTable from '../SystemsTable/SystemsTable';
 import { compose } from 'redux';
 import propTypes from 'prop-types';
 import { connect } from 'react-redux';
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.test.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.test.js
@@ -2,6 +2,7 @@ import { policyFormValues } from './fixtures.js';
 import configureStore from 'redux-mock-store';
 import renderer from 'react-test-renderer';
 import { MockedProvider } from '@apollo/react-testing';
+import { Provider as ReduxProvider } from 'react-redux';
 
 const mockStore = configureStore();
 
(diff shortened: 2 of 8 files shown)


bastilian:
@akofink Thanks for testing and reviewing! 

Searching `onChange` is now enabled for the "Name" filter, so far it was disabled, because it will hit the graphql API with every character, which in the past led to undesired behaviour when requests would take long. It should be fine now, both on the backend and frontend. It will still request with each character, but should now always keep the inventory table up to date. Clearing the name filter should now also work as expected. 

Regarding the filtering, I think I'm missing something. Clicking through it here it looks ok to me. Can you elabortate what results you would expect to appear?

akofink:
Notice on the GIF above, the Clear Filters link results in the table showing reporter = insights, even though the filter chip doesn't show that. Clear Filters should show all my systems, without any reporter or staleness filter. Again, the filter chips don't reflect the filter, but the table results make it clear that the filters are still active.

akofink:
I just tested the same on prod, and it does the same thing, so this PR doesn't cause the issue. When you click clear filters on the systems table, the following request is sent: https://cloud.redhat.com/api/inventory/v1/hosts?per_page=50&page=1&order_by=updated&order_how=DESC&staleness=fresh&staleness=stale&&registered_with=insights

See how that has filters on staleness and reporter?

bastilian:
@akofink This might be an issue with the inventory. Looking at the request to the host inventory it seems to always include the `registered_with`. I'm unsure if that is expected, we should verify that.

bastilian:
@akofink I tried using `clearInventoryFilter` when clearing all, but even then the fresh/stale and registered with filters are used for the request to the inventory.

akofink:
Yeah, it's okay. Maybe we can ask platform about this, but it doesn't really affect this PR.

PR #567, opened 2020-05-01
RHICOMPL-543 - Fix CSV export

bastilian:
This fixes JSON and CSV export to export either what is shown in the table or what has been selected and should not contain any systems of other policies.

The selected items are now kept separately in state to avoid loosing the selection when filters are applied or pages are turned and preserve them for export. This also fixes issues elswhere with selecting systems.

It also implements a `exportValue` for cells to handle cells that could contain React elements.


src/Utilities/Export.js
```diff
@@ -51,10 +56,11 @@ export const downloadCsv = (state) => {
     }
 };
 
-export const exportToJson = (selectedEntities) => {
+export const exportToJson = (selectedEntities, page, perPage) => {
+    const offset = perPage * (page - 1);
     const url = (selectedEntities.length > 0) ?
         COMPLIANCE_API_ROOT + '/systems.json' + '?search=(id ^ (' + selectedEntities.join(',') + '))' :
-        COMPLIANCE_API_ROOT + '/systems.json';
+        COMPLIANCE_API_ROOT + `/systems.json?limit=${ perPage }` + (offset > 0 ? `&offset=${ offset }` : '');
```
bastilian, line 63: For some reason, even though the `offset` is in the request the backend fails to give back any results?

src/Utilities/Export.js
```diff
@@ -1,6 +1,7 @@
 import { COMPLIANCE_API_ROOT } from '../constants';
 
 const CSV_FILE_PREFIX = 'compliance-export';
+const CSV_DELIMITER = ';';
```
dLobatog, line 4: Why? AFAIK CSV - 'comma separated values', if I try to import this export into Google Spreadsheets or Libreoffice it fails to recognize the ';'

src/Utilities/Export.js
dLobatog: Notice `offset` is actually like `page` the way it's implemented in the db, so we might have to redo it. For now I guess you can use it like that, pages start at `1`

src/Utilities/Export.js
bastilian: The profilesNames are coma separated and it was a quick solution. I researched a more and it seems we simply need to quote the strings for values. I'll implement that.

PR #576, opened 2020-05-07
RHICOMPL-611 - Refreshing on details view should not redirect

dLobatog:
Before this change, the app was listening to the APP_NAVIGATION event in
redux for all pages. This meant that if you refresh on any details page,
the APP_NAVIGATION event will retrigger and try to redirect to the base
URL for that details page (e.g: /scappolicies for /scappolicies/:policy_id)

Instead, we should only listen to the events we care about, which are
the ones for moving to the vertical navigation items. All other links
through the app, shouldn't have to go through this.

Initially I was going to dismiss the APP_NAVIGATION events completely, as I've seen other apps do, but we can't because of the vertical navigation links, which rely on that. I took inspiration from the [advisor's App.js](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/App.js) to only react to these vertical-navigation link events. 

To test: 
* Ensure moving around the app still works, any links you can think of.
* Refresh on usual pages that didn't work before, confirm that it doesn't take you to anywhere else. Usual suspects include scap policy details, report details, reports by system, system details


akofink:
I also just got 

```
TypeError: INVENTORY_ACTION is undefined ComplianceSystems.js:339660:7
```

akofink:
Also seeing some weird behavior with the error notification. I'm deleting an already-deleted policy here.

Peek 2020-05-07 09-21

src/App.js
```diff
@@ -1,41 +1,48 @@
...
+    const appNavClick = {
+        reports(redirect) { insights.chrome.appNavClick({ id: 'reports', redirect }); },
+        scappolicies(redirect) { insights.chrome.appNavClick({ id: 'scappolicies', redirect }); },
+        systems(redirect) { insights.chrome.appNavClick({ id: 'systems', redirect }); }
+    };
+
+    useEffect(() => {
         insights.chrome.init();
         insights.chrome.identifyApp('compliance');
+        const baseComponentUrl = props.location.pathname.split('/')[1] || 'reports';
```
dLobatog, line 19: Changed this one - it should default to 'reports' if no URL is set

src/App.js
```diff
@@ -1,41 +1,48 @@
...
 import { Routes } from './Routes';
 import { NotificationsPortal } from '@redhat-cloud-services/frontend-components-notifications';
 import '@redhat-cloud-services/frontend-components-notifications/index.css';
 import './App.scss';
 
-class App extends Component {
-    componentDidMount () {
+const App = (props) => {
+    const appNavClick = {
+        reports(redirect) { insights.chrome.appNavClick({ id: 'reports', redirect }); },
```
dLobatog, line 11: This one is problematic. insights-chrome doesn't have an ID for 'reports' - so that's why it failed to show the current page in the sidebar (see @akofink 's screen).  

I have tried to `appNavClick({ id: '', redirect })` and it works! But there are many other applications without an ID - so insights-chrome understands that as "highlight every item without an ID on the sidebar" which highlights Drift comparison for example.

PR #582, opened 2020-05-08
Use NotAuthorized component to handle 403s

akofink:
Requires https://github.com/RedHatInsights/frontend-components/pull/510 to be merged and released and for us to bump the FC version in this repo.

### DIFF
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -54,31 +54,28 @@ export const CompliancePolicies = () => {
         policies = data.profiles.edges.map(profile => profile.node);
     }
 
-    return <StateView stateValues={ { error, data, loading } }>
-        <StateViewPart stateKey='error'>
-            <ErrorPage error={error}/>
-        </StateViewPart>
-        <StateViewPart stateKey='loading'>
-            <PageHeader className='page-header'>
-                <PageHeaderTitle title="SCAP policies" />
-            </PageHeader>
-            <Main>
-                <LoadingPoliciesTable />
-            </Main>
-        </StateViewPart>
-        <StateViewPart stateKey='data'>
-            <PageHeader className='page-header'>
-                <PageHeaderTitle title="SCAP policies" />
-            </PageHeader>
-            <Main>
-                { policies && policies.length === 0 ?
-                    <Grid gutter='md'><ComplianceEmptyState title='No policies'
-                        mainButton={<CreatePolicy onWizardFinish={() => refetch()} />} /></Grid> :
-                    <PoliciesTable onWizardFinish={() => refetch()} policies={ policies } />
-                }
-            </Main>
-        </StateViewPart>
-    </StateView>;
+    return <React.Fragment>
+        <PageHeader className='page-header'>
+            <PageHeaderTitle title="SCAP policies" />
+        </PageHeader>
+        <Main>
+            <StateView stateValues={ { error, data, loading } }>
+                <StateViewPart stateKey='error'>
+                    <ErrorPage error={error}/>
+                </StateViewPart>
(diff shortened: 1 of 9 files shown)


akofink:
This one has changed quite a bit and isn't ready for re-review yet. It's mostly moving titles on our pages up so that they show when we render the ErrorPage. I'll let you know when it's ready for re-review. Thanks!

akofink:
Ok, this one should be ready for re-review.

src/SmartComponents/CompliancePolicies/CompliancePolicies.js
```diff
@@ -54,31 +54,28 @@ export const CompliancePolicies = () => {
...
-                        mainButton={<CreatePolicy onWizardFinish={() => refetch()} />} /></Grid> :
-                    <PoliciesTable onWizardFinish={() => refetch()} policies={ policies } />
-                }
-            </Main>
-        </StateViewPart>
-    </StateView>;
+    return <React.Fragment>
+        <PageHeader className='page-header'>
+            <PageHeaderTitle title="SCAP policies" />
+        </PageHeader>
```
akofink, line 60: This is now at the top level, outside the `StateView`. The header should show even when the page is loading or there is an error.

src/SmartComponents/Reports/Reports.js
```diff
@@ -82,30 +82,27 @@ export const Reports = () => {
...
-                <div className="policies-donuts">
-                    <Grid gutter='md'>
-                        { reportCards }
-                    </Grid>
-                </div>
-            </Main>
-        </StateViewPart>
-    </StateViewWithError>;
+    return <React.Fragment>
+        { pageHeader }
```
akofink, line 86: Same thing here wrt the page title.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -162,6 +167,8 @@ class SystemsTable extends React.Component {
             if (JSON.stringify(newSystems) === JSON.stringify(prevSystems)) {
                 this.props.updateRows();
             }
+        }).catch((error) => {
+            this.setState({ ...this.state, error });
```
akofink, line 171: SystemsTable - now with error handling! I'm not sure if this is the 'right way' to do it, but it works this way with the `StateView` component, similar to other pages. It's a bit complicated because either compliance or inventory could raise an error. Ideally, we'd handle either case.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -321,34 +329,45 @@ class SystemsTable extends React.Component {
             inventoryTableProps.variant = pfReactTable.TableVariant.compact;
         }
 
+        if (error === undefined) {
+            noError = true;
+        }
```
akofink, line 334: This is a bit of a hack to get `StateView` working. We want all the old code for both `loading` or `data` states, since the loading component is handled for us by the `InventoryCmp`.

akofink:
This is a lot easier to review if you 'hide whitespace changes'

image

src/SmartComponents/CompliancePolicies/CompliancePolicies.js
bastilian: Are there mocks for the error pages somewhere. We should confirm this with UX once more. The GDoc linked in Jira does show a screenshot with a header, but it it doesn't explicitly mention it.

akofink: @katierik CC for confirmation that page headers should show during errors. Here's the platform doc we mentioned https://docs.google.com/document/d/1xEy7ieZUCe-R8FI0Lgk61b74npbPuD-PIRdixjsN5ac/edit#bookmark=id.fg1ljx9olsds

PR #588, opened 2020-05-13
RHICOMPL-604 - Add warning when no systems assigned

bastilian:
This adds a warning to the policy creation wizard when no systems are assigned:

Screenshot 2020-05-13 at 13 51 27

@katierik Could you confirm the text colour? In the mocks it is darker, but I'm unable to find a matching colour.


katierik:
The color is one darker: --pf-global--warning-color--200
The 100 version doesn't pass contrast standards for WCAG :(

PR #599, opened 2020-05-28
RHICOMPL-552 - Update rules step in policy wizard

bastilian:
This updates the "Rules" step in the policy creation wizard to further match the mockups.

Kapture 2020-05-28 at 10 52 56

**Test-protocol:**

1. Open the Policy page and the "Create Policy" wizard
2. Fill out steps 1&2 and continue to the "Rules" step.
3. Select or deselect rules
	- The rules count should update
	- A "Restore to default selection" link should appear
		- Clicking should update the selected rules to the default policy rules.
4. Continue to the "Systems" step and return to Rules
	* The rules should still match the previously selected ones
5. Return to the first step and change the policy type
6. Go forward to the "Rules" step
	* Description and default rules should match the newly selected policy type
7. Continue forward and finish the wizard to create a policy with the selected rules.

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -106,7 +106,11 @@ const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
                 </Text>
                 }
                 <FormGroup label="Policy type" isRequired fieldId="policy-type">
-                    <ProfileTypeSelect profiles={selectedBenchmark && validProfiles } />
+                    <ProfileTypeSelect
+                        profiles={selectedBenchmark && validProfiles }
+                        onClick={ () => {
+                            change('selectedRuleRefIds', null);
+                        }}/>
                 </FormGroup>
             </Form>
         </React.Fragment>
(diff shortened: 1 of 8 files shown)


src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -106,7 +106,11 @@ const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
                 </Text>
                 }
                 <FormGroup label="Policy type" isRequired fieldId="policy-type">
-                    <ProfileTypeSelect profiles={selectedBenchmark && validProfiles } />
+                    <ProfileTypeSelect
+                        profiles={selectedBenchmark && validProfiles }
+                        onClick={ () => {
+                            change('selectedRuleRefIds', null);
```
bastilian, line 112: This creates a teeny-tiny issue... When a user returns to the first step and changes the policy, the selected rules need to be invalidated, but we at this point we don't know what rules a policy has in order to select them.

We could make sure that all steps till the rules is invalidated, but the best way, i believe, to fix this is to allow creating a policy with only the policy type and optionally rules to remove or add to the policy types set of rules.

dLobatog: Agreed, to be fair if you go back to the first step and change the policy, also the page where the name is selected is going to be broken and the whole rules page is already broken - RHICOMPL-484 covers that

PR #638, opened 2020-07-05
Pass react redux to inventory loader as well

karelhala:
We are working on inventory refactoring, since we'd like to fully use hooks and such we have to use App's react dependency. With this being used react-redux needs to be passed from application as well. If you are cautious about the size of your bundle I can pick only the functions really required by inventory.


karelhala:
ping @dLobatog

PR #668, opened 2020-07-21
Add config for local chrome

akofink:
Chrome uses the same image with npm installed as the frontend; we just run `npm ci` to clear out the node_modules and install chrome's.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,10 @@
 DEFAULT_HOST=ci.foo.redhat.com
 HOT_RELOAD='false'
 
+# Enable/uncomment this to run chrome locally
+# CHROME_DIR=../insights-chrome
+# LOCAL_CHROME=true
+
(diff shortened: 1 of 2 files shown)


bastilian:
Nice! Thank you @akofink!

There was an issue with getting this up for me. `npm ci` didn't install anything it needed `npm i` for it to work.

Setting up chrome would also be worth mentioning in the README and might be better disabled by default.
It is not required to run the frontend and is only needed when testing upcoming chrome changes.

For using it to verify changes of frontend components that are loaded via the chrome this might not work, or only with a lot of effort. We would need to get anything to link into the context of the container and set it up within. 

At the moment If I do work on chrome or things through chrome, the way I run it is by only mounting the build directory to the proxy, enable `LOCAL_CHROME` and run the build processes outside the container.

PR #676, opened 2020-07-24
Upgrade version of react-redux to newest version

karelhala:
@bastilian


akofink:
Looks like something happened with the build. Could you please force push to re-trigger the build?

bastilian:
Not really sure what is happening TravisCI. I tested it locally and tests pass.

It did throw an error for `clearInventoryFilter` not being a function, we added this to fix an issue with the inventory filters, I can't remember what exactly, but it does seem to work still even without it.

karelhala:
Looks like memory ran out on travis, try increasing it https://medium.com/@vuongtran/how-to-solve-process-out-of-memory-in-node-js-5f0de8f8464c

bastilian:
@karelhala I tried increasing the max old space like suggested in the post in #677, but it doesn't seem to do the trick. :/

Do you have a(nother) hunch what it could be?

karelhala:
I've updated the snapshosts (some needed to be changed because of how react-redux now wraps components), however I can reproduce the memory leaks on my machine as well.

I've investigated further more and looks like those huge snapshots are causing the error, probably becuase of how redux now uses context and that could lead to unwanted memory leaks during test phases while comparing snapshots. I've located 3 tests that were causing this error, 2 were modals, so I just used shallow mode of enzyme to json (we are not really interested in full DOM interpratation here) and the last was Form when creating SCAP Policy. I've looked at the code and tried finding nodes that are relevant for the test - `Form Text small` (rendered with tooltip); `Button` (to check what actually is in each button); `ProfileTypeSelect` (it's the biggest chunk, again we are not really interested into the full DOM).

NOTE: Why not interested in full DOM representation of some components? That's something called dependency injection, you can't relly change how the DOM is represented in these components, we are only intersted in what is passed in these components and realying on these components to behave correctly. There can be usecases to look at the whole DOM representation in integration tests, however these should not be part of unit testing.

bastilian:
@karelhala Thank you for looking further into it!

Our snapshots are huge indeed. I totally agree with not testing third party components. Most tests are only using shallow and snapshots are smaller for those. These bigger snapshots are mostly tests that are still using `react-test-renderer`, that we actually want to phase out to only use one test library (enzyme). (and maybe migrate to React Testing Library at some point).

It seems travis now can pass, but I'll get on it and move those tests to enzyme.

bastilian:
Thank you again @karelhala++!

PR #678, opened 2020-07-27
Pass store down to inventory components

karelhala:


karelhala:
@akofink sure! the snapsho is failing, because we are wrapping he inventory table with new component that passes store down the component chain.

bastilian:
@karelhala I tested it locally with the latest master of chrome and it seems that it can't find `useStore`:

Screenshot 2020-07-27 at 16 12 53

karelhala:
@bastilian try clearing your `node_modules`, looks like you are still using older version of `react-redux`.

Screenshot from 2020-07-27 16-19-21

bastilian:
@karelhala Very odd. I wiped `node_modules` in the chrome repo (I always do) and checked `react-redux` within it to be at `7.2.0`, rebuilt and checked that the proxy caught the change, but even after restarting the proxy, it still had the same issue. It happens as well with proxying to get the chrome from CI.

karelhala:
@bastilian what about in compliance repo? What version of `react-redux` is used there?

bastilian:
@karelhala it's `7.2.0` as well.

PR #680, opened 2020-07-29
RHICOMPL-604 - Only show warning for no systems on details page

bastilian:
This fixes the issue with the no systems warning showing up on other pages then the policy details systems tab.


bastilian:
@akofink Thank you for testing and reviewing this! It now should work like it should have in the first place and only show up in the policy details page when there are no systems assigned and no filter is active.

bastilian:
@akofink At the moment it is hard to distinguish between the resulting items being 0 because it the policy has none or because the current filters have no results. We should revisit this when we (have) refactor(ed) the `SystemsTable` with **E-132**

PR #683, opened 2020-08-05
Pass pf react core to inventory to properly render inventory

karelhala:
There is a new feature brewing in chrome, building chrome with webpack [1]. This PR passes newly required dependencies. In future we'll use Federated modules [2] so there won't be a need for these changes, but this is required as a mid-step solution.

Also, for enabling A/B testing of inventory detail we are working on new inventory component, for now this component will be undefined so use `React.Fragment` as fallback.

[1] https://github.com/RedHatInsights/insights-chrome/pull/842
[2] https://webpack.js.org/concepts/module-federation/


karelhala:
ping @bastilian

bastilian:
@karelhala \o/ awesome! this means we will reduce issues with unaligned package versions, or?

karelhala:
@bastilian this means that inventory will take packages from compliance application, when rendering table or detail. Shared libraries means that we won't be bumping into problems with multiple reacts and importing inventory will be much easier.

Something like (the name of package is not yet decided, the package will not be installed in your package, but rather given trough federated modules)
```JS
import { InventoryTable } from 'hot-loaded-inventory'
```

PR #686, opened 2020-08-07
RHICOMPL-743 - Add tooltip indicating the score is from openscap

bastilian:


bastilian:
Thank you for reviewing @akofink!

PR #685, opened 2020-08-07
[RHICOMPL-742] Update "Edit Policy" modal to allow editing systems

bastilian:
This adds tabs to edit systems, more policy details and in the future rules to the "Edit Policy" modal, which is accessible via the policy overview in each row-action menu, the details page of a policy and via an "Edit systems" button on the systems tab on the policy details.

Screenshot 2020-08-24 at 11 42 34

⚠️ Required backend changes:

 * https://github.com/RedHatInsights/compliance-backend/pull/540

PR #687, opened 2020-08-07
RHICOMPL-666 - Remove "By Policy, by systems" tabs

bastilian:
To be merged after #685 is completed


src/PresentationalComponents/ReportCardGrid/ReportCardGrid.js
```diff
@@ -0,0 +1,35 @@
...
+        title={ 'No policies are reporting' }
+        mainButton={ <BackgroundLink to='/scappolicies/new'>
+            <Button variant='primary'>Create new policy</Button>
+        </BackgroundLink> }
+    />
+);
+
+const ReportCardGrid = ({ policies }) => (
+    policies && policies.length > 0 ?
+        policies.map((policy, i) => (
```
akofink, line 21: I think this should be `profiles` instead of `policies`?

bastilian: Oh, right. will fix right away.

PR #689, opened 2020-08-07
RHICOMPL-667 - Enable policy filter on systems page

bastilian:
This enables filters on the Systems page for policy:
Screenshot 2020-08-31 at 17 10 21

and operating systems:
Screenshot 2020-08-31 at 17 10 31

### DIFF
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -1,8 +1,26 @@
 import React from 'react';
-import { SystemsTable } from 'SmartComponents';
+import gql from 'graphql-tag';
+import { useQuery } from '@apollo/react-hooks';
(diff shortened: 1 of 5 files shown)


bastilian:
@akofink Yeah, here it works fine: 
Kapture 2020-09-01 at 17 30 06

What are the policies look like that are available to filter?

akofink:
Weird, I don't even see a dropdown, just a text box. I'm testing it first with only an external policy:

Peek 2020-09-01 11-45

bastilian:
@akofink Having only external policies or no policies at all should not cause any issues anymore.
PM also clarified that external policies indeed should not be included and shown as filters.

bastilian:
@akofink Yes, this is expected. Since there is only a host with an external policy, which we should not be shown to be filter by, you won't be able to filter by that policy or it's OS major version. If other policies are created, these will then show up and can be used to filter.

PR #690, opened 2020-08-08
Improve routing and allow modal routes

bastilian:
We have certain components that are nested in others, primarily modal components.

This implements a router that allows to be passed an array of routes that can also be set to be a modal route:

```js
{
  path: '/path/to/component',
  component: asyncComponent(() =>
    import(/* webpackChunkName: "Component" */ 'SmartComponents/Component/Component')
  ),
  modal: true
}
```

It also cleans up the use of `routerParams` that has led to unexpected behaviour in the past.

Every `Link` component or interaction with a router via `react-dom-router` should now refer to the same router.
Any location change that previously required to check if it is under `/beta` or not is now unnecessary.

The PatternFly `Breadcrumb` components have a wrapper to use the router `Link` component to also use the correct router.

To create links to a modal there is `BackgroundLink` which wraps the `Link` component with predefined settings for opening a modal. This is demonstrated with the "Create Policy" button to open the wizard.

Tabs can now be linked to too via setting an anchor/hash in the location (`#details`) and the `RouteredTabSwitcher` component can be used for content, while the `RouteredTabs` component can be used instead of the raw PatterFly component. Both are now used in the `PolicyDetails` to demonstrate.

Interaction with the browser history should also be improved and each route should work as expected when refreshing.


src/PresentationalComponents/ReportTabs/ReportTabs.js
```diff
@@ -1,16 +1,15 @@
...
 import { Tabs, Tab } from '@patternfly/react-core';
 import invert from 'lodash/invert';
 
 export const ReportTabs = (props) => {
     const { match: { path } } = props;
 
     const tabPaths = {
-        0: paths.reports,
-        1: paths.reportsSystems
+        0: '/reports',
```
bastilian, line 11: This is just to please Travis. `ReportTabs` is on the list of endangered components.

bastilian:
It would be good to get this in before #685 as it would help addressing the issues with the interaction of modal and tabs.

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -6,40 +6,45 @@ import {
...
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_PROFILE } from 'Utilities/graphql/mutations';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications';
 import { dispatchAction } from 'Utilities/Dispatcher';
 
-const DeletePolicy = ({ isModalOpen, policy, toggle, onDelete }) => {
+const DeletePolicy = ({ toggle }) => {
+    const location = useLocation();
+    const history = useHistory();
+    const { name, id } = location.state.policy;
```
bastilian, line 18: I wanted to see if this works to pass the policy not using a prop and it does, but it may not be ideal, however it also seemed redundant to fetch the policy only to get information for deleting it.

The only thing that seemed to not work with going to paths like this is that refreshing will not find the route and 404. We could also call this a feature.

src/SmartComponents/PoliciesTable/PoliciesTable.js
```diff
@@ -151,31 +148,23 @@ export class PoliciesTable extends React.Component {
...
+    actionResolver = () => ([
+        {
+            title: 'Delete policy',
+            onClick: (_event, _index, { policyId }) => {
+                const policy = this.props.policies.find((policy) => (
+                    policy.id === policyId
+                ));
+                this.props.history.push(`/scappolicies/${ policyId }/delete`, {
+                    policy,
+                    background: this.props.location
```
bastilian, line 160: When it's not possible to use `BackgroundLink` to link a modal, this is how it can be done using `history`. It simply does what the link would do as well and set the current location as a background.

src/SmartComponents/CompliancePolicies/CompliancePolicies.js
```diff
@@ -46,7 +46,10 @@ const QUERY = gql`
 `;
 
 export const CompliancePolicies = () => {
-    let { data, error, loading, refetch } = useQuery(QUERY, { fetchPolicy: 'cache-and-network' });
+    const location = useLocation();
+    const createLink = <BackgroundLink to='/scappolicies/new'>Create new policy</BackgroundLink>;
+    let { data, error, loading, refetch } = useQuery(QUERY);
+    useEffect(() => { refetch(); }, [location]);
```
bastilian, line 52: This watches the location and refetches when it changes, like when the wizard closes.

bastilian:
@akofink Thank you for testing and reviewing! The tabs are back now. I overlooked removing those while working it!

bastilian:
@akofink yeah, I see this, but it doesn't seem to have anything to do with these changes, but rather https://github.com/RedHatInsights/frontend-components/pull/678. It also appears in master.

There seems to be already a fix for this, that probably just needs to land in the chrome that ships with the proxy. https://github.com/RedHatInsights/frontend-components/pull/685

bastilian:
@akofink \o/ Thank you for reviewing and testing!

PR #691, opened 2020-08-09
Remove isFooterLeftAligned attribute from modals

bastilian:
`isFooterLeftAligned` has been removed and is now default styling for modals. 

See: https://www.patternfly.org/v4/documentation/react/overview/upgrade-guide#modal


src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -109,7 +108,7 @@ CreatePolicy.propTypes = {
     profile: propTypes.string,
     name: propTypes.string,
     refId: propTypes.string,
-    selectedRuleRefIds: propTypes.arrayOf(propTypes.string).isRequired,
+    selectedRuleRefIds: propTypes.arrayOf(propTypes.string),
```
bastilian, line 111: Since I was in that area... selectedRuleRefIds are not "required" (which raises a warning atm) and they are not passed to the component, but come from the redux(form) store.

PR #718, opened 2020-08-27
[RHICOMPL-832] Improve validation for "Create Policy"

bastilian:
* Adds validation for rules to be selected
 * Fixes compliance threshold validation

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreatePolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreatePolicy.js
@@ -12,7 +12,9 @@ import ReviewCreatedPolicy from './ReviewCreatedPolicy';
 import FinishedCreatePolicy from './FinishedCreatePolicy';
 import { validateFirstPage, validateSecondPage, validateThirdPage } from './validate';
 
-export const CreatePolicy = ({ benchmark, profile, name, refId, selectedRuleRefIds }) => {
+export const CreatePolicy = ({
+    benchmark, complianceThreshold, name, profile, refId, selectedRuleRefIds
+}) => {
     const history = useHistory();
     const [stepIdReached, setStepIdReached] = useState(1);
     const onNext = ({ id }) => {
@@ -35,7 +37,7 @@ export const CreatePolicy = ({ benchmark, profile, name, refId, selectedRuleRefI
             name: 'Details',
             component: <EditPolicyDetails/>,
             canJumpTo: stepIdReached >= 2,
-            enableNext: validateSecondPage(name, refId)
+            enableNext: validateSecondPage(name, refId, complianceThreshold)
         },
         {
             id: 3,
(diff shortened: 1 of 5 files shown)


src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -99,9 +102,10 @@ const selector = formValueSelector('policyForm');
 export default connect(
     state => ({
         benchmark: selector(state, 'benchmark'),
-        profile: selector(state, 'profile'),
+        complianceThreshold: selector(state, 'complianceThreshold'),
         name: selector(state, 'name'),
+        profile: selector(state, 'profile'),
         refId: selector(state, 'refId'),
-        selectedRuleRefIds: selector(state, 'selectedRuleRefIds') || []
```
bastilian, line 105: This changes makes the rules validate-able.

src/SmartComponents/CreatePolicy/validate.js
```diff
@@ -6,13 +6,9 @@ export const validateFirstPage = (benchmark, profile) => {
...
 
-export const validateSecondPage = (name, refId) => {
-    if (!name || !refId) {
-        return false;
-    } else {
-        return true;
-    }
-};
+export const validateSecondPage = (name, refId, complianceThreshold) => (
+    !name || !refId  || !complianceThreshold || complianceThreshold > 100 || complianceThreshold < 0 ? false : true
```
bastilian, line 10: This is the validation that will prevent moving forward in the wizard with an invalid threshold.

bastilian:
@akofink Thank you testing and reviewing! I had committed part of the changes to #685 🤦 

It now gives feedback when the threshold is invalid and allows to move forward if it's valid:

Kapture 2020-08-27 at 16 31 33

The other cases where the second screen would not allow to move forward is if name or refId aren't set due to not loading properly. This is the same behaviour we have now in master.

We should definitely extend/improve validations here and in the edit policy modal, but that would be much easier with/after moving to something, probably most likely, data driven forms, which also integrates with the PF4 Wizard.

bastilian:
Just a side note: The hashed filename configuration is causing very odd issues. 
When testing be sure to clear any caches and best apply #719 till it is merged.

src/SmartComponents/ProfileThresholdField/ProfileThresholdField.js
```diff
@@ -7,22 +7,18 @@ import propTypes from 'prop-types';
...
-            this.setState({ validThreshold: true, threshold: round(threshold, 1) });
-        }
+    state = {
+        validThreshold: (this.props.previousThreshold < 101 && this.props.previousThreshold > 0),
+        threshold: round(this.props.previousThreshold || 100, 1)
     };
 
+    handleThresholdChange = (threshold) => (
+        this.setState({
+            validThreshold: (threshold < 101 && threshold > 0),
```
akofink, line 17:
```suggestion
            validThreshold: (threshold < 101 && threshold >= 0),
```

bastilian: 👍 will update that and actually extract it to use the same everywhere this occurs.

bastilian: Update! thanks for the suggestion!

PR #739, opened 2020-09-02
[RHICOMPL-691] - Remove "Edit policies" action from Systems Page

bastilian:
Editing policies can now be done via the policy edit modal and the "Edit policies" can be removed from the systems view.

Screenshot 2020-09-02 at 21 12 13

### DIFF
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -56,6 +56,7 @@ export const ComplianceSystems = () => {
                     <StateViewPart stateKey="data">
                         { policies && <SystemsTable
                             showOsFilter
+                            enableEditPolicy={ false }
                             remediationsEnabled={ false }
                             columns={ columns }
                             policies={ policies } /> }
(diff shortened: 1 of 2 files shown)


src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -56,6 +56,7 @@ export const ComplianceSystems = () => {
                     <StateViewPart stateKey="data">
                         { policies && <SystemsTable
                             showOsFilter
+                            enableEditPolicy={ false }
```
akofink, line 59: I guess this prop is on the chopping block next time we review tech debt? :)

akofink (approved):
Works as expected. Thanks, @bastilian!

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
bastilian: 👍 yup. There are a few things that can now go, but we should remove it all carefully, to not remove anything that is required or that we may have to eventually put back.

sthirugn:
@bastilian looks like it is good to merge? are you planning to push this to prod tomorrow with the rest of COMP-E-132?

PR #740, opened 2020-09-03
[RHICOMPL-574] Fix issues with versions containing 0

bastilian:
There were issues with how the conditions got interpreted with `&&`.

With version numbers like `8.0` it would show just `0`.
This fixes these issues and OS information is correctly shown.

### DIFF
--- a/src/store/Reducers/SystemStore.test.js
+++ b/src/store/Reducers/SystemStore.test.js
@@ -1,7 +1,8 @@
 import {
-    lastScanned,
     compliant,
     findProfiles,
+    hasOsInfo,
+    lastScanned,
     profileNames,
     systemsToInventoryEntities
 } from './SystemStore';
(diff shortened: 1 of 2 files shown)


src/store/Reducers/SystemStore.test.js
```diff
@@ -153,3 +154,36 @@ describe('.profileNames', () => {
...
+
+    it('return true for 8.0', () => {
+        const result = hasOsInfo({
+            osMajorVersion: 8,
+            osMinorVersion: 0
+        });
+        expect(result).toEqual(true);
+    });
+
+    it('return false for both versions being 0', () => {
```
bastilian, line 174: This should in theory not happen with a backend that has proper data and imported hosts correctly and we should not have a OS version `0.0`, but just to be sure.

vkrizan:
LGTM

akofink (approved):
Works as expected! Thanks, @bastilian!

PR #741, opened 2020-09-03
[RHICOMPL-742] Always update state for selected entities

bastilian:
When deselecting systems the reducer will set `selectedEntities` to `undefined` when there are no more systems selected.
The `hosts` property for the `updatedPolicy` should still be updated, but set to an empty array.

Best to test:

 1. Create a policy in the wizard assigning only one host
 2. Open the Edit Policy modal and remove the host.

Before this change when only one host is assigned it wouldn't allow to unassign this host.


vkrizan (approved):
LGTM. Verified.

PR #743, opened 2020-09-04
[RHICOMPL-784] Provide Remediations button with proper system nodes

bastilian:
When we fixed issues with selecting systems for exporting, we changed `selectedEntities` from containing only IDs to hold the system nodes and did not adjust the selecting of systems for the Remediations button.

How to test:

 1) Make sure to have a report with systems that have failed rules, which have remediations available
 2) Got a the report details to verify this behaviour:

Kapture 2020-09-04 at 10 24 32

Before this the behaviour was broken and the "Remediations" button would stay disabled.

Note: To test this with no remediations imported, the `rulesHelpers.js` can be changed to mock this and test the button still:

```
export const systemsWithRuleObjectsFailed = (systems) => (
    systems.map(system => (
        {
            ruleObjectsFailed: systemRulesFailed(system)
            .map((r) => ({ ...r, remediationAvailable: [true, false][Math.floor(Math.random() * 2)] })),
            ...system
        }
    ))
);
```

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -288,7 +288,7 @@ class SystemsTable extends React.Component {
     render() {
         const {
             remediationsEnabled, compact, enableExport, showAllSystems, showActions,
-            selectedEntities, systems, total, policyId, enableEditPolicy
+            selectedEntities, selectedEntitiesIds, systems, total, policyId, enableEditPolicy
         } = this.props;
         const {
             page, perPage, InventoryCmp, selectedSystemId, activeFilters,
@@ -381,9 +381,9 @@ class SystemsTable extends React.Component {
                         { remediationsEnabled &&
                             <reactCore.ToolbarItem style={{ marginLeft: 'var(--pf-global--spacer--lg)' }}>
                                 <ComplianceRemediationButton
-                                    allSystems={ systemsWithRuleObjectsFailed(
-                                        systems.filter(edge => selectedEntities.includes(edge.node.id)
-                                        ).map(edge => edge.node))}
+                                    allSystems={ systemsWithRuleObjectsFailed(systems.filter((edge) => (
+                                        selectedEntitiesIds.includes(edge.node.id)
+                                    )).map(edge => edge.node)) }
                                     selectedRules={ [] } />
                             </reactCore.ToolbarItem>
                         }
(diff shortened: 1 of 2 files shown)


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -381,9 +381,9 @@ class SystemsTable extends React.Component {
                         { remediationsEnabled &&
                             <reactCore.ToolbarItem style={{ marginLeft: 'var(--pf-global--spacer--lg)' }}>
                                 <ComplianceRemediationButton
-                                    allSystems={ systemsWithRuleObjectsFailed(
-                                        systems.filter(edge => selectedEntities.includes(edge.node.id)
-                                        ).map(edge => edge.node))}
+                                    allSystems={ systemsWithRuleObjectsFailed(systems.filter((edge) => (
+                                        selectedEntitiesIds.includes(edge.node.id)
```
vkrizan, line 385: I was wondering if the `edge.node` is the same object as in selected entities, then it could work w/o ids:
```
selectedEntities.includes(edge.node)
```
but it might be a bit ridgit.

bastilian: No, it's not the same object, unfortunately. 

"nodes" are data structures from GraphQL, but selectedEntities are an object compiled out of the GraphQL data and host inventory data. This "compiling" happens in the `entitiesReducer`, or more specifically in the `systemsToInventoryEntities` function.  https://github.com/RedHatInsights/compliance-frontend/blob/master/src/store/Reducers/SystemStore.js#L105

bastilian: Even if it would be the "same" object, it may never really be exactly the same object, unless both places reference the exact same object. 

There are ways to "deep compare" them using lodash for example, but these objects can have a lot of levels and comparing all these would be costly.

vkrizan: Yes, thank you for explanation.

```
>> Object() == Object()
false
```

vkrizan (approved):
LGTM. Verified (w/ the snippet).

@bastilian Would it make sense to add a help message (`title`) on the button to let user's know why it is not activated?

bastilian:
@vkrizan That's a good suggestion, we can bring it up in the next UI/UX review and get a proper text we can show.

PR #744, opened 2020-09-04
Clean up entry files

bastilian:
* Remove unnecessary `getUser` call
 * use `getBaseName` from frontend components utils

### DIFF
--- a/src/entry-dev.js
+++ b/src/entry-dev.js
@@ -2,11 +2,11 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import logger from 'redux-logger';
 import { IntlProvider } from 'react-intl';
-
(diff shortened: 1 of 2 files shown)


bastilian:
@akofink @vkrizan I've been testing these changes to the entry files locally in my dev setup and all remains working.

I believe the `window.insights.chrome.auth.getUser()` is not needed. 

I'm unsure why it was added, the original commit of this doesn't provide much context. It has been added prior to handling authentication errors in a centralised way and may just be a remnant of a workaround.

vkrizan (approved):

akofink (approved):
Seems to continue to work as expected. Thanks, @bastilian!

PR #752, opened 2020-09-11
[RHICOMPL-906] Remove "Edit policies for system"

bastilian:
The "Edit policies for system" should not be used anymore and can therefore be removed.

**Before:**

Screenshot 2020-09-11 at 18 20 18

**After:**

Screenshot 2020-09-11 at 18 18 42


akofink (approved):
Nice! Works as expected, and thanks for the cleanup!

PR #763, opened 2020-09-21
[RHICOMPL-732] Default to failed rules on systems detail page

vkrizan:
Note, that the `useLocation().query` from React Router no longer works.

Screenshot_2020-09-21 Compliance

### DIFF
--- a/src/SmartComponents/SystemDetails/SystemDetails.js
+++ b/src/SmartComponents/SystemDetails/SystemDetails.js
@@ -4,7 +4,6 @@ import {
     useQuery
 } from '@apollo/react-hooks';
 import {
-    useLocation,
     useParams
 } from 'react-router-dom';
 import {
@@ -39,8 +38,6 @@ export const SystemDetails = () => {
     const { data, error, loading } = useQuery(QUERY, {
         variables: { inventoryId }
     });
-    const location = useLocation();
-    const hidePassed = location.query && location.query.hidePassed;
     const systemName = data?.system?.name;
 
     return <StateViewWithError stateValues={ { error, data, loading } }>
@@ -56,7 +53,7 @@ export const SystemDetails = () => {
                 <br/>
             </PageHeader>
             <Main>
-                <ComplianceSystemDetails hidePassed={ hidePassed } inventoryId={ inventoryId } />
+                <ComplianceSystemDetails hidePassed inventoryId={ inventoryId } />
             </Main>
         </StateViewPart>
         <StateViewPart stateKey='loading'>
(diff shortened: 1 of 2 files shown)


vkrizan:
@bastilian The "X" on the "failed" filter does not work for me. But, the clear all filters work. Any idea?

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -56,7 +53,7 @@ export const SystemDetails = () => {
                 <br/>
             </PageHeader>
             <Main>
-                <ComplianceSystemDetails hidePassed={ hidePassed } inventoryId={ inventoryId } />
+                <ComplianceSystemDetails hidePassed={ true } inventoryId={ inventoryId } />
```
bastilian, line 56: `hidePassed={ true }` can be just `hidePassed` without explicitly setting it to true. 
React will assume it to be true if it is set.

vkrizan: Done.

bastilian:
@vkrizan There seems to be an issue with the filter/chip building. That's why you also see just "passed" or "failed" instead of the full correct label. I'm looking into that.

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -39,8 +38,6 @@ export const SystemDetails = () => {
     const { data, error, loading } = useQuery(QUERY, {
         variables: { inventoryId }
     });
-    const location = useLocation();
-    const hidePassed = location.query && location.query.hidePassed;
```
bastilian, line 43: We shouldn't have to worry about that at all anymore. This seems to have been introduced a while back before there were proper filters: https://github.com/RedHatInsights/compliance-frontend/pull/90

akofink (commented):
Looks good and works nearly as expected. I'd expect the filter dropdown itself to also be updated:
image

Also, the filter to show/hide passed rules should be used so the chip looks like this:
image

vkrizan:
@akofink @bastilian Unfortunately the filter label would have to be changed in frontend-components:
* https://github.com/RedHatInsights/frontend-components/blob/2fb1a324622e7bfc5d98108abd9a3d6c69143be3/packages/inventory-compliance/src/SystemRulesTable.js#L36
* https://github.com/RedHatInsights/frontend-components/blob/2fb1a324622e7bfc5d98108abd9a3d6c69143be3/packages/inventory-compliance/src/Utilities/FilterBuilderConfigBuilder.js#L50

bastilian:
@akofink @vkrizan The issue with the wrong label and weird behaviour should be fixed with this: https://github.com/RedHatInsights/frontend-components/pull/744

bastilian (approved):
Thank you @vkrizan! Works as expected! (aside from the oddity that will be fixed with https://github.com/RedHatInsights/frontend-components/pull/744)

PR #771, opened 2020-09-28
[RHICOMPL-977] Add info about what systems are shown

bastilian:
This adds a dismiss-able Info message to the Systems page:

Screenshot 2020-09-28 at 17 34 23

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -120,7 +120,8 @@ class SystemsTable extends React.Component {
         policyId: this.props.policyId,
         perPage: 50,
         totalCount: 0,
-        activeFilters: initFilterState(this.filterConfig)
+        activeFilters: initFilterState(this.filterConfig),
+        showInfoAlert: !(localStorage.getItem('insights:compliance:hidesystemsinfoalert') === 'true')
     }
 
     componentDidMount = () => {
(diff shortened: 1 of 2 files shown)


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -361,6 +369,13 @@ class SystemsTable extends React.Component {
                 <ErrorPage error={error}/>
             </StateViewPart>
             <StateViewPart stateKey='noError'>
+                { (showComplianceSystemsInfo && showInfoAlert) && <reactCore.Alert
+                    isInline
+                    variant="info"
+                    actionClose={ <reactCore.AlertActionCloseButton onClose={ () => this.dismissInfoAlert() } /> }
```
akofink, line 375: Is there an ouia ID on this so QE can check that it's on the page or not?

akofink (approved):
Works as expected, thanks @bastilian! I had one QE-related question.

src/SmartComponents/SystemsTable/SystemsTable.js
bastilian: The defaults should result in attributes like: `data-ouia-component-type="PF4/Alert" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-Alert-info-1"`

Victoremepunto: wrt OUIA there should be ... 
- `Page` attribute: navigate to a specific page with a unique page type, for example, where the same page with / without a modal open are two different "page types". 
- `Component`: unique identifier (like page) for a specific component.
- `PageSafe` attribute: whether if it's true to interact with the page (the components have loaded)

I think @akofink  comment referred to only "Component" , but all three are very welcome :)

Victoremepunto: https://ouia.readthedocs.io/en/latest/README.html#specification-parts

PR #775, opened 2020-09-29
[RHICOMPL-977] Remove possibility to dismiss message about systems

bastilian:
The message should not be dismissable.

Screenshot 2020-09-29 at 19 25 54


akofink (approved):
Works as expected. Thanks!

PR #776, opened 2020-09-29
[RHICOMPL-680] Implement reports as a table view

bastilian:
This implements a table view for reports:

Screenshot 2020-09-29 at 21 22 32

It can be enabled via https://prod.foo.redhat.com:1337/insights/compliance/reports?reportsTableView=enable and disabled by using `disable` instead.

### DIFF
--- a/src/PresentationalComponents/ReportsEmptyState/ReportsEmptyState.js
+++ b/src/PresentationalComponents/ReportsEmptyState/ReportsEmptyState.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import { ComplianceEmptyState } from '@redhat-cloud-services/frontend-components-inventory-compliance';
+import { BackgroundLink } from 'PresentationalComponents';
+
+const ReportsEmptyState = () => (
+    <ComplianceEmptyState
+        title={ 'No policies are reporting' }
+        mainButton={ <BackgroundLink to='/scappolicies/new'>
+            <Button variant='primary'>Create new policy</Button>
+        </BackgroundLink> }
+    />
+);
+
+export default ReportsEmptyState;
--- a/src/PresentationalComponents/ReportsTable/Cells.js
+++ b/src/PresentationalComponents/ReportsTable/Cells.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Label, Text, TextContent, TextVariants, Progress } from '@patternfly/react-core';
+import { Link } from 'react-router-dom';
+import { PolicyPopover } from 'PresentationalComponents';
+
+export const GreySmallText = ({ children }) => (
+    <Text
+        style={{ color: 'var(--pf-global--Color--200)' }}
+        component={ TextVariants.small }>{ children }</Text>
+);
+
+GreySmallText.propTypes = {
+    children: propTypes.node
+};
+
+export const Name = (profile) => (
+    <TextContent>
+        <Link to={'/reports/' + profile.id} style={ { marginRight: '.5rem' }}>
+            { profile.name }
+        </Link>
+        { profile.policy && <PolicyPopover { ...{ profile, position: 'right' } } /> }
+        { !profile.policy && <Label color="red">
+            External
+        </Label> }
+        { profile.policy && <GreySmallText>{ profile.policy.name }</GreySmallText> }
+    </TextContent>
+);
+
+Name.propTypes = {
+    id: propTypes.string,
+    name: propTypes.string,
+    policy: propTypes.object
+};
+
+export const rhelColorMap = {
+    7: 'cyan',
+    8: 'purple'
+};
+export const OperatingSystem = ({ majorOsVersion }) => (
+    <Label color={ rhelColorMap[majorOsVersion] }>RHEL { majorOsVersion }</Label>
+);
+
+OperatingSystem.propTypes = {
+    majorOsVersion: propTypes.string
+};
+
+export const CompliantSystems = ({ totalHostCount = 0, compliantHostCount = 0 }) => (
+    <React.Fragment>
+        <Progress
+            measureLocation={ 'outside' }
+            value={ (100 / totalHostCount) * compliantHostCount } />
+        <GreySmallText>{ `${ compliantHostCount } of ${ totalHostCount } systems` }</GreySmallText>
+    </React.Fragment>
+);
+
+CompliantSystems.propTypes = {
+    totalHostCount: propTypes.number,
+    compliantHostCount: propTypes.number
+};
--- a/src/PresentationalComponents/ReportsTable/Filters.js
+++ b/src/PresentationalComponents/ReportsTable/Filters.js
@@ -1,0 +1,62 @@
+import { conditionalFilterType } from '@redhat-cloud-services/frontend-components';
+
+export const policyNameFilter = [{
+    type: conditionalFilterType.text,
+    label: 'Policy name',
+    filter: (profiles, value) => {
+        const lowerCaseValue = value.toLowerCase();
+        return profiles.filter((profile) => (
+            [
+                profile.name,
+                (profile?.policy?.name || '')
+            ].join().toLowerCase().includes(lowerCaseValue)
+        ));
+    }
+}];
+
+export const policyTypeFilter = (policies) => ([{
+    type: conditionalFilterType.checkbox,
+    label: 'Policy type',
+    filter: (profiles, values) => (
+        profiles.filter(({ policy }) => values.includes(policy?.id))
+    ),
+    items: policies.map((policy) => ({
+        label: policy.name,
+        value: policy.id
+    }))
+}]);
+
+export const operatingSystemFilter = (operatingSystems) => ([{
+    type: conditionalFilterType.checkbox,
+    label: 'Operating system',
+    filter: (profiles, values) => (
+        profiles.filter(({ majorOsVersion }) => (
+            values.includes(majorOsVersion)
+        ))
+    ),
+    items: operatingSystems.map((operatingSystem) => ({
+        label: `RHEL ${operatingSystem}`,
+        value: operatingSystem
+    }))
+}]);
+
+export const policyComplianceFilter = [{
+    type: conditionalFilterType.checkbox,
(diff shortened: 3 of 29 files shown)


src/Utilities/hooks/useFeature.js
```diff
@@ -0,0 +1,51 @@
...
+    history.push(path);
+};
+
+// Queries the local storage for feature flag values
+const getLocatStateFlag = (feature) => (
+    localStorage.getItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`)
+);
+
+// A hook to query feature values
+const useFeature = (feature) => {
```
vkrizan, line 35: You could have submitted/committed the useFeature stuff separately to make this PR a bit smaller. The justification is that it is not directly part of the feature, the feature only depends on it. Smaller bits are conventionally reviewed efficiently.

src/PresentationalComponents/ReportsTable/Filters.js
```diff
@@ -0,0 +1,58 @@
+import { conditionalFilterType } from '@redhat-cloud-services/frontend-components';
+
+export const policyNameFilter = [{
+    type: conditionalFilterType.text,
+    label: 'Policy name',
+    filter: (profiles, value) => (
+        profiles.filter((profile) => (
+            profile.name.includes(value) || (profile?.policy?.name || '').includes(value)
```
vkrizan, line 8: Observation: Case-sensitive filtering.

Just curious, isn't there a common way to search/filter on a string value, so we do not DRY?

src/PresentationalComponents/ReportsTable/Filters.js
```diff
@@ -0,0 +1,58 @@
...
+}]);
+
+export const policyComplianceFilter = [{
+    type: conditionalFilterType.checkbox,
+    label: 'Systemsmeetingcompliance', // FIX in FC: The filterconfig helper can't handle two much space...
+    filter: (profiles, values) => (
+        profiles.filter(({ totalHostCount, compliantHostCount }) => {
+            const compliantHostsPercent = Math.round((100 / totalHostCount) * compliantHostCount);
+            const matching = values.map((value) => {
+                const [min, max] = value.split('-');
```
vkrizan, line 46: Observation: If the filter value can be array, the split would not be necessary.

src/PresentationalComponents/ReportsTable/Filters.js
```diff
@@ -0,0 +1,58 @@
...
+    ),
+    items: operatingSystems.map((operatingSystem) => ({
+        label: `RHEL ${operatingSystem}`,
+        value: operatingSystem
+    }))
+}]);
+
+export const policyComplianceFilter = [{
+    type: conditionalFilterType.checkbox,
+    label: 'Systemsmeetingcompliance', // FIX in FC: The filterconfig helper can't handle two much space...
```
vkrizan, line 41: Try maybe a shorter version, like:
* Meeting Compliance
* Compliant
* Score
* Success

src/PresentationalComponents/ReportsTable/Cells.js
```diff
@@ -0,0 +1,60 @@
...
+GreySmallText.propTypes = {
+    children: propTypes.node
+};
+
+export const Name = (profile) => (
+    <TextContent>
+        <Link to={'/reports/' + profile.id} style={ { marginRight: '.5rem' }}>
+            { profile.name }
+        </Link>
+        { profile.policy && <PolicyPopover { ...{ profile, position: 'right' } } /> }
```
vkrizan, line 22: This does not popup anything for me, at least not for HIPAA.

src/Utilities/hooks/useFeature.js
bastilian: True, I thought about it as well, but in any case I probably still would have had to make changes/fixes on parts that the feature "only" depends on, which in case I would have a PR open with those parts, means rebase/merge chaos.

Small PRs are great for bigger teams, where PRs get reviewed fast, but tend to make things complicated for smaller ones.

src/PresentationalComponents/ReportsTable/Cells.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Label, Text, TextContent, TextVariants, Progress } from '@patternfly/react-core';
+import { Link } from 'react-router-dom';
+import { PolicyPopover } from 'PresentationalComponents';
+
+export const GreySmallText = ({ children }) => (
+    <Text
+        style={{ color: 'var(--pf-global--Color--200)' }}
```
vkrizan, line 9: Observation: Isn't there a class like `text-muted`?

src/PresentationalComponents/ReportsEmptyState/ReportsEmptyState.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import { ComplianceEmptyState } from '@redhat-cloud-services/frontend-components-inventory-compliance';
+import { BackgroundLink } from 'PresentationalComponents';
+
+const ReportsEmptyState = () => (
+    <ComplianceEmptyState
+        title={ 'No policies are reporting' }
+        mainButton={ <BackgroundLink to='/scappolicies/new'>
+            <Button variant='primary'>Create new policy</Button>
```
vkrizan, line 10: Observation: This button seems to be repeated.

src/PresentationalComponents/ReportsTable/Cells.js
```diff
@@ -0,0 +1,60 @@
...
+    children: propTypes.node
+};
+
+export const Name = (profile) => (
+    <TextContent>
+        <Link to={'/reports/' + profile.id} style={ { marginRight: '.5rem' }}>
+            { profile.name }
+        </Link>
+        { profile.policy && <PolicyPopover { ...{ profile, position: 'right' } } /> }
+        { !profile.policy && <Label color="red">
```
vkrizan, line 23: Observation: Profile has `external` attribute (`profile.external`). Is it different?

src/Utilities/hooks/useTableSort.js
```diff
@@ -0,0 +1,51 @@
...
+    })
+);
+
+export const orderByArray = (objectArray, orderProp, orderArray, direction) => (
+    (direction !== 'asc' ? orderArray.reverse() : orderArray).flatMap((orderKey) => (
+        objectArray.filter((item) => (item[orderProp] === orderKey))
+    ))
+);
+
+const useTableSort = (array, columns) => {
```
vkrizan, line 27: Observation: Isn't there something similar in the `react-table` package? [`useSortBy`](https://react-table.tanstack.com/docs/api/useSortBy) maybe?

vkrizan (commented):
Result looks great!

There are some small cases that re-introduce methods existing in other packages or in JS.

src/PresentationalComponents/ReportsTable/Filters.js
bastilian: It has to be "Systems meeting compliance" according to the mockups. I'll just have to fix the issue in the filter helpers in the frontend components, which i'm already working on.

src/PresentationalComponents/ReportsTable/Filters.js
bastilian: Unfortunately it can only be a string in this case.

src/Utilities/hooks/useTableSort.js
bastilian: This `useSortBy` hook is for when using `react-table`, but we use `@patternfly/react-table`.

src/PresentationalComponents/ReportsTable/Cells.js
bastilian: Not in the Patternfly 4 https://www.patternfly.org/v4/get-started/global-css-variables

src/PresentationalComponents/ReportsTable/Cells.js
bastilian: It requires a click. I'll change it so it'll show a different cursor to signal that.

src/PresentationalComponents/ReportsEmptyState/ReportsEmptyState.js
bastilian: We can make note of it and refactor it at some point.

src/PresentationalComponents/ReportsTable/Cells.js
bastilian: We may eventually get rid of `external`, or? This follows the same reasoning as in the ReportsCard: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/ReportCard/ReportCard.js#L139

src/Utilities/hooks/useFeature.js
```diff
@@ -0,0 +1,51 @@
...
+// Example: #reportTableView|(true,clear)
+const setFlagFromUrl = () => {
+    const { search, pathName: path } = useLocation();
+    const history = useHistory();
+
+    if (!search) {
+        return;
+    }
+
+    const [feature, urlValue] = search.replace('?', '').split('=');
```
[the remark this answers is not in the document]
bastilian: Nice I hadn't used this before. It now works even nicer and supports multiple flags in one go.

src/constants.js
```diff
@@ -70,3 +70,9 @@ export const COMPLIANT_SYSTEMS_FILTER_CONFIGURATION = [
         ]
     }
 ];
+
+export const features = {
+    // Enable via /insights/compliance/reports?reportsTableView=true
+    // Clear/disable via /insights/compliance/reports?reportsTableView=clear
```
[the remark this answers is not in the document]
bastilian: Done.

src/PresentationalComponents/ReportsTable/Filters.js
bastilian: It's now case insensitive, by lowercasing everything.

src/PresentationalComponents/ReportsTable/Cells.js
bastilian: The (?) should now show a proper cursor on hover to make it more obvious.

vkrizan (approved):

src/PresentationalComponents/ReportsTable/Cells.js
vkrizan: :+1:

PR #779, opened 2020-10-01
Cleanup & Increasing test coverage

bastilian:
This is a cleanup of a few things that have been waiting to be done:

* Remove unused files
   We haven't been using these and they contribute negatively to our coverage
* Move PoliciesTable & ProfileThresholdField
   These two better belong in PresentationalComponents
* Add DeleteReport as modal route
   This is the last of the Modal components that has not yet been opened via a route
* Add missing tests to up coverage
* RHICOMPL-918 - Set Remediations as dedicated action
   It also fixes the issue of React warning about `tableProps` being set on an element.


akofink (approved):
It seems to continue to work as expected. Thanks, @bastilian!

PR #781, opened 2020-10-01
refactor: Update Ansible icons to use patternfly version

bastilian:
Just to be consistent and use the same icon as we do/will in `inventory-compliance`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -43,6 +43,10 @@ const frontendComponentsMappe = {
     TableFooter: 'Table'
 };
 
+const IconMapper = {
+    AnsibeTowerIcon: 'ansibeTower-icon'
+};
+
 module.exports = {
     presets: [
         [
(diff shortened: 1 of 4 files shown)


babel.config.js
```diff
@@ -43,6 +43,10 @@ const frontendComponentsMappe = {
     TableFooter: 'Table'
 };
 
+const IconMapper = {
+    AnsibeTowerIcon: 'ansibeTower-icon'
```
vkrizan, line 47: What are these transform-imports necessary for?

If the icons in patternfly follow a pattern, then this can be a dynamic transform 
```
importName[0].toLowerCase() + importName.slice(1).replace(/Icon$/, '-icon')
```

bastilian: These transforms are to make babel rewrite imports for certain packages and modules to include specific module exports.

I'm unsure if they follow a pattern, it could also be that there are two components in one file.
Eventually we will migrate to the frontend-components-config package, which will allow us to remove these mappings or at least have those centralised.

vkrizan (approved):
LGTM

PR #782, opened 2020-10-01
fix(RHICOMPL-730): Request IDs for resources over GraphQL

bastilian:
Resources requested from GraphQL should have an ID to avoid issues with "caching" when re-requesting data.

The issue can be reproduced (without this fix) by opening the "Create policy" wizard twice.
The rules step will throw an error, with this fix the error does not appear and the rules step works as expected.


akofink (approved):
Seems to fix the issue. Thanks, @bastilian!

PR #784, opened 2020-10-05
fix(RHICOMPL-730): Add id for benchmark object as well

bastilian:
The Rule step in the wizard was still failing for opening it a second time.
Giving an ID to benchmarks has solved the issue for RHEL + Policy Type where it would raise an error before this change.

Please see Jira for these cases and how to reproduce the issue.


akofink (approved):
Thanks, @bastilian!

PR #786, opened 2020-10-06
fix(RHICOMPL-1023): Add withRouter back to make actions work again

bastilian:
https://github.com/RedHatInsights/compliance-frontend/pull/779 Remove the withRouter decorator, which causes the actionResolver functions not to work properly.


vkrizan (approved):
Both Edit and Remove Policy modals work from the Policy table kebab menu.

PR #787, opened 2020-10-06
fix(RHICOMPL-1014): Show custom policy name primarily on Reports table and Popover

vkrizan:
Custom policy name should be shown primarily on the Reports table, and the canonical policy name shown underneath.
The same should be shown on the Policy Popover.

Screenshot_2020-10-06 Compliance
Screenshot from 2020-10-06 15-06-23

### DIFF
--- a/src/PresentationalComponents/PolicyPopover/PolicyPopover.js
+++ b/src/PresentationalComponents/PolicyPopover/PolicyPopover.js
@@ -3,11 +3,13 @@ import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import {
     Button,
     Popover,
+    Text,
     TextList,
     TextListVariants,
     TextListItem,
     TextListItemVariants,
-    TextContent
+    TextContent,
+    TextVariants
 } from '@patternfly/react-core';
 import { fixedPercentage } from 'Utilities/TextHelper';
 import {
@@ -20,7 +22,12 @@ const PolicyPopover = ({ profile, position = 'top' }) => {
     return (
         <Popover
             { ...{ position } }
-            headerContent={name}
+            headerContent={
+                <TextContent>
+                    { policy.name }
+                    <Text component={ TextVariants.small }>{ name }</Text>
+                </TextContent>
+            }
             footerContent={
                 <Link to={'/scappolicies/' + policy.id} >
                     View policy
--- a/src/PresentationalComponents/ReportsTable/Cells.js
+++ b/src/PresentationalComponents/ReportsTable/Cells.js
@@ -17,13 +17,13 @@ GreySmallText.propTypes = {
 export const Name = (profile) => (
     <TextContent>
         <Link to={'/reports/' + profile.id} style={ { marginRight: '.5rem' }}>
-            { profile.name }
+            { profile.policy ? profile.policy.name : profile.name }
         </Link>
         { profile.policy && <PolicyPopover { ...{ profile, position: 'right' } } /> }
         { !profile.policy && <Label color="red">
             External
         </Label> }
-        { profile.policy && <GreySmallText>{ profile.policy.name }</GreySmallText> }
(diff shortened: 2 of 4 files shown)


src/PresentationalComponents/ReportsTable/Cells.js
```diff
@@ -17,13 +17,13 @@ GreySmallText.propTypes = {
...
         <Link to={'/reports/' + profile.id} style={ { marginRight: '.5rem' }}>
-            { profile.name }
+            { profile.policy ? profile.policy.name : profile.name }
         </Link>
         { profile.policy && <PolicyPopover { ...{ profile, position: 'right' } } /> }
         { !profile.policy && <Label color="red">
             External
         </Label> }
-        { profile.policy && <GreySmallText>{ profile.policy.name }</GreySmallText> }
+        { profile.policy && <GreySmallText>{ profile.name }</GreySmallText> }
```
bastilian, line 26: This is confusing.  Why check for the `profile.policy` and show the `profile.name`?

I think it should be the same as the above change.

src/PresentationalComponents/PolicyPopover/PolicyPopover.js
```diff
@@ -20,7 +22,15 @@ const PolicyPopover = ({ profile, position = 'top' }) => {
     return (
         <Popover
             { ...{ position } }
-            headerContent={name}
+            headerContent={
+                <TextContent>
+                    { policy.name }
+                    <br />
```
bastilian, line 28: Please use a header component with proper fitting bottom margin.

src/PresentationalComponents/ReportsTable/Cells.js
vkrizan: If `profile.policy` is defined then it's name is shown at the top, therefore the profile custom name must be shown underneath it (with grey). I can rework it to `if-else` block if you advise how.

src/PresentationalComponents/PolicyPopover/PolicyPopover.js
vkrizan: Do you mean to wrap the main header and the sub-header with some component? Can you please elaborate?

Isn't the bottom margin handled by the Popover component?
Screenshot from 2020-10-06 15-17-15

bastilian: Oh, that already gets wrapped.

In this case, i'd move the subheader to the body content and and use an appropriate heading level from: https://www.patternfly.org/v4/components/text

bastilian: It's just because `<br /> is brrrrrr.... plus it makes more sense to keep heading and subheading in two separate headings.

vkrizan: The heading already is `h6`, which is the minimal level. `<small>` (or similar) could work with it. I'll need to check PF possibilities. Feel free to suggest a code change.

Yeah, I don't like `<br>` either :wink:.

vkrizan: If the sub-header is moved to the main content, then its available width would be wider than the main heading (see the box in the picture above). Not sure if that is okay or if it would create another paper-cut. :wink:

vkrizan: `TextVariants.small` alone in the header did the trick.

vkrizan: Do you want it to be part of the body?
Like this
Screenshot_2020-10-07 Compliance

akofink (approved):
Works well and matches the mocks. Thanks, @vkrizan!

akofink:
I was able to re-send the payload for the pr-merge job in the webhooks settings.

PR #788, opened 2020-10-06
fix(RHICOMPL-880): Use selectRulesTableColumns to not redefine columns

bastilian:
This uses the helper function to get column definitions for the rules table introduced with https://github.com/RedHatInsights/frontend-components/pull/765 and will also fixes the issues with html in the tooltip, by avoiding it.


akofink (approved):
Seems to work as expected. Thanks, @bastilian!

PR #799, opened 2020-10-13
fix(RHICOMPL-1030): closing report delete modal after deletion

vkrizan:


akofink (approved):
Thanks! Works as expected.

PR #803, opened 2020-10-21
fix(RHICOMPL-1052): external profiles means no policy

akofink:
This changes the systems view in the UI to determine if a policy is
external by looking at the profile's policy attribute rather than the
external attribute.


vkrizan:
`PolicyDetails` seems to use the `external` flag. It suppose to show an error when a policy is external. Could you consider updating it (probably removing it)?

vkrizan (approved):
LGTM. Not tested.

akofink:
@vkrizan I think the `PolicyDetails` are using `external` in a correct way. It's never valid to view a profile as a policy that has `external=true`:

image

PR #813, opened 2020-10-29
feat(RHICOMPL-689): Implement support for multi-ssg-version policies

bastilian:
This implements support for policies with multiple SSG versions:

Screenshot 2020-10-29 at 12 59 08

Enable via: /insights/compliance/scappolicies/POLICY_ID?multiversionTabs=enable

Note: There are issues with using other components in a Patternfly `Tab` title where even `!important` properties are ignored


akofink (approved):
Works as expected. Thanks, @bastilian!

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -0,0 +1,46 @@
...
+import { SystemRulesTable, selectRulesTableColumns } from '@redhat-cloud-services/frontend-components-inventory-compliance';
+import { RoutedTabs } from 'PresentationalComponents';
+
+const PolicyMultiversionRules = ({ policy: { policy: { profiles } } }) => (
+    <React.Fragment>
+        <Alert variant="info" isInline title="Rule editing coming soon" />
+        <PageSection variant={ PageSectionVariants.light }>
+            <RoutedTabs level={ 1 } defaultTab={ profiles[0].ssgVersion }>
+                {
+                    profiles.map((profile) => (
```
vkrizan, line 13: What's the order of the profiles / tabs?

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -0,0 +1,46 @@
...
+                                    }}>{ profile.rules.length }</Label>
+                                </>
+                            }
+                            eventKey={ profile.ssgVersion }>
+                            <SystemRulesTable
+                                remediationsEnabled={false}
+                                columns={ selectRulesTableColumns(['Rule', 'Severity', 'Ansible']) }
+                                profileRules={[{
+                                    profile: { ...profile },
+                                    rules: profile.rules
```
vkrizan, line 32: A bit unrelated: The Ansible is shown as "No" for rules on other profiles/SSGs

Screenshot_2020-10-30 Compliance

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -0,0 +1,46 @@
...
+                    profiles.map((profile) => (
+                        <Tab
+                            key={ profile.ssg_version }
+                            title={
+                                <>
+                                    SSG { profile.ssgVersion }
+                                    <Label color="blue" style={{
+                                        marginRight: '.5rem !important',
+                                        backgroundColor: '#1C66CB !important',
+                                        color: '#FFFFFF !important'
```
vkrizan, line 22: Observation: The color and gutter values could come from variables.

src/PresentationalComponents/TabSwitcher/TabSwitcher.js
```diff
@@ -17,8 +28,8 @@ TabSwitcher.propTypes = {
     children: propTypes.node
 };
 
-export const RoutedTabSwitcher = ({ children, defaultTab }) => {
-    const currentAnchor = useAnchor(defaultTab);
+export const RoutedTabSwitcher = ({ children, defaultTab, level }) => {
+    const { currentAnchor } = useAnchorLevels(defaultTab, level);
```
vkrizan, line 32: I wonder if there is an official way on how to handle deeper tabs selection via URL.

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
bastilian: As they are returned from the backend.

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
bastilian: Yep. This has been like this from the start. The mockups do not show any no state yet, but we could look into replacing it. There is a `<Ansible />` component in the frontend-components repo that may help.

src/PresentationalComponents/TabSwitcher/TabSwitcher.js
bastilian: Well, there is no "official" way to use the anchors for Tab delegation at all.

vkrizan:
There is a weird behavior when coming from the policies table into the details, where title, breadcrumbs and "Policy Type" changes when after the GQL request is complete.

I have a policy named "My CIS". On the table I see "CIS Red Hat Enterprise Linux 7 Benchmark" and when I load it (click) then it keeps showing the "CIS Red Hat Enterprise Linux 7 Benchmark" for a while and then it swaps it to "My CIS".

bastilian:
@vkrizan Thank you for testing this and catching that. However, this seems to already also happen on CI without this feature even being available. I suggest we open a Jira and look into this separately

Kapture 2020-10-30 at 10 20 20

vkrizan (approved):

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
vkrizan: Which means that they are unordered. How they should be ordered? The mockups show descending order.

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
bastilian: These were not working at all actually. Patternfly strips some styles under certain conditions, even with !important. Just using "blue" as the color looks fine for now, it's not the colours we want, but we'll get those when we can resolve the issues with custom styling.

PR #814, opened 2020-10-29
fix(RHICOMPL-1069): SystemsTable to show systems for a policy

vkrizan:
querying GQL `policy_id`, which also accepts profile id of policy that
the system is assigned to.

Requires backend changes RedHatInsights/compliance-backend#629


akofink (approved):
Thanks, @vkrizan!

PR #817, opened 2020-11-04
[RHICOMPL-1027][RHICOMPL-1036] Reports grouped by policy

vkrizan:
The reports page should group profiles by policy.

It queries internal profiles (act as a policy to a user) that have a policy test result, and, it queries external profiles (not part of any policy) that have a test result. 
Note, that a policy test result happens when at least one policy profile has a test result.

The system reducer should pick profiles that are part of the policy.

This covers case when a policy report is shown (with internal profile
loaded) having a system scanned with a different SSG.  In that case the
host would have different profileId, however it is still part of the
policy.

### DIFF
--- a/src/SmartComponents/Reports/Reports.js
+++ b/src/SmartComponents/Reports/Reports.js
@@ -10,8 +10,8 @@ import {
 import useFeature from 'Utilities/hooks/useFeature';
 
 const QUERY = gql`
-{
-    profiles(search: "has_test_results = true", limit: 1000){
+query Profiles($filter: String!) {
+    profiles(search: $filter, limit: 1000){
         edges {
             node {
                 id
@@ -71,9 +71,21 @@ export const Reports = () => {
     let profiles = [];
     let showView = false;
     const location = useLocation();
-    let { data, error, loading, refetch } = useQuery(QUERY);
+    const profilesGroupedByPolicy = useFeature('profilesGroupedByPolicy');
     const showTableView = useFeature('reportsTableView');
     const View = showTableView ? ReportsTable : ReportCardGrid;
+
+    let searchFilter;
+    if (profilesGroupedByPolicy) {
+        searchFilter = `(has_policy_test_results = true AND external = false)
+                        OR (has_policy = false AND has_test_results = true)`;
+    } else {
+        searchFilter = 'has_test_results = true';
+    }
+
+    let { data, error, loading, refetch } = useQuery(QUERY, {
+        variables: { filter: searchFilter }
+    });
 
     useEffect(() => {
         refetch();
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -22,6 +22,7 @@ import {
 
 import { exportFromState, selectAll, clearSelection, SELECT_ENTITY } from 'Store/ActionTypes';
 import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
+import useFeature from 'Utilities/hooks/useFeature';
 import { FilterConfigBuilder } from '@redhat-cloud-services/frontend-components-inventory-compliance';
 import { entitiesReducer } from 'Store/Reducers/SystemStore';
 import {
@@ -57,6 +58,9 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
                         title
                         compliant
                         remediationAvailable
+                    }
+                    policy {
+                        id
                     }
                 }
             }
@@ -257,7 +261,7 @@ class SystemsTable extends React.Component {
     }
 
     async fetchInventory() {
-        const { columns, policyId, showAllSystems, clearInventoryFilter } = this.props;
+        const { columns, policyId, showAllSystems, clearInventoryFilter, profilesGroupedByPolicy } = this.props;
         const {
             inventoryConnector,
             INVENTORY_ACTION_TYPES,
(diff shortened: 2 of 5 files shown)


akofink (commented):
Seems to work as expected. Thanks! I'll wait to ACK until it's updated with the pending backend changes.

vkrizan:
Please re-test with RedHatInsights/compliance-frontend#820.

The query has been changed to
```
(external = false AND has_policy_test_results = true) OR (has_policy = false AND has_test_results = true)
```

src/SmartComponents/Reports/Reports.js
```diff
@@ -11,7 +11,7 @@ import useFeature from 'Utilities/hooks/useFeature';
 
 const QUERY = gql`
 {
-    profiles(search: "has_test_results = true", limit: 1000){
+    profiles(search: "(external = false AND has_policy_test_results = true) OR (has_policy = false AND has_test_results = true)", limit: 1000){
```
vkrizan, line 14: Do you guys know how to break the line?

akofink: We should probably define a constant for it at this point :)

akofink: This (or similar) seems to work:

```suggestion
    profiles(search: "(external = false AND has_policy_test_results = true)
                      OR (has_policy = false AND has_test_results = true)",
                      limit: 1000){

```

akofink (approved):
Seems to work as expected with the backend changes. Thanks, @vkrizan! Feel free to merge after figuring out the eslint issue.

src/store/Reducers/SystemStore.js
```diff
@@ -24,7 +24,10 @@ export const findProfiles = (system, profileIds) => {
     let profiles = system.profiles;
 
     if (profileIds !== undefined && profileIds.toString() !== '') {
-        profiles = profiles.filter((profile) => profileIds.includes(profile.id));
+        profiles = profiles.filter((profile) => {
+            return profileIds.includes(profile.id) ||
+                  (profile.policy && profileIds.includes(profile.policy.id));
```
vkrizan, line 29: @bastilian This should not affect old views after the migration, as reports form non-latest SSGs would not have a policy. It could affect showing systems for profiles that are not grouped by a policy. Then it depends if we want have grouping under a feature flag or if we want to release it altogether (enabled).

src/store/Reducers/SystemStore.js
```diff
@@ -18,13 +18,18 @@ import {
     profilesRulesPassed,
     profilesRulesFailed
 } from 'Utilities/ruleHelpers';
+import useFeature from 'Utilities/hooks/useFeature';
 import Truncate from 'react-truncate';
 
 export const findProfiles = (system, profileIds) => {
     let profiles = system.profiles;
+    const profilesGroupedByPolicy = useFeature('profilesGroupedByPolicy');
```
vkrizan, line 26: @bastilian How can this be used here, or be refactored in a way that the find policies can be controlled by a feature flag?

bastilian: Does the featureflag not work here?

vkrizan: Nope
```
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
    React 2
    useLocation react-router.js:727
    setFlagsFromUrl useFeature.js:19
    useFeature useFeature.js:43
    findProfiles SystemStore.js:26
```

bastilian: Oh, right. You could pass in the feature flag to the entitiesReducer, similarly to the policyId: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/SystemsTable.js#L279

bastilian: It'll probably also require a prop that can be passed in by the Reports Details page.

vkrizan: Changed to `useFeature` in `ConnectedSystemsTable`. It would set `profilesGroupedByPolicy` prop on the `SystemsTable`.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -486,6 +493,11 @@ const mapDispatchToProps = dispatch => {
 };
 
 const ConnectedSystemsTable = (props) => {
+    const profilesGroupedByPolicy = useFeature('profilesGroupedByPolicy');
+    props = Object.assign({
```
bastilian, line 497: Use ES6 magic: `<SystemsTable { ...props, profilesGroupedByPolicy }` and no need to overwrite props.

vkrizan: :+1:

vkrizan: Does not work :-/
```
Unexpected token, expected "}" (497:34)
```

bastilian: Oh, sorry, there is probably on more level required  `<SystemsTable { ...{ ...props, profilesGroupedByPolicy } } store={ReactRedux.useStore()} />`

vkrizan:
Agreed to rework this to use filtering on backend, relieve Systems reducer, and not use the feature flag.

/cc @akofink @bastilian

PR #819, opened 2020-11-04
Refactor inventory connector by introducing new component

karelhala:
### Refactor inventory connector

The inventory connector is too complex and mixes multiple approaches together, this may cause unwanted refreshes and problematic debugging, this PR attempts to solve that by introducing new smaller component that connects to inventory with hooks and simpler approach of refreshing data, running exports etc.

### TODO
* [x] write new component
* [x] refactor inventory store (partially done, just need to test if nothing really breaks)
* [x] test new component in all places (I've ran this in all places and it doesn't break, just a few tweaks are required to make it fully working)


bastilian:
Thank you @karelhala ! This is great! I’ve been testing it and it works great, for the most part.

One issue that would need to be fixed before we could merge this is when the newInventory is disabled, the “Systems” page crashes. Other than that all appears to working normally with the old SystemsTable.

With the newInventory enabled it does work nicely and more stable than now, but the issue with the shared store has become more prevalent. There also seem to be some oddities when filtering, but that could possibly be due to changes in the backend and/or the issues with the store. 

Once the issue with it when it’s disabled is fixed, it has my approval to merge, but I’ll leave it up, till I'm back to @vkrizan and @akofink to give it a look and merge it.

karelhala:
@bastilian good point! I've fixed it so it's backwards compatible as well. And good points on the improvements for shared store. We can take a look at it later on and improve it even more.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -1,24 +1,32 @@
...
         key: 'facts.compliance.display_name',
         title: 'System name',
         props: {
             width: 40, isStatic: true
+        },
+        ...newInventory && {
+            key: 'display_name'
         }
     }, {
         key: 'facts.compliance.profiles',
```
vkrizan, line 23: Replace: `s/profiles/policies/`

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -1,24 +1,32 @@
...
         title: 'Policies',
         props: {
             width: 40, isStatic: true
+        },
+        ...newInventory && {
+            key: 'profiles'
```
vkrizan, line 29: Replace: `s/profiles/policies/`

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -100,9 +106,13 @@ export const ReportDetails = ({ route }) => {
...
         props: {
             width: 30
+        },
+        ...newInventory && {
+            key: 'display_name',
+            renderFunc: systemName
         }
     }, ...showSsgVersions ? [{
-        key: 'facts.compliance',
+        key: 'facts.compliance.ssg_version',
```
vkrizan, line 115: Can you please check this one? It does not seems to show the version correctly on both variants.

Old SystemsTable variant:
Screenshot_2020-12-14 Report My RHEL7 HIPAA - Reports - Compliance Red Hat Insights

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -121,15 +131,26 @@ export const ReportDetails = ({ route }) => {
...
         }
     }, {
         key: 'facts.compliance.last_scanned',
         title: 'Last scanned',
         props: {
             width: 10
+        },
+        ...newInventory && {
+            key: 'lastScanned',
+            renderFunc: (lastScanned) => (lastScanned instanceof Date) ?
```
vkrizan, line 147: Observation: Another copy of `facts.compliance.last_scanned`.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -0,0 +1,226 @@
+import React, { useEffect, useState, useRef, useCallback } from 'react';
+import { withApollo } from '@apollo/react-hoc';
+import PropTypes from 'prop-types';
+import { useStore, useDispatch, useSelector, shallowEqual } from 'react-redux';
```
vkrizan, line 4: Just curious, would it be possible to use vanilla React, without Redux?

src/SmartComponents/SystemsTable/constants.js
```diff
@@ -0,0 +1,109 @@
...
+query getSystems($filter: String!, $perPage: Int, $page: Int) {
+    systems(search: $filter, limit: $perPage, offset: $page) {
+        totalCount
+        edges {
+            node {
+                id
+                name
+                osMajorVersion
+                osMinorVersion
+                profiles {
```
vkrizan, line 81: Replace: `s/profiles/testResultProfiles/`.

Plus add policies at the same level (sync with SystemsTable).

src/SmartComponents/SystemsTable/constants.js
```diff
@@ -0,0 +1,109 @@
...
```
vkrizan, line 50: Replace: `s/profiles/testResultProfiles/`.

Plus add policies at the same level (sync with SystemsTable).

vkrizan (commented):
@karelhala Thank you very much for the refactoring!

@bastilian One semi-major thing here is the breakage of SSG Version display. Once resolved, we're good to go.

src/SmartComponents/SystemsTable/InventoryTable.js
karelhala: That's a good point, it is possible to surface selectors and dispatched functions to parent components, that would unfortunately mean duplicate code (for instance the selectors are all the same). I wouldn't do it in this PR, but it is possible. However the `useStore` is mandatory and would need to be surfaced to parent component as well and passed down to inventory component.

vkrizan: No need to rewrite it. I was just wondering how to do it and whether it is possible, motivated their own unofficial motto "don't use it unless you really need to". The vanilla React is seemingly improving its capabilities...

Instead Redux store the React's Context might be used. Maybe the context approach would create less maintainable variant (?).

vkrizan (approved):

PR #820, opened 2020-11-05
fix(RHICOMPL-1005): Correct filter name on reports table

bastilian:
Follow up to https://github.com/RedHatInsights/frontend-components/pull/761


akofink (approved):
Thanks!

PR #821, opened 2020-11-06
feat(RHICOMPL-1059): Add notice on external report deprecation

bastilian:
Screenshot 2020-11-06 at 14 45 53

### DIFF
--- a/src/SmartComponents/Reports/Reports.js
+++ b/src/SmartComponents/Reports/Reports.js
@@ -3,6 +3,7 @@ import propTypes from 'prop-types';
 import { useLocation } from 'react-router-dom';
 import { useQuery } from '@apollo/react-hooks';
 import gql from 'graphql-tag';
+import { Alert, Text } from '@patternfly/react-core';
 import { PageHeader, PageHeaderTitle, Main, SkeletonTable } from '@redhat-cloud-services/frontend-components';
 import {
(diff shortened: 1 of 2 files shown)


src/SmartComponents/Reports/Reports.js
```diff
@@ -63,7 +64,19 @@ LoadingView.propTypes = {
 
 const ReportsHeader = () => (
     <PageHeader>
-        <PageHeaderTitle title="Reports" />
+        <PageHeaderTitle title="Reports" className="pad-bottom" />
+        <Alert
+            variant="info"
+            isInline
+            title="Rule editing coming soon"
```
vkrizan, line 71: This might be changed, e.g. to "Deprecation of external reports".

akofink (commented):
Looks good, other than changing the title. :+1: Thanks!

src/SmartComponents/Reports/Reports.js
bastilian: 🤦

bastilian: Corrected to what is in the mockup.

akofink (approved):
Thanks!

vkrizan (commented):
The alert should be dismissable, as per mock. Once dismissed, I believe it should not be shown any more.

bastilian:
@vkrizan The mockup seems to be incorrect. According to PM this message should not be dismissible.

PR #823, opened 2020-11-10
feat(RHICOMPL-684): Add SSG version to report details systems

bastilian:
Screenshot 2020-11-10 at 14 01 55

### DIFF
--- a/src/SmartComponents/ReportDetails/ReportDetails.js
+++ b/src/SmartComponents/ReportDetails/ReportDetails.js
@@ -12,6 +12,7 @@ import {
 } from '@redhat-cloud-services/frontend-components';
 
 import { fixedPercentage, pluralize } from 'Utilities/TextHelper';
+import useFeature from 'Utilities/hooks/useFeature';
 import {
     BackgroundLink, BreadcrumbLinkItem, ReportDetailsContentLoader, ReportDetailsDescription,
     StateViewWithError, StateViewPart
@@ -48,6 +49,8 @@ query Profile($policyId: String!){
 `;
 
 export const ReportDetails = () => {
+    let showSsgVersions;
+    let showSsgVersionsFeature = useFeature('showSsgVersions');
     const { report_id: policyId } = useParams();
     const { data, error, loading } = useQuery(QUERY, {
         variables: { policyId }
@@ -60,6 +63,7 @@ export const ReportDetails = () => {
 
     if (!loading && data) {
         profile = data.profile;
+        showSsgVersions = !!profile?.policy && showSsgVersionsFeature;
         const compliantHostCount = profile.compliantHostCount;
         const totalHostCount = profile.totalHostCount;
         donutId = profile.name.replace(/ /g, '');
@@ -87,7 +91,13 @@ export const ReportDetails = () => {
         props: {
             width: 5
         }
-    }, {
+    }, ...showSsgVersions ? [{
+        key: 'facts.compliance.ssg_version',
+        title: 'SSG version',
+        props: {
+            width: 5
+        }
+    }] : [], {
         key: 'facts.compliance.compliance_score',
         title: 'Compliance score',
         props: {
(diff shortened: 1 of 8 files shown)


akofink (commented):
Looking good! The mocks show SSG version only for ~external~ internal policies, but I see them on ~internal~ external policy reports as well. Also, I'm getting multiple SSG versions listed in the column, despite only having one report uploaded for that host.

image

bastilian:
>  The mocks show SSG version only for external policies, but I see them on internal policy reports as well. 

We should clarify that, the mocks don't show an SSG version for an external policy, but does mention a SSG filter to be added in both cases.

> Also, I'm getting multiple SSG versions listed in the column, despite only having one report uploaded for that host.

This is dependent on fixing the profiles association to only return profiles of that policy. This should then only show one SSG version, unless the host does indeed have two reports of different SSG versions of that policy. It is implemented to deal with multiple SSG versions, but in the default scenario it should only show one on the SystemsTable.

akofink:
Oops, sorry for confusing the mock. The point still stands, though: external policies shouldn't have the SSG version column. The reason why is because an external policy only has a single SSG version, whereas an internal policy has potentially many SSG versions. Here's what I see currently for an external policy report:

image

The mock shows the SSG version in the top header portion of the page, which we can add in another PR if you'd like:

image

bastilian:
@akofink The SSG version should now only be visible on internal profiles.

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -60,6 +62,7 @@ export const ReportDetails = () => {
 
     if (!loading && data) {
         profile = data.profile;
+        showSsgVersions = profile?.policy !== undefined && useFeature('showSsgVersions');
```
vkrizan, line 65: React hooks must not be defined within a condition, IIRC. Suggestion, to move it into top of the `ReportDetails` (where other hooks are defined).

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -52,6 +52,7 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
                     compliant
                     external
                     score
+                    ${ showSsg && 'ssgVersion' || '' }
```
vkrizan, line 55: Observation: It's safe to always query for `ssgVersion` (unless there is a condition along the way).

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -87,7 +90,13 @@ export const ReportDetails = () => {
         props: {
             width: 5
         }
-    }, {
+    }, ...showSsgVersions && [{
```
vkrizan, line 93: Would it be possible to improve code readability here, please. :slightly_smiling_face:

src/SmartComponents/ReportDetails/ReportDetails.js
bastilian: Moved it up.

src/SmartComponents/SystemsTable/SystemsTable.js
bastilian: That is not always true. The backend does need to have the field defined or it will error out:

Screenshot 2020-11-10 at 18 05 43

However, I did remove the condition and it's now requested in any case, even if the SSG version is not shown.

src/SmartComponents/ReportDetails/ReportDetails.js
bastilian: A classic ternary operation, that's the best I can do to put this into the columns array in this position without having to slice and dice the array.

akofink (approved):
Thanks, @bastilian! It seems to work as expected now.

vkrizan (approved):
Thank you Sebastian.

PR #824, opened 2020-11-10
[RHICOMPL-1027] Group reports by policy and aggregate systems

vkrizan:
The reports page should group profiles by policy.
It queries internal profiles (act as a policy to a user) that have a
policy test result, and, it queries external profiles (not part of any
policy) that have a test result

Policy systems are aggregated from all policy profiles, by
leveraging filtering of system by a policy on GraphQL.

**Requires backend RedHatInsights/compliance-backend#647**

Please note, that the policy deduplication on Systems page will be done in separate PR.

Screenshot_2020-11-10 Compliance
Screenshot_2020-11-10 Compliance(1)

### DIFF
--- a/src/SmartComponents/Reports/Reports.js
+++ b/src/SmartComponents/Reports/Reports.js
@@ -11,8 +11,8 @@ import {
 import useFeature from 'Utilities/hooks/useFeature';
 
 const QUERY = gql`
-{
-    profiles(search: "has_test_results = true", limit: 1000){
+query Profiles($filter: String!) {
+    profiles(search: $filter, limit: 1000){
         edges {
             node {
                 id
@@ -84,9 +84,14 @@ export const Reports = () => {
     let profiles = [];
     let showView = false;
     const location = useLocation();
-    let { data, error, loading, refetch } = useQuery(QUERY);
     const showTableView = useFeature('reportsTableView');
     const View = showTableView ? ReportsTable : ReportCardGrid;
+    const filter = `(has_policy_test_results = true AND external = false)
+                    OR (has_policy = false AND has_test_results = true)`;
+
+    let { data, error, loading, refetch } = useQuery(QUERY, {
+        variables: { filter }
+    });
 
     useEffect(() => {
         refetch();
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -35,7 +35,7 @@ import {
 } from 'PresentationalComponents';
 
 export const GET_SYSTEMS = gql`
-query getSystems($filter: String!, $perPage: Int, $page: Int) {
+query getSystems($filter: String!, $policyId: ID, $perPage: Int, $page: Int) {
     systems(search: $filter, limit: $perPage, offset: $page) {
         totalCount
         edges {
@@ -44,7 +44,7 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
                 name
                 osMajorVersion
                 osMinorVersion
-                profiles {
+                profiles(policyId: $policyId) {
                     id
                     name
                     refId
@@ -67,7 +67,7 @@ query getSystems($filter: String!, $perPage: Int, $page: Int) {
 `;
 
 export const GET_SYSTEMS_WITHOUT_FAILED_RULES = gql`
-query getSystems($filter: String!, $perPage: Int, $page: Int) {
+query getSystems($filter: String!, $policyId: ID, $perPage: Int, $page: Int) {
     systems(search: $filter, limit: $perPage, offset: $page) {
         totalCount
         edges {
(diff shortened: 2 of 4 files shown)


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -167,20 +167,24 @@ class SystemsTable extends React.Component {
         const { client, showOnlySystemsWithTestResults, remediationsEnabled } = this.props;
         const { policyId, perPage, page, activeFilters } = this.state;
         let filter = this.filterBuilder.buildFilterString(activeFilters);
+        let variables;
 
         if (showOnlySystemsWithTestResults) {
             filter = `has_test_results = true ${filter.length > 0 ? `and ${filter}` : ''}`;
         }
 
         if (policyId && policyId.length > 0) {
```
akofink, line 176: before, `policyId` was an empty string by default; checking length made sense. Now it's `null`, so perhaps we shouldn't be checking length?

vkrizan: Yes, that's true. I'm not sure about upper layers.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -167,20 +167,24 @@ class SystemsTable extends React.Component {
...
             filter = `policy_id = ${policyId} and ${filter}`;
+            variables = { filter, perPage, page, policyId };
```
bastilian, line 178: You could define 

```javascript
let variables = { filter, perPage, page }

```

 and here just have 

```js
variables = { ...variables, policyId }
```

 and get rid of the `else { ...}`

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -425,7 +429,7 @@ SystemsTable.propTypes = {
 };
 
 SystemsTable.defaultProps = {
-    policyId: '',
+    policyId: null,
```
bastilian, line 432: It could just be removed from defaultProps, I think.

src/store/Reducers/SystemStore.js
```diff
@@ -20,32 +20,22 @@ import {
 } from 'Utilities/ruleHelpers';
 import Truncate from 'react-truncate';
 
-export const findProfiles = (system, profileIds) => {
```
bastilian, line 23: 🎉

akofink (commented):
Seems to work as expected with your backend changes. Thanks, @vkrizan!

src/SmartComponents/SystemsTable/SystemsTable.js
vkrizan: `null !== undefined`, but I guess it could, unless it is needed to keep the list of used props.

bastilian: `undefined` is still falsy. It should work and once the `.length > 0` on 176 is gone it should work properly and pass the `policyId` properly to the query.

src/store/Reducers/SystemStore.js
```diff
@@ -20,32 +20,22 @@ import {
...
     const last = new Date(Math.max.apply(null, dates.filter((date) => isFinite(date))));
     const result = (last instanceof Date && isFinite(last)) ? last : 'Never';
 
     return result;
 };
 
-export const compliant = (system, profileId) => {
-    const profiles = findProfiles(system, [profileId]);
+export const compliant = (system) => {
+    const profiles = system.profiles || [];
```
bastilian, line 33: You could write these two lines as one:
```js
export const compliant = ({profiles = []}) => {
```

src/SmartComponents/Reports/Reports.js
```diff
@@ -71,9 +71,14 @@ export const Reports = () => {
     let profiles = [];
     let showView = false;
     const location = useLocation();
-    let { data, error, loading, refetch } = useQuery(QUERY);
     const showTableView = useFeature('reportsTableView');
     const View = showTableView ? ReportsTable : ReportCardGrid;
+    const searchFilter = `(has_policy_test_results = true AND external = false)
```
bastilian, line 76: It could be named `filter` then on 80 it would only need the shorthand like `variables: { filter }`

src/store/Reducers/SystemStore.js
bastilian: Nitpicky: this function then could also use implicit return.

src/SmartComponents/SystemsTable/SystemsTable.js
vkrizan: Thanks! Changed.

src/SmartComponents/Reports/Reports.js
vkrizan: Thanks! Changed.

src/store/Reducers/SystemStore.js
vkrizan: Thanks. Changed.

bastilian (approved):
Looks great! Thank you @vkrizan!

PR #825, opened 2020-11-11
[RHICOMPL-1014] Fix policy name and type on Report pages

vkrizan:
Fixes Reports table and details to show policy type and user defined policy name:
* on reports table
* on report cards
* on policy popover
* on policy details (title and breadcrumbs)

Additional fixes to #787.

Screenshot_2020-11-11 Compliance
Screenshot_2020-11-11 Compliance(1)
Screenshot_2020-11-11 Compliance(2)


src/PresentationalComponents/ReportsTable/Cells.js
```diff
@@ -19,11 +19,14 @@ export const Name = (profile) => (
...
-        </Label> }
-        { profile.policy && <GreySmallText>{ profile.name }</GreySmallText> }
+        { profile.policy
+            ? <React.Fragment>
+                <PolicyPopover { ...{ profile, position: 'right' } } />
+                <GreySmallText>{ profile.policyType }</GreySmallText>
+            </React.Fragment>
+
+            : <Label color="red">External</Label>
+        }
```
akofink, line 29: The table looks good, but the cards are still incorrect:

image

image

vkrizan: Isn't it out of scope? We're going to enable the table feature within its deployment or shortly after.

akofink: Are we? I thought it was still an open question with PM that we will resolve after our demo.

akofink:
I'm also still seeing the policy type instead of the policy name in the policies table (both policies below have custom names that are not shown):

image

vkrizan:
> I'm also still seeing the policy type instead of the policy name in the policies table (both policies below have custom names that are not shown):

@akofink It's a profile name to be exact. This only fixes Report pages. We'll fix it separately.

akofink:
Oh, gotcha! I thought it was policy table as well from the PR description
> on policy table

vkrizan:
Sorry, my bad, it supposed to be "on reports table".

src/PresentationalComponents/ReportsTable/Cells.js
vkrizan: Fixed:
Screenshot_2020-11-12 Compliance

akofink (approved):
Awesome, thanks @vkrizan! I guess we are going to disable the cards soon after all.

PR #826, opened 2020-11-13
fix(RHICOMPL-1104): policies table to show policy type policy name

vkrizan:
The policies table should show user-defined policy name and policy type.

Screenshot_2020-11-13 Compliance

Please note, that the policy name is still buggy because of caching (RHICOMPL-1077).


akofink (approved):
It works as expected. Thanks, @vkrizan!

PR #827, opened 2020-11-13
refactor(RHICOMPL-1105): contextual useFeature hook

vkrizan:
useFeature hook now uses global EnabledFeaturesContext.

This improves:
* parsing URL only once
* setting and checking stored values only once
* globally accessible context through useFeature hook

Also fixes:
* warnings on non-boolean value passed from a stored value
* `reportsTableView` constant (typo)

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -14,3 +14,27 @@ global.mount = mount;
 global.React = React;
 global.fetch = fetch;
 global.toJson = toJson;
+
+// https://github.com/facebook/jest/issues/2098#issuecomment-260733457
+const localStorageMock = (() => {
+    let store = {};
+
+    return {
+        getItem(key) {
+            return store[key] || null;
+        },
+        setItem(key, value) {
+            store[key] = value.toString();
+        },
+        clear() {
+            store = {};
+        }
+    };
+
+})();
+
+global.localStorageMock = localStorageMock;
+
+Object.defineProperty(window, 'localStorage', {
+    value: localStorageMock
+});
--- a/src/Utilities/EnabledFeatures.js
+++ b/src/Utilities/EnabledFeatures.js
@@ -1,0 +1,65 @@
+import React, { useState, useEffect } from 'react';
+import propTypes from 'prop-types';
+import { useLocation } from 'react-router-dom';
+import { features } from '@/constants';
+
+const LOCAL_STORE_FEATURE_PREFIX = 'insights:compliance';
+
+export const EnabledFeaturesContext = React.createContext(features);
+
+export const getStoredFeatures = () => {
+    let savedState = { ...features };
+    Object.keys(features).forEach((feature) => {
+        let stored = localStorage.getItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`);
+        if (stored) {
+            savedState[feature] = !!stored;
+        }
+    });
+    return savedState;
+};
+
+const storeFeatureFlag = (feature, value) => {
+    if (!value) {
+        console.log(`Removing feature setting of ${feature}`);
+        localStorage.removeItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`);
+    } else {
+        console.log(`Setting feature value for ${feature} to ${value}`);
+        localStorage.setItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`, value);
+    }
+};
+
+export const EnabledFeatures = ({ children }) => {
+    const [enabledFeatures, setEnabledFeatures] = useState(getStoredFeatures);
+    const { search } = useLocation();
+
+    useEffect(() => {
+        let setFeatures = {};
+
+        const urlParams = new URLSearchParams(search || '');
+        urlParams.forEach((value, feature) => {
+            let featureEnabled = value === 'enable';
+            if (features[feature] === undefined && !featureEnabled) {
+                return;
+            }
+
+            storeFeatureFlag(feature, featureEnabled);
+            setFeatures[feature] = featureEnabled;
+        });
+
+        if (Object.keys(setFeatures).length > 0) {
+            setEnabledFeatures({ ...enabledFeatures, ...setFeatures });
+        }
+    }, [search]);
+
+    return (
+        <EnabledFeaturesContext.Provider value={enabledFeatures}>
+            { children }
+        </EnabledFeaturesContext.Provider>
+    );
+};
+
+EnabledFeatures.propTypes = {
+    children: propTypes.node
+};
+
+export default EnabledFeatures;
--- a/src/App.js
+++ b/src/App.js
@@ -5,6 +5,7 @@ import { Routes } from './Routes';
 import { NotificationsPortal } from '@redhat-cloud-services/frontend-components-notifications';
 import '@redhat-cloud-services/frontend-components-notifications/index.css';
 import './App.scss';
+import { EnabledFeatures } from './Utilities/EnabledFeatures';
 
 const App = (props) => {
     const appNavClick = {
@@ -36,7 +37,9 @@ const App = (props) => {
     return (
         <React.Fragment>
             <NotificationsPortal />
-            <Routes childProps={props} />
+            <EnabledFeatures>
+                <Routes childProps={props} />
+            </EnabledFeatures>
         </React.Fragment>
     );
 };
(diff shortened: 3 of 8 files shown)


src/Utilities/EnabledFeatures.js
```diff
@@ -0,0 +1,65 @@
...
+    if (!value) {
+        console.log(`Removing feature setting of ${feature}`);
+        localStorage.removeItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`);
+    } else {
+        console.log(`Setting feature value for ${feature} to ${value}`);
+        localStorage.setItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`, value);
+    }
+};
+
+export const EnabledFeatures = ({ children }) => {
```
akofink, line 31: :100:

akofink (approved):
Nice refactoring! It continues to work as expected, and the warnings have disappeared. I'll defer to @bastilian for a more detailed code review, but the new context seems really nice :)

src/Utilities/EnabledFeatures.js
```diff
@@ -0,0 +1,65 @@
...
+    const [enabledFeatures, setEnabledFeatures] = useState(getStoredFeatures);
+    const { search } = useLocation();
+
+    useEffect(() => {
+        let setFeatures = {};
+
+        const urlParams = new URLSearchParams(search || '');
```
vkrizan, line 38: There is one difference with the previous code, and that is redirection. In this context-based version there is no redirection (the passed parameters will stay in location/URL until navigation). We can add it if you think it is needed.

src/Utilities/EnabledFeatures.js
```diff
@@ -0,0 +1,65 @@
...
+        console.log(`Removing feature setting of ${feature}`);
+        localStorage.removeItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`);
+    } else {
+        console.log(`Setting feature value for ${feature} to ${value}`);
+        localStorage.setItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`, value);
+    }
+};
+
+export const EnabledFeatures = ({ children }) => {
+    const [enabledFeatures, setEnabledFeatures] = useState(getStoredFeatures);
```
bastilian, line 32: We might not need to keep this in a state. It would only give us a benefit if we'd have a way to set the feature flags other than via the URL.

src/Utilities/EnabledFeatures.js
```diff
@@ -0,0 +1,65 @@
...
+import { useLocation } from 'react-router-dom';
+import { features } from '@/constants';
+
+const LOCAL_STORE_FEATURE_PREFIX = 'insights:compliance';
+
+export const EnabledFeaturesContext = React.createContext(features);
+
+const getStoredFeatures = () => {
+    let savedState = { ...features };
+    Object.keys(features).forEach((feature) => {
```
bastilian, line 12: Prefer `.map` & spread over `.forEach` & `X[Y] = Z`

vkrizan: Can you please write an example?

bastilian: This function could for example read like this:

```js
const getStoredFeatures = () => {
    const featuresEnabled = Object.assign(...Object.keys(features).map((feature) => (
        { [feature]: !!localStorage.getItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`) || false }
    )).filter((v) => !!v));
    return {
        ...features,
        ...featuresEnabled
    };
};
```

src/Utilities/EnabledFeatures.js
bastilian: The redirect was/is intended.
 
There are other URL dependent features on the platform that we should avoid to interfere with.

vkrizan: Sure, I can add it.

Currently (in CI) *any* URL query string does a redirect. I guess we already broke that interfere promise. :wink:

src/Utilities/EnabledFeatures.js
```diff
@@ -0,0 +1,65 @@
...
+        console.log(`Setting feature value for ${feature} to ${value}`);
+        localStorage.setItem(`${LOCAL_STORE_FEATURE_PREFIX}:${feature}`, value);
+    }
+};
+
+export const EnabledFeatures = ({ children }) => {
+    const [enabledFeatures, setEnabledFeatures] = useState(getStoredFeatures);
+    const { search } = useLocation();
+
+    useEffect(() => {
```
bastilian, line 35: This seems to just reimplement what was the `useFeature` hook.

Why not try to preserve most of the original hook and make it "context aware" and simply invoke it here?

src/Utilities/EnabledFeatures.js
bastilian: > Currently (in CI) any URL query string does a redirect. I guess we already broke that interfere promise. 😉
 Write

I'm not sure if that is actually a bug, but I'd need to (re)investigate that. I believe this is correct behaviour of react-router. If no component renders it will redirect. I think. For now there don't seem to be issues with that.

src/Utilities/EnabledFeatures.js
vkrizan: The implementation move from the hook here is the whole point of this. The hook would have to access a context an would have to be able to write to it. The problem of the hook was that each time it was used, the same was parsed over and over again. There was no global state, which this new context adds.

src/Utilities/EnabledFeatures.js
vkrizan: Sorry, but I don't find it more readable.

src/Utilities/EnabledFeatures.js
vkrizan: Partially true, but a state does not hurt. We have to realize that the useEffect (defined down) can happen asynchronously, hence the state. 

Further enhancements are also possible, like live feature views.

src/Utilities/EnabledFeatures.js
vkrizan: It redirects, because the original `useFeature` does not check available features, sets anything to local storage, and then redirects back (clearing query string). react-router does not impact it.

I believe that query string isn't used by react-router in our app.

config/setupTests.js
```diff
@@ -14,3 +14,27 @@ global.mount = mount;
...
+            store[key] = value.toString();
+        },
+        clear() {
+            store = {};
+        }
+    };
+
+})();
+
+global.localStorageMock = localStorageMock;
```
vkrizan, line 36: @bastilian Is this the right way to do global mocks?

src/Utilities/EnabledFeatures.js
bastilian: We will never need this.

src/Utilities/EnabledFeatures.js
bastilian: The implementation should be in a hook.

The point of using hooks is to keep Component bodies small and make them less complex.

src/Utilities/EnabledFeatures.js
bastilian: Using `forEach` and assigning properties in an object is an obsolete way of doing this.

We aim for JavaScript written towards ES6(+), and therefore should make use of `.map` (and other functions) and utilise spreading.

If you look around other places in this codebase as well as other projects in the Insights organisation you'll see that it is a common (and encouraged) pattern.

config/setupTests.js
bastilian: You can also only mock it in the tests themselves using `jest.mock`.

Setting up a global mock shouldn't be needed, since only one set of tests interacts with it.

vkrizan: I'm not sure if jest allows to mock globals, as per the issue https://github.com/facebook/jest/issues/2098#issuecomment-260733457

There is also a package to help with the localStorage (which I've skipped) https://www.npmjs.com/package/jest-localstorage-mock

bastilian: In test you should not mock a global, you'd mock `useFeature` and for the tests of hook related code `useLocation`.

vkrizan:
Converted to draft, after discussion with @bastilian. There are concerns about the implementation with a global context, that might possibly impact next deployment.

akofink:
@vkrizan @bastilian it sounds like there were some fundamental disagreements on this one. Should we continue to try and get it merged or close it?

bastilian:
The main objective of this was to do parsing of URL as well as only setting and checking stored values only once, this has meanwhile been addressed when correcting the hook to comply with the linter[1].

[1] https://github.com/RedHatInsights/compliance-frontend/commit/7a46666efe5e8e0eb3f79f572110e24e67f5e18a

akofink:
Ok! So we can close this one? @vkrizan?

PR #831, opened 2020-11-18
fix(CompliancePolicies): RHICOMPL-1128 Apollo error on refetch

vkrizan:
Essential identifier seemed to be missing.

Here is the error:
```
Error writing result to store for query:
Store error: the application attempted to write an object with no provided id
but the store already contains an id of Profile:UUID for this object.
The selectionSet that was trying to be written is: ...
```


bastilian (approved):
Looks good to me! Thank you @vkrizan!

akofink (approved):
:raised_hands: thank god - this is very annoying :laughing:

PR #832, opened 2020-11-18
feat(scoring): RHICOMPL-1085 report accurate scores for a policy

akofink:
Requires https://github.com/RedHatInsights/compliance-backend/pull/662

### DIFF
--- a/src/PresentationalComponents/ReportsTable/Filters.js
+++ b/src/PresentationalComponents/ReportsTable/Filters.js
@@ -44,8 +44,8 @@ export const policyComplianceFilter = [{
     type: conditionalFilterType.checkbox,
     label: 'Systems meeting compliance',
     filter: (profiles, values) => (
-        profiles.filter(({ totalHostCount, compliantHostCount }) => {
-            const compliantHostsPercent = Math.round((100 / totalHostCount) * compliantHostCount);
+        profiles.filter(({ testResultHostCount, compliantHostCount }) => {
+            const compliantHostsPercent = Math.round((100 / testResultHostCount) * compliantHostCount);
             const matching = values.map((value) => {
                 const [min, max] = value.split('-');
                 return compliantHostsPercent >= min && compliantHostsPercent <= max;
--- a/src/PresentationalComponents/ReportsTable/ReportsTable.js
+++ b/src/PresentationalComponents/ReportsTable/ReportsTable.js
@@ -25,8 +25,8 @@ const ReportsTable = ({ profiles }) => {
         {
             title: 'Systems meeting compliance',
             transforms: [sortable],
-            sortByFunction: ({ totalHostCount, compliantHostCount }) => (
-                (100 / totalHostCount) * compliantHostCount
+            sortByFunction: ({ testResultHostCount, compliantHostCount }) => (
+                (100 / testResultHostCount) * compliantHostCount
             )
         }
     ];
--- a/src/SmartComponents/Reports/Reports.js
+++ b/src/SmartComponents/Reports/Reports.js
@@ -21,6 +21,7 @@ query Profiles($filter: String!) {
                 description
                 policyType
                 totalHostCount
+                testResultHostCount
                 compliantHostCount
                 majorOsVersion
                 complianceThreshold
@@ -50,8 +51,6 @@ query Profiles($filter: String!) {
 const profilesFromEdges = (data) => (
     (data?.profiles?.edges || []).map((profile) => (
         profile.node
-    )).filter((profile) => (
-        profile.totalHostCount > 0
     ))
 );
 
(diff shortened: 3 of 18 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -29,7 +29,7 @@ export const lastScanned = ({ profiles = [] }) => {
 };
 
 export const compliant = ({ profiles = [] }) => (
-    profiles.every(profile => profile.compliant === true)
+    profiles.every(profile => profile.lastScanned === 'Never' || profile.compliant === true)
```
bastilian, line 32: Why does it check the lastScanned? Wouldn't a profile the has 'Never' been scanned count as un-compliant, or the other way around?

Also, it wouldn't be the worst to change `lastScanned` to return undefined and move the 'Never' to the rendering table component, but that is for another day.

akofink: `profiles` here includes profiles without any test results (`lastScanned === 'Never'`) because we have to show assigned policies in the systems table, not just those with test results. If a host doesn't have a test result for a profile, we can't say if it's compliant or non-compliant; we can only base the 'compliant' icon on existing test results.

Without this check, a system that scans against a different SSG than the initial policy profile will show as non-compliant because both policy profiles are returned, one with a test result, the other with no test result. There are many scenarios to test, so I may have missed one that this doesn't work for. Please let me know if you are able to see some incorrect behavior.

akofink: I guess, to clarify, this change fixes the icon next to the compliance score in the report details view:

image

src/PresentationalComponents/ReportsTable/ReportsTable.js
```diff
@@ -25,8 +25,8 @@ const ReportsTable = ({ profiles }) => {
         {
             title: 'Systems meeting compliance',
             transforms: [sortable],
-            sortByFunction: ({ totalHostCount, compliantHostCount }) => (
-                (100 / totalHostCount) * compliantHostCount
+            sortByFunction: ({ testResultHostCount, compliantHostCount }) => (
+                (100 / testResultHostCount) * compliantHostCount
```
vkrizan, line 29: Observation: This formula is being repeated.

src/SmartComponents/Reports/Reports.js
```diff
@@ -51,7 +52,7 @@ const profilesFromEdges = (data) => (
     (data?.profiles?.edges || []).map((profile) => (
         profile.node
     )).filter((profile) => (
-        profile.totalHostCount > 0
+        profile.testResultHostCount > 0
```
vkrizan, line 55: Observation: This filter seems to be ensured by the GQL filter:
```
const filter = `(has_policy_test_results = true AND external = false)
                OR (has_policy = false AND has_test_results = true)`;
```

src/store/Reducers/SystemStore.js
vkrizan: The use of the "Never" constant is a bit unfortunate. Could it rather check profile rules?

```
profiles.every(profile => !profile.rules || profile.rules.length == 0 || profile.compliant === true)
```

akofink: As @bastilian said, we can refactor `lastScanned` to return undefined so we don't need to check if it's `'Never'`. I initially wrote it to check rules but thought it wasn't as readable; the safe navigation operator doesn't help either. Also, if `rules` isn't requested in the GraphQL request and it's undefined, this logic makes it compliant. Checking an undefined lastScanned results in non-compliant which is the safer failure.

As an aside, I wish we had a "Never" constant instead of a string literal! I assume that's what you meant. ;) I'd be happy to add one.

src/PresentationalComponents/ReportsTable/ReportsTable.js
akofink: Where? I see one other place in [Cells.js](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/ReportsTable/Cells.js#L45). Any others?

src/PresentationalComponents/ReportsTable/Filters.js
```diff
@@ -44,8 +44,8 @@ export const policyComplianceFilter = [{
     type: conditionalFilterType.checkbox,
     label: 'Systems meeting compliance',
     filter: (profiles, values) => (
-        profiles.filter(({ totalHostCount, compliantHostCount }) => {
-            const compliantHostsPercent = Math.round((100 / totalHostCount) * compliantHostCount);
+        profiles.filter(({ testResultHostCount, compliantHostCount }) => {
+            const compliantHostsPercent = Math.round((100 / testResultHostCount) * compliantHostCount);
```
akofink, line 48: This was missed the first time. Now the filter by compliance score should work better :)

vkrizan (approved):
Thank you!

PR #833, opened 2020-11-18
fix(SystemsTable): RHICOMPL-1131 RHICOMPL-1032 policy names and filtering

vkrizan:
Deduplicates system's policy names and fixes the filter on a policy for systems reporting with a different/non-latest SSG.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -22,7 +22,7 @@ export const DEFAULT_SYSTEMS_FILTER_CONFIGURATION = [
 export const systemsPolicyFilterConfiguration = (policies) => ([{
     type: conditionalFilterType.checkbox,
     label: 'Policy',
-    filterString: (value) => (`profile_id = ${value}`),
+    filterString: (value) => `policy_id = ${value}`,
     items: policies.map((policy) => ({
         label: policy.name,
         value: policy.id
--- a/src/store/Reducers/SystemStore.js
+++ b/src/store/Reducers/SystemStore.js
@@ -47,9 +47,13 @@ export const score = ({ profiles = [] }) => {
 export const profileNames = (system) => {
     if (system === {}) { return ''; }
 
-    return system.profiles.map(
-        (profile) => `${profile.policy ? '' : '(External) ' }${profile.name}`
-    ).join(', ');
+    let names = system.profiles.map(
+        (profile) => (
+            profile.policy ? profile.policy.name : `(External) ${profile.name}`
+        )
+    );
+
+    return [...new Set(names)].sort().join(', ');
 };
 
 export const policiesCell = (system) => {
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -53,6 +53,10 @@ query getSystems($filter: String!, $policyId: ID, $perPage: Int, $page: Int) {
                     external
                     score
                     ssgVersion
+                    policy {
+                        id
+                        name
+                    }
                     rules {
                         refId
                         title
@@ -85,6 +89,7 @@ query getSystems($filter: String!, $policyId: ID, $perPage: Int, $page: Int) {
                     score
                     policy {
                         id
+                        name
                     }
                 }
             }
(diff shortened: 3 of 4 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -47,9 +47,13 @@ export const score = ({ profiles = [] }) => {
...
-    return system.profiles.map(
-        (profile) => `${profile.policy ? '' : '(External) ' }${profile.name}`
-    ).join(', ');
+    let names = system.profiles.map(
+        (profile) => (
+            profile.policy ? profile.policy.name : `(External) ${profile.name}`
+        )
+    );
+
+    return names.filter((name, index) => !names.includes(name, index + 1)).join(', ');
```
bastilian, line 56: What should this do?

bastilian: If it is de-duplicating then please don't do this here. This is data returned from the backend, where it should be de-duplicated and/or ensured it only returns uniq profiles.

src/constants.js
```diff
@@ -22,7 +22,7 @@ export const DEFAULT_SYSTEMS_FILTER_CONFIGURATION = [
 export const systemsPolicyFilterConfiguration = (policies) => ([{
     type: conditionalFilterType.checkbox,
     label: 'Policy',
-    filterString: (value) => (`profile_id = ${value}`),
+    filterString: (value) => (`(profile_id = ${value} or policy_id = ${value})`),
```
bastilian, line 25: Wouldn't it be better to pass the policy_id as a variable to the GraphQL query?

bastilian: Ohm, sorry, no. that would limit the profiles then.

vkrizan: Yes, that would be very ideal. Is the filter framework able to do that?

vkrizan: Result of this already part of a filter/search variable.

src/store/Reducers/SystemStore.js
vkrizan: Deduplication of policy/profile names.

src/constants.js
bastilian: > Yes, that would be very ideal. Is the filter framework able to do that?

The filters don't do the GraphQL querying. It only provides a string based upon the functions defined here to be used.

src/store/Reducers/SystemStore.js
bastilian: Again this should be done with either properly requesting via GraphQL and/or ensuring that the API returns only uniq profiles/policies. This is only hiding the issue, possibly a bug.

Ps.: De-Duplication in JavaScript the clever way: https://codewithhugo.com/es6-dedupe-set/

vkrizan: Backend returns the data correctly: all system's profiles. With the v1 API deduplication on backend side would demolish the purpose of the system profiles return value. There is no key for deduplication of profile objects, only a key for deduplicating names. Therefore this is done on a presentation layer, in frotend.

bastilian: Then let's make sure the frontend request it correctly and the backend can return the proper result.

vkrizan: Isn't this a concern of presentation? 
Can you please give some example?

bastilian: > Isn't this a concern of presentation?

Unless unavoidable, no. At least in my opinion. The view is to tell the controlling layer what data it needs, but it is not responsible for cultivating that data in most cases. An extreme example would be the backend returning all profiles regardless if assigned or not in the database and the frontend needs to filter down the list for each system. 

The frontends responsibility is to interact with backend data with data from the backend.

akofink: I think we can either alter the existing `profiles` relationship on a host or add another like `policyProfiles`. It is not used anywhere else from what I can tell. I've put together a draft that seems to work as expected: https://github.com/RedHatInsights/compliance-backend/pull/664. The frontend changes then become just changing the constants.js line from `profile_id` to `policy_id`, like this: https://github.com/RedHatInsights/compliance-frontend/compare/master...akofink:1116

@bastilian this sounds like more of what you were thinking?

bastilian: @akofink Yes, that sounds like a good approach! We should maybe do a review of all instances where we make use of the profiles association and make sure the backend returns the appropriate list fitting the context.

vkrizan:
Closing to re-do with more backend support.

PR #842, opened 2020-11-20
fix(SystemsTable): RHICOMPL-1131 filter by policy

vkrizan:
(Taken from #833)


akofink (commented):
Is there a backend component? It doesn't seem to work as expected:

Peek 2020-11-20 10-32

akofink (approved):
Thanks, @vkrizan! Works as expected (once I actually checked out the changes :P)

vkrizan:
@akofink Thanks for the review.

Now I value that we've spitted the Jiras by a separate concern. This is a good example, where it had the benefit.

PR #844, opened 2020-11-24
feat(FinishedCreatePolicy): RHICOMPL-1147 show errors

vkrizan:
Adds reasons list on create policy wizard in case there is an error.

Screenshot_2020-11-24 Compliance

Set width of messages and reasons container ensures that the text will
wrap at the with of the progress bar.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -39,6 +39,14 @@
     color: var(--pf-global--danger-color--100);
     font-weight: bold;
     text-align: left;
+    width: var(--pf-global--breakpoint--sm);
+}
+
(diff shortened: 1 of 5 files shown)


akofink (approved):
Works as expected, thanks! I noticed a small usability issue that we can report and fix separately (or in this PR, up to you): the 'Return to application' button isn't displayed on error, but that's the only thing the user can do at that point, so I think it'd make sense to always show it.

vkrizan:
@akofink Thanks for the review. Yeah, we could do the button separately. There is still an X at the top of the modal. :wink:

PR #847, opened 2020-11-26
fix(SystemsTable): RHICOMPL-1096 show unique policy names in the column 

vkrizan:
**Requires backend RedHatInsights/compliance-backend#670**

Fixes duplicate policy names on the "Policies" table column by using fetched policy names from backend.

External policies are temporary, taken from the list of profiles.

Screenshot_2020-12-01 Compliance

### DIFF
--- a/src/store/Reducers/SystemStore.test.js
+++ b/src/store/Reducers/SystemStore.test.js
@@ -2,7 +2,7 @@ import {
     compliant,
     hasOsInfo,
     lastScanned,
-    profileNames,
+    policyNames,
     systemsToInventoryEntities
 } from './SystemStore';
 import { systems, entities } from '@/__fixtures__/systems.js';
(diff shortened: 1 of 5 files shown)


src/store/Reducers/SystemStore.test.js
```diff
@@ -96,15 +96,15 @@ describe('.compliant', () => {
     });
 });
 
-describe('.profileNames', () => {
-    it('should include external if profile is external', () => {
+describe('.policyNames', () => {
+    it("should return all system's policies", () => {
```
akofink, line 100: Is this why the eslint rule is changed? Why not use single quotes and not change the rule?

akofink (approved):
Seems to work as expected! Just one minor comment from me.

src/store/Reducers/SystemStore.test.js
vkrizan: Yes, this is the line. There is a single quote, so it would need escaping. I did not find a reason for a strict lint rule on such cases, and changed it to match Rubocop's behavior.

vkrizan: I've dropped the eslint change and the apostrophe.

/cc @bastilian

src/store/Reducers/SystemStore.test.js
```diff
@@ -96,15 +96,20 @@ describe('.compliant', () => {
...
+                { name: 'CIS', policy: null }
+            ],
+            policies: [
                 { name: 'HIPAA' },
-                { policy: {}, name: 'PCI' }
+                { name: 'PCI Policy' }
             ]
         };
-        expect(profileNames(system)).toEqual('(External) HIPAA, PCI');
+        expect(policyNames(system)).toEqual('HIPAA, PCI Policy, (External) CIS');
```
bastilian, line 112: We could start using `toMatchInlineSnapshot`, but it might only be useful for more complex strings, but not so big to qualify for a real snapshot: https://jestjs.io/docs/en/snapshot-testing#inline-snapshots

vkrizan: Nice. Good to know.

akofink (approved):
Thanks, @vkrizan!

PR #849, opened 2020-11-27
fix(SystemsTable): RHICOMPL-1137 show reported SSG version(s)

vkrizan:
**Requires backend RedHatInsights/compliance-backend#670**

Fixes showing only reported SSG versions in the table by pulling testResultProfiles out of backend, instead of combined
`profiles` field (assigned and test result profiles).

### DIFF
--- a/src/store/Reducers/SystemStore.js
+++ b/src/store/Reducers/SystemStore.js
@@ -22,7 +22,7 @@ import Truncate from 'react-truncate';
 
 const NEVER = 'Never';
 
-export const lastScanned = ({ profiles = [] }) => {
+export const lastScanned = ({ testResultProfiles: profiles = [] }) => {
     const dates = profiles.map((profile) => new Date(profile.lastScanned));
     const last = new Date(Math.max.apply(null, dates.filter((date) => isFinite(date))));
     const result = (last instanceof Date && isFinite(last)) ? last : NEVER;
@@ -30,11 +30,11 @@ export const lastScanned = ({ profiles = [] }) => {
     return result;
 };
 
-export const compliant = ({ profiles = [] }) => (
+export const compliant = ({ testResultProfiles: profiles = [] }) => (
     profiles.every(profile => profile.lastScanned === NEVER || profile.compliant === true)
 );
 
-export const score = ({ profiles = [] }) => {
+export const score = ({ testResultProfiles: profiles = [] }) => {
     const scoreTotal = profiles.reduce((acc, profile) => acc + profile.score, 0);
     const numScored = profiles.reduce((acc, profile) => {
         if (profilesRulesPassed([profile]).length + profilesRulesFailed([profile]).length > 0) { return acc + 1; }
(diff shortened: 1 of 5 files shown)


akofink (approved):
Works well, thanks!

src/store/Reducers/SystemStore.test.js
```diff
@@ -96,15 +96,15 @@ describe('.compliant', () => {
     });
 });
 
-describe('.profileNames', () => {
-    it('should include external if profile is external', () => {
+describe('.policyNames', () => {
+    it("should return all system's policies", () => {
```
bastilian, line 100: There is no "system's"

.eslintrc.yml
```diff
@@ -93,6 +93,7 @@ rules:
     - error
     - single
     - allowTemplateLiterals: true
+      avoidEscape: true
```
bastilian, line 96: What is the reason for enabling this rule?

src/store/Reducers/SystemStore.js
```diff
@@ -107,16 +105,27 @@ export const systemsToInventoryEntities = (systems, entities, showAllSystems, se
...
-        matchingSystem.rulesFailed = profilesRulesFailed(matchingSystem.profiles).length;
+        matchingSystem.policyNames = policyNames(matchingSystem);
+        matchingSystem.rulesPassed = profilesRulesPassed(matchingSystem.testResultProfiles).length;
+        matchingSystem.rulesFailed = profilesRulesFailed(matchingSystem.testResultProfiles).length;
         matchingSystem.lastScanned = lastScanned(matchingSystem);
         matchingSystem.compliant = compliant(matchingSystem);
         matchingSystem.score = score(matchingSystem);
 
+        let detailsLink;
+        if (matchingSystem.testResultProfiles && matchingSystem.testResultProfiles.length > 0) {
```
bastilian, line 119: This could be put into a "cell" function similar to `policiesCell`

src/store/Reducers/SystemStore.js
```diff
@@ -46,27 +46,25 @@ export const score = ({ profiles = [] }) => {
...
 };
 
-export const profileNames = (system) => {
+export const policyNames = (system) => {
     if (system === {}) { return ''; }
 
-    return system.profiles.map(
-        (profile) => `${profile.policy ? '' : '(External) ' }${profile.name}`
-    ).join(', ');
+    return system.policies.map(({ name }) => name).join(', ');
```
bastilian, line 52: We will still need to differentiate between internal and external for the foreseeable future.

.eslintrc.yml
vkrizan: Please see https://github.com/RedHatInsights/compliance-frontend/pull/847#discussion_r532726750

src/store/Reducers/SystemStore.js
vkrizan: :+1:

src/store/Reducers/SystemStore.js
vkrizan: Indeed. Will change.

src/store/Reducers/SystemStore.test.js
bastilian: Just to elaborate, since it might sound harsh :D: There is in some cases, but this is a possessive case in which a `'` should not be added.

vkrizan: You mean that the apostrophe is actually not welcome here? I see your point. The test framework replaces such characters to form a test id. I can gladly remove it.

src/store/Reducers/SystemStore.js
vkrizan: Done with peeking at profiles and checking absence of their policy.

src/store/Reducers/SystemStore.js
vkrizan: Updated. Thank you.

PR #857, opened 2020-12-02
feat(reports): RHICOMPL-1179 Update external support message

akofink:
With the next deployment, reports are no longer parsed. Warn the user
that this is the case and that external reports will eventually be
removed.


bastilian (approved):
LGTM! Thank you @akofink !

vkrizan (approved):
Thank you.

PR #860, opened 2020-12-04
feat(Routes): RHICOMPL-830 - Implement consistent title

bastilian:
Implements support for route titles and sets consistent titles for each route.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,19 +1,24 @@
 import React, { lazy } from 'react';
 import Router from './Utilities/Router';
 
+const defaultReportTitle = 'Reports';
 const reportsRoutes = [
     {
         path: '/reports',
+        title: defaultReportTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "Reports" */ './SmartComponents/Reports/Reports')
         )
     }, {
         path: '/reports/:report_id',
+        title: `$entityTitle - ${ defaultReportTitle }`,
+        defaultTitle: defaultReportTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "ReportDetails" */ 'SmartComponents/ReportDetails/ReportDetails')
         )
     }, {
         path: '/reports/:report_id/delete',
+        title: `Delete report - ${ defaultReportTitle }`,
         component: lazy(() =>
             import(/* webpackChunkName: "DeleteReport" */ 'SmartComponents/DeleteReport/DeleteReport')
         ),
@@ -21,31 +26,39 @@ const reportsRoutes = [
     }
 ];
 
+const defaultPoliciesTitle = 'SCAP policies';
 const policiesRoutes = [
     {
         path: '/scappolicies',
+        title: defaultPoliciesTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "CompliancePolicies" */ 'SmartComponents/CompliancePolicies/CompliancePolicies')
         )
     }, {
         path: '/scappolicies/new',
+        title: defaultPoliciesTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "CreatePolicy" */ 'SmartComponents/CreatePolicy/CreatePolicy')
         ),
         modal: true
     }, {
         path: '/scappolicies/:policy_id',
+        title: `$entityTitle - ${ defaultPoliciesTitle }`,
+        defaultTitle: defaultPoliciesTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "PolicyDetails" */ 'SmartComponents/PolicyDetails/PolicyDetails')
         )
     }, {
         path: '/scappolicies/:policy_id/edit',
+        title: `$entityTitle - ${ defaultPoliciesTitle }`,
+        defaultTitle: defaultPoliciesTitle,
         component: lazy(() =>
             import(/* webpackChunkName: "EditPolicy" */ 'SmartComponents/EditPolicy/EditPolicy')
         ),
         modal: true
     }, {
         path: '/scappolicies/:policy_id/delete',
+        title: `Delete policy - ${ defaultPoliciesTitle }`,
         component: lazy(() =>
             import(/* webpackChunkName: "DeletePolicy" */ 'SmartComponents/DeletePolicy/DeletePolicy')
         ),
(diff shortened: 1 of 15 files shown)


src/Routes.js
```diff
@@ -2,66 +2,89 @@ import React from 'react';
...
     }
 ];
 
+const defaultSystemsTitlePrefix = 'Systems';
+const defaultSystemsTitleSuffix = 'Compliance systems';
+const defaultSystemsTitle = `${ defaultSystemsTitlePrefix }: Compliance systems - ${ defaultSystemsTitleSuffix }`;
 const systemsRoutes = [
     {
         path: '/systems',
+        title: defaultSystemsTitle,
```
vkrizan, line 80: This one draws: `Systems: Compliance systems - Compliance systems - Compliance | Red Hat Insights`.

Similar to my previous comment, this should be:
```
Systems - Compliance | Red Hat Insights
```
and the header should match the top nav (`Compliance systems` -> `Systems`) to make it less granular.

/cc @katierik

src/Routes.js
```diff
@@ -2,66 +2,89 @@ import React from 'react';
...
 const reportsRoutes = [
     {
         path: '/reports',
+        title: defaultReportTitle,
         component: asyncComponent(() =>
             import(/* webpackChunkName: "Reports" */ './SmartComponents/Reports/Reports')
         )
     }, {
         path: '/reports/:report_id',
+        title: `${ defaultReportTitlePrefix }: $entityTitle - ${ defaultReportTitleSuffix }`,
```
vkrizan, line 17: Subpages should follow guideline structure
> Page header - Sub nav app - Top nav app | Bundle`

This would make it less granular and cleaner.

For example:
```
My RHEL7 HIPAA - Reports - Compliance | Red Hat Insights
```

src/Routes.js
```diff
@@ -2,66 +2,89 @@ import React from 'react';
 import Router from './Utilities/Router';
 import asyncComponent from './Utilities/asyncComponent';
 
+const defaultReportTitlePrefix = 'Reports';
+const defaultReportTitleSuffix = `Compliance reports`;
```
vkrizan, line 6: Header could match the navigation item (less granular and cleaner):
`Compliance reports` -> `Reports`

Page title end result:
```
Reports - Compliance | Red Hat Insights
```

src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -93,4 +97,8 @@ export const EditPolicy = () => {
     </Modal>;
 };
 
+EditPolicy.propTypes = {
+    route: propTypes.object
```
vkrizan, line 101: Can the `route` be in a context accessible globally on the scope instead of passing it via props downwards?

vkrizan (commented):
Thank you Sebastian.

Titles work as coded. They are very granular with the parts being repeated.

src/SmartComponents/EditPolicy/EditPolicy.js
bastilian: Using a context would add unnecessary complexity. We do not need this to be accessible globally for now.

bastilian:
@vkrizan Thanks for testing this. I went over the title definitions again and they should now align better with the guidlines given in the document linked in the Jira ticket. Some may still need to be changed, as some were not explicitly mentioned, but we should go over these with UX and PM.

src/Routes.js
```diff
@@ -2,66 +2,89 @@ import React from 'react';
...
     }
 ];
 
+const defaultPoliciesTitlePrefix = 'SCAP policies';
+const defaultPoliciesTitleSuffix = defaultPoliciesTitlePrefix;
+const defaultPoliciesTitle = `${ defaultPoliciesTitlePrefix }: ${ defaultPoliciesTitleSuffix }`;
 const policiesRoutes = [
     {
         path: '/scappolicies',
+        title: defaultPoliciesTitle,
```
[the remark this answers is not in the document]
katierik: What you have there as defined in the guidelines is correct.

src/Routes.js
katierik: Yup!

src/Routes.js
vkrizan: Note, that for the headers there is Jira RHICOMPL-910.

src/Routes.js
vkrizan: Note, that for the headers there is Jira RHICOMPL-910.

src/Routes.js
```diff
@@ -1,66 +1,83 @@
 import React, { lazy } from 'react';
 import Router from './Utilities/Router';
 
+const defaultReportTitle = 'Compliance reports';
```
vkrizan, line 4: Can this be "Reports" to match the header and navigation?

src/Utilities/hooks/useDocumentTitle.js
```diff
@@ -0,0 +1,26 @@
...
+const defaultTitleSuffixed = (title) => (
+    `${ title }${ DEFAULT_TITLE_SUFFIX }`
+);
+
+export const useTitleEntity = (route, entityTitle) => {
+    const location = useLocation();
+
+    useEffect(() => {
+        const title = entityTitle ?
+            route.title.replace('$entityTitle', entityTitle) : route.defaultTitle;
```
vkrizan, line 15: Observation: Custom templating system.

Easiest could be to prepend the entityTitle to the exisisting default title.
More advanced variant could be to keep some structure, into which the entityTitle (or a sub title) is pushed. Then it would draw the structure into a string.

vkrizan (approved):
Thanks again!

src/Utilities/hooks/useDocumentTitle.js
bastilian: If it was a framework/library, used by others or we'd have a lot more pages to take care of, then it would justify a more sophisticated solution. This works as intended and gives us a base to extent upon. If we'd need a more capable solution we would probably be better off investigating existing solutions.

src/Routes.js
bastilian: Changed.

src/Routes.js
```diff
@@ -1,66 +1,83 @@
...
         path: '/scappolicies/:policy_id/delete',
+        title: `Delete policy - ${ defaultPoliciesTitle }`,
         component: lazy(() =>
             import(/* webpackChunkName: "DeletePolicy" */ 'SmartComponents/DeletePolicy/DeletePolicy')
         ),
         modal: true
     }
 ];
 
+const defaultSystemsTitle = 'Compliance systems';
```
akofink, line 69: This should just be `Systems`, right? Check the mock: https://marvelapp.com/prototype/2304ie16/screen/68634890

bastilian: Changed. The page header title is now also just "Systems" to match the guideline to match these.

akofink (approved):
Thanks, @bastilian! Works as expected.

PR #861, opened 2020-12-04
refactor(Router): Use React.lazy instead of custom async loader

bastilian:


vkrizan (approved):
Nice refactoring :scissors: :+1: 

Would you mind creating a tracking Jira?

PR #862, opened 2020-12-07
fix(Reports): RHICOMPL-1189 remediation button

vkrizan:
The remediation button on Report detail page was not working due to
incorrect read from the system's data.  Instead of `profiles` the
rules passed to the remediation button should be collected from
`testResultProfiles`.


akofink (approved):
Makes sense, thanks!

vkrizan:
Need to fix the ComplianceButton by providing `profiles` within a system.

PR #863, opened 2020-12-08
fix(donut): RHICOMPL-1197 Empty donuts are gray

akofink:
So, the patternfly donut chart doesn't behave as expected (as shown in the patternfly demo/docs site). This hacks it to look as expected with 0 compliant hosts. The solution matches what we've done in the dashboard as well (copied from Advisor). I'm only fixing it here now because the dashboard folks wanted it fixed there before merging the supportability work.

Before:
image

After:
image

### DIFF
--- a/src/SmartComponents/ReportDetails/ReportDetails.js
+++ b/src/SmartComponents/ReportDetails/ReportDetails.js
@@ -1,11 +1,16 @@
 /* eslint-disable react/display-name */
 import React from 'react';
+import {
+    global_palette_black_300 as black300,
+    chart_color_blue_200 as blue200,
+    chart_color_blue_300 as blue300
+} from '@patternfly/react-tokens';
 import propTypes from 'prop-types';
 import { useParams } from 'react-router-dom';
 import { useQuery } from '@apollo/react-hooks';
 import gql from 'graphql-tag';
 
-import { ChartDonut, ChartThemeColor, ChartThemeVariant } from '@patternfly/react-charts';
+import { ChartDonut, ChartThemeVariant } from '@patternfly/react-charts';
 import { Breadcrumb, BreadcrumbItem, Button, Grid, GridItem, Text } from '@patternfly/react-core';
 
 import {
@@ -60,6 +65,7 @@ export const ReportDetails = ({ route }) => {
     });
     let donutValues = [];
     let donutId = 'loading-donut';
+    let chartColorScale;
     let profile = {};
     let policyName;
     let legendData;
@@ -73,9 +79,11 @@ export const ReportDetails = ({ route }) => {
         const testResultHostCount = profile.testResultHostCount;
         donutId = profile.name.replace(/ /g, '');
         donutValues = [
-            { x: 'Compliant', y: compliantHostCount },
+            { x: 'Compliant', y: testResultHostCount ? compliantHostCount : '0' },
             { x: 'Non-compliant', y: testResultHostCount - compliantHostCount }
         ];
+        chartColorScale = testResultHostCount === 0 && [black300.value] ||
+            [blue300.value, blue200.value];
         legendData = [
             { name: donutValues[0].y + ' ' + pluralize(donutValues[0].y, 'system') + ' compliant' },
             { name: donutValues[1].y + ' ' + pluralize(donutValues[1].y, 'system') + ' non-compliant' }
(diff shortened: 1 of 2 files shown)


src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -73,9 +79,11 @@ export const ReportDetails = ({ route }) => {
         const testResultHostCount = profile.testResultHostCount;
         donutId = profile.name.replace(/ /g, '');
         donutValues = [
-            { x: 'Compliant', y: compliantHostCount },
+            { x: 'Compliant', y: testResultHostCount && compliantHostCount || '0' },
```
akofink, line 82: Updated here based on @vkrizan's suggestion to make it a bit more concise.

We explored the [ChartDonutUtilization](https://www.patternfly.org/2020.06/documentation/react/charts/chartdonututilization/) component a bit. It seems to handle the `0` case, unlike the `ChartDonut` component, but we've decided not to swap them for now.

vkrizan: These ANDs and ORs combined... :see_no_evil: 

Anyhow, since `'0'` is interpreted as 100% and there is "Compliant" label, it looks like its Compliant when hovered. I guess we cannot win with this. :smile: I guess there is a reason why 0/0 isn't displayed.

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -73,9 +79,11 @@ export const ReportDetails = ({ route }) => {
...
             { x: 'Non-compliant', y: testResultHostCount - compliantHostCount }
         ];
+        chartColorScale = testResultHostCount === 0 && [black300.value] ||
```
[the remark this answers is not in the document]
vkrizan: The Patternfly's utilization chart seems to use lighter color.

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -154,8 +162,8 @@ export const ReportDetails = ({ route }) => {
                                     identifier={donutId}
                                     title={compliancePercentage}
                                     subTitle="Compliant"
-                                    themeColor={ChartThemeColor.blue}
                                     themeVariant={ChartThemeVariant.light}
+                                    colorScale={chartColorScale}
```
vkrizan, line 166: This now has different, unexpected, outcome...

Actual:
Screenshot_2020-12-09 My RHEL7 HIPAA - Reports - Compliance Red Hat Insights(1)

Expected (previous):
Screenshot_2020-12-09 My RHEL7 HIPAA - Reports - Compliance Red Hat Insights

vkrizan (commented):
The 0/0 works correctly, but the normal cases show different result. :confused:

src/SmartComponents/ReportDetails/ReportDetails.js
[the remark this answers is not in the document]
bastilian: > Ternary operator could make it bit more readable

Readability is up to who reads it. Both styles should be familiar by now and until there is any solid argument against or for a certain style both should be valid and it should be up the the code-authors preference.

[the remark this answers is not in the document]
akofink: It seems like every time I use one, I get asked to use the other! lol

src/SmartComponents/ReportDetails/ReportDetails.js
akofink: @vkrizan sorry, how did you get into this state? It seems to work fine for me in this case (0 compliant, 1 non-compliant, 1 unsupported):

image

akofink: Ok! So this is a place where using boolean logic is problematic, and a ternary operator is actually what we want ;) Updated!

vkrizan (approved):
Works like a charm. Thanks!

src/SmartComponents/ReportDetails/ReportDetails.js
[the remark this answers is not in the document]
vkrizan: The rationale is a Cognitive Complexity. If there is a branching `if-else` is preferred over perplexing statements with just ANDs and ORs. Both variants would do the same, but one quicker to comprehend and reduces possible maintenance issues. I'd recommend reading https://redirect.sonarsource.com/doc/cognitive-complexity.html

PR #864, opened 2020-12-08
fix(ReportsTable): RHICOMPL-1134 Fix policyType filter

akofink:
This uses the newly available policyType attribute on the profile
GraphQL API to filter in the same way as operating systems.

It also consolidates some logic with the operating systems filter.


src/PresentationalComponents/ReportsTable/__snapshots__/Filters.test.js.snap
```diff
@@ -360,175 +360,4 @@ Array [
...
-        },
-      ],
-    },
-    "refId": "xccdf_org.ssgproject.content_profile_pci-dss3",
-    "testResultHostCount": 8,
-    "totalHostCount": 10,
-  },
-]
-`;
+exports[`policyTypeFilter expect to return results 1`] = `Array []`;
```
bastilian, line 363: Mind adding the `policyType` field to the fixtures in `policies.js`? That should allow the test to take a snapshots with results.

src/PresentationalComponents/ReportsTable/Filters.js
```diff
@@ -14,15 +14,15 @@ export const policyNameFilter = [{
...
     filter: (profiles, values) => (
-        profiles.filter(({ policy }) => values.includes(policy?.id))
+        profiles.filter(({ policyType }) => values.includes(policyType))
     ),
-    items: policies.map((policy) => ({
-        label: policy.name,
-        value: policy.id
+    items: policyTypes.map((policyType) => ({
+        label: policyType,
+        value: policyType
```
bastilian, line 25: Just as a note: This works, but could we eventually have a `policyType` object returned in the API, so we could use a proper id for the values?

vkrizan: Proper object identification would be good here, I agree. Another reason of having policy types as a model. :wink:

src/PresentationalComponents/ReportsTable/ReportsTable.js
```diff
@@ -30,10 +30,9 @@ const ReportsTable = ({ profiles }) => {
             )
         }
     ];
-    const policyTypes = profiles.map(({ policy }) => (policy)).filter((i) => (!!i));
-    const operatingSystems = [...new Set(
-        profiles.map(({ majorOsVersion }) => (majorOsVersion)).filter((i) => (!!i))
-    )];
+    const uniq = (collection) => [...new Set(collection)];
```
vkrizan, line 33: Nitpick: We might think of extracting this to some util function. :slightly_smiling_face:

vkrizan (approved):
Besides the test, it looks and works good. Thank you.

src/PresentationalComponents/ReportsTable/Filters.js
akofink: Sorry, but why is it problematic to have label and value be the same? This is the same way we filter operating systems, no ID required.

bastilian: I don't think it's problematic at all, an id would just be nicer, eventually if we decide it makes sense to return an object. From my side it was really just a note how we could improve it in the future.

vkrizan: @akofink have you tried filtering by a value that contains "and" or quotes? I'm not sure whether it escapes these characters for the `search`.

src/PresentationalComponents/ReportsTable/ReportsTable.js
akofink: Agreed - this is a utility function, but I guess you mean we should put it somewhere globally accessible. Any recommendations for where to put it?

src/PresentationalComponents/ReportsTable/__snapshots__/Filters.test.js.snap
akofink: Argh! I always forget to do this :P Sorry, I've added it now.

src/PresentationalComponents/ReportsTable/ReportsTable.js
vkrizan: `src/Utilities/funcTools.js` maybe?

akofink: Ok, I was asking because I thought maybe we already had a place for this. I've created `Utilities/helpers.js` for this now.

src/PresentationalComponents/ReportsTable/Filters.js
akofink: Good point, @vkrizan, I'll try it out.

akofink: @vkrizan so... this doesn't actually hit the API at all. It's a frontend only table filter, just like the operating system attribute. This table doesn't even appear to have pagination, though. I tested it with 26 rows, and still nothing. I think we're OK here.

vkrizan: :+1: 

The problem would be on those filter that reach to API.

PR #865, opened 2020-12-09
fix(ReportDetails): RHICOMPL-1138 Show hosts with policy results

akofink:
This changes the systems table to accept a defaultFilter prop which will
be added to user filters during the API request. This is to allow
specifying a new filter for the ReportDetails page to only show hosts
that have test results for a given policy (a set of profiles associated
to the policy). The new filter takes policy ID or any profile ID
associated to the policy. It also reports test result hosts for external policy profiles.

Requires https://github.com/RedHatInsights/compliance-backend/pull/678


src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -176,19 +176,15 @@ class SystemsTable extends React.Component {
...
+            this.filterBuilder.buildFilterString(activeFilters)
+        ].filter((e)=>!!e).join(' and ');
 
-        if (showOnlySystemsWithTestResults) {
-            filter = `has_test_results = true ${filter.length > 0 ? `and ${filter}` : ''}`;
-        }
-
-        let variables = { filter, perPage, page };
-        if (policyId) {
-            filter = `policy_id = ${policyId} and ${filter}`;
```
bastilian, line 189: Do we not want need to use this filter anymore?

vkrizan: Indeed. This affects showing assigned-only systems on a SCAP Policy. Mind to adding a default filter for that case (`PolicyDetails`). Also please check other usages, like policy edit.

vkrizan (changes requested):

vkrizan:
Works correctly on the reports page. Thanks.

src/SmartComponents/SystemsTable/SystemsTable.js
akofink: Whoops! Sorry, I meant to check for this and forgot :see_no_evil: Updated!

vkrizan (approved):
Seems to work on all pages. Thank you.

PR #867, opened 2020-12-09
fix(ReportsTable): RHICOMPL-1206 Properly indicate unsupported

akofink:
The external policy should show unsupported beneath the operating system
label only if there are any unsupported hosts results reporting against
the policy.

Note for screenshots below, SSG version 0.1.43 is supported on RHEL 7.7

Before:
image

After:
image

### DIFF
--- a/src/PresentationalComponents/ReportsTable/Cells.js
+++ b/src/PresentationalComponents/ReportsTable/Cells.js
@@ -24,21 +24,21 @@ Name.propTypes = {
     profile: propTypes.object
 };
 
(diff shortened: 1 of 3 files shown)


vkrizan (approved):
Nice. Thank you.

Interesting that no snapshots needed change. Probably there is no coverage?

akofink:
You are right! No coverage :see_no_evil: I've updated the associated test to cover this scenario.

bastilian (approved):

PR #878, opened 2020-12-15
feat(features): RHICOMPL-1009 - Enable SSG versions by default

bastilian:


vkrizan (approved):
Thanks!

src/SmartComponents/ReportDetails/ReportDetails.test.js
```diff
@@ -77,7 +77,7 @@ describe('ReportDetails', () => {
     };
 
     beforeEach(() => {
-        useFeature.mockImplementation(() => (false));
+        useFeature.mockImplementation(() => (true));
```
bastilian, line 80: We should have a default mock for this in `setupTests.js`, not mocking the localstorage underneath, but just the `useFeature` and implement it such a way that it reads the defaults and normally reads and writes to localstorage, but doesn't require a `Router`, for tests that should test a feature disabled by default.

akofink (approved):

PR #883, opened 2020-12-16
fix(SSGVersion): RHICOMPL-1233 Handle unknown SSGs

akofink:
Before a system has uploaded any reports, the SSG version returned from
the API is a blank string. The SSGVersion component should render
'Not available' in that case.

image

### DIFF
--- a/src/SmartComponents/PolicyDetails/PolicySystemsTab.js
+++ b/src/SmartComponents/PolicyDetails/PolicySystemsTab.js
@@ -32,8 +32,8 @@ const PolicySystemsTab = ({ policy, systemTableProps }) => {
             }, ...showSsgVersions ? [{
                 key: 'facts.compliance',
                 title: 'SSG version',
-                renderFunc: (_name, _id, profile) => (
-                    <Cells.SSGVersion { ...{ profile } } />
+                renderFunc: (profile) => (
+                    <Cells.SSGVersion supported={ profile.supported } ssgVersion={ profile.ssg_version } />
                 )
             }] : []]}
             complianceThreshold={ policy.complianceThreshold }
--- a/src/SmartComponents/ReportDetails/ReportDetails.js
+++ b/src/SmartComponents/ReportDetails/ReportDetails.js
@@ -117,8 +117,8 @@ export const ReportDetails = ({ route }) => {
         props: {
             width: 5
         },
-        renderFunc: (_name, _id, profile) => (
-            profile && <Cells.SSGVersion { ...{ profile } } />
+        renderFunc: (profile) => (
+            profile && <Cells.SSGVersion supported={ profile.supported } ssgVersion={ profile.ssg_version } />
         )
     }] : [], {
         key: 'facts.compliance.rules_failed',
--- a/src/SmartComponents/SystemsTable/Cells.js
+++ b/src/SmartComponents/SystemsTable/Cells.js
@@ -2,15 +2,17 @@ import React from 'react';
 import propTypes from 'prop-types';
 import { UnsupportedSSGVersion } from 'PresentationalComponents';
 
-const SSGVersion = ({ profile }) => (
-    profile.supported && profile.ssgVersion ||
-    <UnsupportedSSGVersion messageVariant='singular'>
-        { profile.ssgVersion }
-    </UnsupportedSSGVersion>
-);
+const SSGVersion = ({ supported, ssgVersion }) => {
+    ssgVersion ||= 'Not available';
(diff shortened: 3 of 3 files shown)


src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -2,12 +2,13 @@ import React from 'react';
 import propTypes from 'prop-types';
 import { UnsupportedSSGVersion } from 'PresentationalComponents';
 
-const SSGVersion = ({ profile }) => (
-    profile.supported && profile.ssg_version ||
+const SSGVersion = ({ profile }) => {
+    let ssgVersion = profile.ssg_version || 'Not available';
```
vkrizan, line 6: Isn't the attribute `profile.ssgVersion` (a camel case one)?

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -2,12 +2,13 @@ import React from 'react';
...
+    return profile.supported && ssgVersion ||
```
vkrizan, line 7: Observation: This has a side-effect if it would be supported but no version would be reported. That's why I'd prefer using `if-else` (ternary operator). One can argue that the case would never happen, but it is good to account for it nevertheless to avoid future bugs.

akofink: That is exactly what was happening to cause this bug. I almost changed it to ternary with this change; I will update it.

src/SmartComponents/SystemsTable/Cells.js
akofink: I'm not sure why, but the SystemStore defines all its attributes with snake_case. :shrug:

bastilian: I head to change the property to `ssgVersion` in the inventory refactoring as well and I'm really confused when or how we changed that. `ssgVersion` seems to be the one working properly. 🤷

vkrizan: Oh, I see it now. The `profile` is actually `facts.compliance` where `ssg_version` is set. 
Would you mind renaming the variable to avoid further confusion? Something like `complianceFacts` as it is not a profile object. (The same name is used in `renderFunc`.)

vkrizan (approved):

src/SmartComponents/SystemsTable/Cells.js
akofink: Since we only need two props, `ssgVersion` and `supported`, I'll just add those as props instead of accepting a profile. Is that ok?

vkrizan: Sounds good!

akofink: Updated! Please check the ReportDetails page still works. I'm getting issues with the CompliacneRemediationButton that prevents it from loading for me locally for some reason.

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
```diff
@@ -32,8 +32,8 @@ const PolicySystemsTab = ({ policy, systemTableProps }) => {
             }, ...showSsgVersions ? [{
                 key: 'facts.compliance',
                 title: 'SSG version',
-                renderFunc: (_name, _id, profile) => (
-                    <Cells.SSGVersion { ...{ profile } } />
+                renderFunc: (profile) => (
+                    <Cells.SSGVersion supported={ profile.supported } ssgVersion={ profile.ssg_version } />
```
akofink, line 36: @bastilian `_name` and `_id` here came from the systems table refactor, but I'm not sure I understand how it works or why it's needed. The first argument seems to be the profile, so `_name` before was getting assigned to the profile. Any idea if this change here is correct or why we might still need `_name` and `_id`?

With this change, the policy system tab seems to work as expected.

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -117,8 +117,8 @@ export const ReportDetails = ({ route }) => {
         props: {
             width: 5
         },
-        renderFunc: (_name, _id, profile) => (
-            profile && <Cells.SSGVersion { ...{ profile } } />
+        renderFunc: (profile) => (
+            profile && <Cells.SSGVersion supported={ profile.supported } ssgVersion={ profile.ssg_version } />
```
akofink, line 121: Same thing here.

bastilian (approved):

PR #887, opened 2020-12-17
chore(deps): Bump @patternfly & @redhat-cloud-services packages

bastilian:


akofink (approved):
Nothing seems to break! :)

PR #890, opened 2020-12-17
fix(ReportDetails): RHICOMPL-913 -Update policy details

bastilian:
This updates the policy details on the report details page to align more with the mockups:

Screenshot 2020-12-17 at 10 59 23

### DIFF
--- a/src/PresentationalComponents/ReportDetailsDescription/ReportDetailsDescription.js
+++ b/src/PresentationalComponents/ReportDetailsDescription/ReportDetailsDescription.js
@@ -14,51 +14,71 @@ import {
     Link
 } from 'react-router-dom';
 
-const ReportDetailsDescription = ({ profile }) => (
-    <TextContent className='policy-details'>
-        { !profile.policy ? <React.Fragment>
-            <Text component={TextVariants.h2}>Policy details</Text><br/>
-            <Text component={TextVariants.p}>Policy not managed in cloud.redhat.com</Text>
-            <Text component={TextVariants.p}>
-                <a target='_blank' rel='noopener noreferrer'
-                    href='https://access.redhat.com/products/cloud_management_services_for_rhel'>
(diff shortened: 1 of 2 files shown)


src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -204,6 +204,7 @@ export const ReportDetails = ({ route }) => {
                         <SystemsTable
                             showOnlySystemsWithTestResults
                             compliantFilter
+                            remediationsEnabled={ false }
```
vkrizan, line 207: Is this a deliberate change?

src/PresentationalComponents/ReportDetailsDescription/ReportDetailsDescription.js
```diff
@@ -14,51 +14,72 @@ import {
...
+            <Dt>
+                <Text className='ins-c-non-bold-h2'>
+                    Policy details
+                </Text>
+            </Dt>
+            <Dt>
+                Operating system
+            </Dt>
+            <Dd>
+                RHEL { profile.majorOsVersion }
```
vkrizan, line 61: This is the only content change to match the mocks. Is that true?

vkrizan (commented):
The `SystemsTable` change seems like a mistake.

Would it be possible to highlight or isolate fixing changes from the refactoring?

Thank you.

src/SmartComponents/ReportDetails/ReportDetails.js
bastilian: Good catch. This was just accidentally committed. Removed it.

src/PresentationalComponents/ReportDetailsDescription/ReportDetailsDescription.js
bastilian: This is the only content change yes.

vkrizan (approved):

PR #893, opened 2020-12-17
fix(deps): Downgrade inventory-compliance

bastilian:
The latest release causes issues with the remediations button.


sthirugn (approved):

akofink (approved):
Thanks!

Reverts RHICOMPL-1207 and RHICOMPL-1124

PR #895, opened 2020-12-18
fix(PoliciesTable): RHICOMPL-1274 - Use common OSBadge

bastilian:
and remove SSG version


vkrizan (approved):
Looks great!

Screenshot_2020-12-18 SCAP policies - Compliance Red Hat Insights

PR #896, opened 2020-12-18
fix(CreateSCAPPolicy): RHICOMPL-1161 policy-wide profile In Use label

vkrizan:
The "In Used" label, placed on profiles that already exist, would work
policy-wide.  This prevents error from the backend and adds to UX.

It compares benchmark's ref_ids that are tigthly coupled with OS major
version.

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -41,7 +41,9 @@ query benchmarksAndProfiles {
             node {
                 id
                 refId
-                benchmarkId
+                benchmark {
+                    refId
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -51,8 +53,8 @@ query benchmarksAndProfiles {
 const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
     const { data, error, loading } = useQuery(BENCHMARKS_AND_PROFILES, { fetchPolicy: 'no-cache' });
 
-    const userProfileRefIdsForBenchmarkId = (profiles, benchmarkId) => (
-        profiles.filter(profile => benchmarkId === profile.node.benchmarkId).map(profile => profile.node.refId)
+    const inUseProfileRefIds = (profiles, benchmark) => (
+        profiles.filter(profile => benchmark.refId === profile.node.benchmark.refId).map(profile => profile.node.refId)
```
akofink, line 57: There's the edge case that a profile's ref_id can change between SSG versions, but we can continue ignoring that case for now. What we really want to compare is policy type.

akofink (approved):
Thanks, @vkrizan! Works as expected.

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
vkrizan: Indeed. :santa:

PR #898, opened 2020-12-20
chore(docker): Use RedHat UBI image

bastilian:
This replaces the Fedora image with the RHEL 8 UBI image as the base for docker.

(Shamelessly stolen from https://github.com/RedHatInsights/sources-ui)


akofink (approved):
Works beautifully! Thanks!

PR #902, opened 2021-01-04
Add alias to React dependency.

Hyperkid123:
### Changes
- force webpack to use compliance react instance. This will fix package linking.
- update inventory compliance to the latest version

Note: If you want, you can restrict it only to the dev webpack config. 

cc @bastilian @vkrizan 

Why the linking was not working?

Due to the mechanism of node dependency resolutions the linked packages is pulling their own dependencies from a different `node_modules` directory than from where you would expect. That causes the build to include 2 different **instances** of react. One was taken from the compliance frontend node_modules and the other from frontend-components node_modules.


vkrizan:
The error `Could not find "store" in the context of "Connect(ComplianceRemediationButton)".` is still present (on a Report detail page), even with redux and react-redux aliased.

Hyperkid123:
@vkrizan is right. I assumed once the empty state worked its all good. Unfortunately linking inventory components won't be possible until chrome 2.0 is ready, due to the custom dependencies consumption. :disappointed: Feel free to close this PR if you don't want these changes.

vkrizan:
This helped with the packaging problem that we had. I've also added alias for react-intl:
```
'react-intl': resolve(__dirname, '../node_modules/react-intl'),
```

I'll merge this and we can resolve the rest in a next PR.

Thank you @Hyperkid123!

vkrizan (approved):

PR #907, opened 2021-01-05
fix(deps): RHICOMPL-1277 add alias to react-intl

vkrizan:
Continuation of #902


karelhala (approved):
Can you also bump the version of `@redhat-cloud-services/frontend-components-inventory-compliance` to `2.2.22` as that includes the newer fixes for remediations button.

PR #923, opened 2021-01-11
feat(Reports): RHICOMPL-1230 Update external policy message

akofink:
This message is meant to ship at the same time we remove any remaining
external policies from Compliance. It will be removed in mid-Feb with
RHICOMPL-1298.


vkrizan (approved):
LGTM. Thanks!

PR #924, opened 2021-01-12
Migrate webpack 5

Hyperkid123:
### Changes
- migrate compliance to crc common webpack config
  - update to webpack 5
  - prepare compliance UI form chrome 2.0
- update react and react-dom dependencies
  - no breaking changes
  - required to unify react-test-renderer and react-dom
- use new Inventory component system
  - No longer uses window function but react component from FEC. This inventory is ready for chrome 2.0
- changed imports to use ESM modules instead of UMD
  -  reduces the overall build size by 2 MB (there is still some room for improvement but we need chrome 2.0 first)

@vkrizan @bastilian Can you please double-check the UI if nothing is broken? Custom CSS in particular.


Hyperkid123:
We have found one more bug with global filter which is supposed to be disabled but the filter still adds some data to the URL hash.

akofink (commented):
Thank you, @Hyperkid123! It seems to work as expected, though it's a very large change. I'd like at least a second reviewer and ideally @bastilian who is out until tomorrow.

Hyperkid123:
@akofink ofc. Take as much time as you need. We need to make sure it works properly.

config/dev.webpack.config.js
```diff
@@ -1,44 +1,36 @@
...
-    injectClient: hotReload(),
-    inline: hotReload(),
-    allowedHosts: [
-        'ci.foo.redhat.com',
-        'qa.foo.redhat.com',
-        'stage.foo.redhat.com',
-        'prod.foo.redhat.com',
-        default_host
-    ],
-    port: process.env.FRONTEND_PORT ? process.env.FRONTEND_PORT  : '8002',
```
bastilian, line 33: We will need this still for the docker-compose setup to work. I put it locally into `devserverConfig` which seems to be a good place to put this.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -196,12 +193,12 @@ class SystemsTable extends React.Component {
 
     updateSystems = () => {
         const prevSystems = this.props.systems.map((s) => s.node.id).sort();
-        return this.fetchSystems().then((items) => (
-            this.props.updateSystems({
+        return this.fetchSystems().then((items) => {
+            return this.props.updateSystems({
```
bastilian, line 197: Just a nitpick. :) Does this need an explicit return?

bastilian (approved):
Thank you @Hyperkid123! This really great!

Everything appears to be working correctly even with and without the `inventory-compliance` package linked.

There are two minor things, but we can address those later on as well.

config/dev.webpack.config.js
Hyperkid123: Yeah, that is exactly why I have created the file. And then just forgot to add the config :smile:

src/App.js
```diff
@@ -14,6 +15,7 @@ const App = (props) => {
     };
 
     useEffect(() => {
+        localStorage.setItem(`${LOCAL_STORE_FEATURE_PREFIX}:newInventory`, true);
```
bastilian, line 18: Oh, I almost missed that. This is not good, we can not yet enable the new Inventory implementation.

bastilian (changes requested):

src/App.js
Hyperkid123: Is there any specific reason? The old implementation causes some issues in the UI. It actually does not load the data and it's tuck in the loading state.

bastilian: @Hyperkid123 Yeah, the new implementation removes a few things we need before we can switch to not introduce regressions.

The issue you are referring to may be one that we are already aware of and it is marked to be resolved with moving over to the new implementation, but it is not a persistent bug and is resolved most of the times with a refresh.

I just tested it with removing this line and disabling the newInventory and all works as expected.

Hyperkid123: Ah ok. I was under the impression the migration caused the issue! Alright. I will remove it then

bastilian: Awesome! Thank you! The old implementation is buggy... :D

bastilian (approved):
Looks all good now! Thank you again @Hyperkid123! Awesome work!

Hyperkid123:
NP. Let me know if any issues surface.

PR #931, opened 2021-01-20
fix(Wizard): Add catch all in case response is incorrect

bastilian:


akofink (approved):
Thanks!

vkrizan (approved):
Looks good.

akofink:
@bastilian any reason we shouldn't merge this?

bastilian:
@akofink This was only a "quick fix", i'm not sure it this is the right solution as it is not tested well, but there is no reason to work on it more and ensure users always see a "nice" error message.

akofink:
Even if it's not tested well, the solution seems straightforward and shouldn't do any harm. IMO we should merge it and revisit a better solution later, or if it's no longer an issue (I don't think that's the case?), we can close this PR altogether.

vkrizan:
It's good enough to be merged, TBH. It does some small improvement.

PR #935, opened 2021-01-25
Improve build performance

Hyperkid123:
### Changes
- transform FEC imports to use ESM build
- update webpack config and disable hashes in dev build

cc @bastilian

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -30,7 +30,7 @@ module.exports = {
             'transform-imports', {
                 '@patternfly/react-icons': {
                     transform: (importName) => (
-                        `@patternfly/react-icons/dist/js/icons/${IconMapper[importName] || importName
+                        `@patternfly/react-icons/dist/esm/icons/${IconMapper[importName] || importName
                         .split(/(?=[A-Z])/)
                         .join('-')
                         .toLowerCase()}.js`
@@ -38,6 +38,19 @@ module.exports = {
                 }
             },
             'react-icons'
+        ], [
+            'transform-imports', {
+                '@redhat-cloud-services/frontend-components': {
+                    transform: '@redhat-cloud-services/frontend-components/components/esm',
+                    preventFullImport: true,
+                    skipDefaultConversion: true
+                },
+                '@redhat-cloud-services/frontend-components-notifications': {
+                    transform: '@redhat-cloud-services/frontend-components-notifications/esm',
+                    preventFullImport: true,
+                    skipDefaultConversion: true
+                }
+            }
         ]
 
     ]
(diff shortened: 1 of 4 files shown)


bastilian:
@Hyperkid123 Thank you! This does improve the build time! However it still has the issue of returning 404 after another re-compile. I was able to reproduce the issue on two machines.

The issue seems to be around the file hashes option. When hashes are enabled all seems to work as expected, but the build time increases again and "Loading chunk" errors occur more often.

bastilian:
Just in case we are not able to determine what the issue is that causes `webpack-dev-server` to throw 404s, the other changes still somewhat improve build times and iv v3 will even further improve it, we might be good.
It would still have the Load chunk errors, but that's better than the dev server not working properly at all. :D

akofink (commented):
I was also able to reproduce the 404 issue after changing a file in devel.

Hyperkid123:
@bastilian I will look at the issue with hashes. Disabling the chunk mapper in dev env might work. I think it has to go hand in hand with unhashed files.

bastilian:
@Hyperkid123 That would be amazing! Thank you! There is no rush though, for now it seems to work. I did see the 404 with hashes enabled, but just once for now, and only after a longer while with multiple compiles in between. really odd issue.

Hyperkid123:
@bastilian I see what is the issue. For some reason, after the webpack rebuild, the HTML template is just deleted. That is why we see the 404. There is no HTML to get. Hopefully, there will be a simple solution to this.

Hyperkid123:
So the HTML template is held in webpack memory, but we serve it from a disk. We need to find a way to keep it there between rebuilds.

bastilian:
> ok we are trolled by the CleanWebpackPlugin.

@Hyperkid123 OH! Nice catch! :)

Hyperkid123:
@bastilian @akofink it should be working now. Even without hashes. The issue was the following.

Normally on asset rebuild, the index.html would rebuild, because the bundle hash has changed, therefore the HTML template had to be rebuild. But when we removed the hash changes, we no longer need to rebuild the HTML file and it is marked as `stale`. And the webpack clean plugin is removing stale assets. I've added an option to the webpack config file to preserve stale assets so the file is no longer removed after some source file change. It should also resolve the issues with missing chunks. I suspect it had the same problem. Sometimes chunks are not rebuilding and their hashes are not changed so they are stale.

bastilian:
@Hyperkid123 Nice! Compile times are around ~40k ms for the initial compile and about half that on following compiles. 

I'll keep testing it throughout today while working and will merge later.

akofink (approved):
Seems to work fine for me, too. Seeing ~35s initial build time, ~4s after making a change.

PR #957, opened 2021-02-03
feat(devel): Add spandx config for local remediations

akofink:
I used this to test remediations locally with compliance when fixing the 500 error. I thought I'd go ahead and push it in case it's needed again.


vkrizan (approved):
Great. Thanks!

PR #972, opened 2021-02-10
Fix infinite loop

Hyperkid123:
### Changes
- update webpack config to use a stable release
- removes effect that causes infinite loops

cc @bastilian @karelhala


bastilian (approved):

bastilian:
Thank you @Hyperkid123!

PR #973, opened 2021-02-10
feat(CreatePolicy): RHICOMPL-701 filter systems by OS major version 

vkrizan:
Uses new `osMajorVersion` state in the wizard form to keep track of the OS major version for Systems to be filtered on. The `InventoryTable` gets passed the query to trigger the use of the filter. This would be refactored to `getEntities` later.

Requires backend: https://github.com/RedHatInsights/compliance-backend/pull/742

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [x] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -72,6 +72,11 @@ const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
         }));
     }
 
+    const setBenchmark = ({ id, osMajorVersion }) => {
+        change('benchmark', id);
+        change('osMajorVersion', osMajorVersion);
+    };
+
     return (
         <React.Fragment>
             <TextContent>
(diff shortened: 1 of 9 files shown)


src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -88,12 +94,12 @@ const CreateSCAPPolicy = ({ change, selectedBenchmarkId }) => {
...
                     { benchmarks && benchmarks.sort((a, b) => a.refId.localeCompare(b.refId)).map((benchmark) => {
-                        const { refId, id } = benchmark;
+                        const { id, osMajorVersion } = benchmark;
                         return (
-                            <Button key={id} onClick={ () => { change('benchmark', id); } }
+                            <Button key={id} onClick={ () => setBenchmark(benchmark) }
                                 className={`wizard-os-button ${selectedBenchmarkId === id ? 'active-wizard-os-button' : ''}`}
                                 variant="tertiary">
-                                { refId && refId.split('xccdf_org.ssgproject.content_benchmark_')[1].replace('-', ' ') }
+                                { `RHEL ${osMajorVersion}` }
```
bastilian, line 102: 🎉

akofink (approved):
Works as expected. Thanks, @vkrizan!

Probably not related to these changes, but some observations when testing:

The table shown sometimes seems to be showing cached results, perhaps cached in the store?
This table makes two identical requests to the compliance graphql endpoint.

vkrizan:
@akofink I've noticed the same when changing the benchmark. Could it be the debouncer? I'll try to investigate a bit.

vkrizan:
The two GraphQL queries are happening with the double refresh. They are triggered on some event or on redraw.

I can reproduce the cached values when the OS major is changed in a previous step. The component gets them populated on the second refresh (`state.rows`). It could be something between our InventoryTable component and the one from FEC. A quick patch would be to filter inventory loaded rows by our query result. But without knowing the whole picture it can be risky. Hopefully this would be improved with getEntities.

I must say that our Rudux stuff makes this clunky.

/cc @bastilian

PR #975, opened 2021-02-11
Update version of promise middleware.

Hyperkid123:
Fixes issues with incompatible versions of shared dependency. Chrome 2 shared v6+ but compliance uses v5. There was a breaking change when registering the middleware that can cause runtime crashes based on the order of loading the middleware.

Rather than excluding it from dependency sharing, I think it's ok to bump this dependency. It does not really affect anything.

cc @bastilian


akofink (approved):
Nothing appears to break. Thanks, @Hyperkid123!

PR #1003, opened 2021-02-17
feat(sonarqube): RHICOMPL-1260 add initial sonar config

akofink:

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,18 @@
+# must be unique in a given SonarQube instance
+sonar.projectKey=InsightsPlatform:compliance-frontend
+
+# --- optional properties ---
+
+# defaults to project key
+#sonar.projectName=My project
+# defaults to 'not provided'
+#sonar.projectVersion=1.0
+ 
+# Path is relative to the sonar-project.properties file. Defaults to .
+#sonar.sources=.
+ 
+# Encoding of the source code. Default is default system encoding
+#sonar.sourceEncoding=UTF-8
+
+sonar.javascript.lcov.reportPaths=coverage/lcov.info
+sonar.coverage.exclusions=config/*,src/__fixtures__/*,src/**/**/*.test.js,src/App*.js,*.config.js,src/bootstrap*.js,src/entry*.js,src/Routes.js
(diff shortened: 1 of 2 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,18 @@
...
+#sonar.projectVersion=1.0
+ 
+# Path is relative to the sonar-project.properties file. Defaults to .
+#sonar.sources=.
+ 
+# Encoding of the source code. Default is default system encoding
+#sonar.sourceEncoding=UTF-8
+
+sonar.javascript.lcov.reportPaths=coverage/lcov.info
+sonar.coverage.exclusions=config/*,src/__fixtures__/*,src/**/**/*.test.js,src/App*.js,*.config.js,src/bootstrap*.js,src/entry*.js,src/Routes.js
```
akofink, line 18: @bastilian anything here that shouldn't be? None of these were covered, so I suspected they're not expected to be tested.

Victoremepunto: We already have our test coverage reports in Codecov, do we still wanna do this, or we're going to migrate this stats to Sonarqube ? it doesn't makes much sense to have this stats reported in several reporting tools, however I don't have strong opinions about it

akofink: There is definitely duplication here, but sonarqube uses the output of our tests and coverage to report its own coverage - it doesn't run the tests itself or measure coverage itself.

bastilian: @akofink Is there maybe a way to configure it the other way around and configure what to include, not exclude, to match (the not up to date yet) patterns jest uses? https://github.com/RedHatInsights/compliance-frontend/blob/master/package.json#L51

vkrizan (approved):
Thank you!

sonar-project.properties
akofink: Unfortunately not. There is only an exclusion list. I'll update these exclusions to match the package.json.

akofink: Hmm.. that exclusions list does not seem to be quite correct. For example, do you want to track coverage on *.test.js files?

akofink: We can always update this list in the future if we have improvements.

PR #1005, opened 2021-02-18
feat: Use getEntities with InventoryTable

bastilian:
This introduces the use of `getEntities`[1] with the `InventroyTable` to reduce the amount of requests, only use data from our API, remove the need for a large reducer handling data consolidation and overall makes the component more stable.

[1] https://github.com/RedHatInsights/frontend-components/blob/53fa449f1c2b17c773d5a82b282aa4be156f64cb/packages/inventory/doc/custom_fetch.md

https://user-images.githubusercontent.com/7757/108351662-1b567600-71e6-11eb-99de-57e3f835a2b8.mp4

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -0,0 +1,76 @@
...
+    title: 'Last scanned',
+    column: 'scanned',
+    props: {
+        width: 10,
+        isStatic: true
+    },
+    renderFunc: renderComponent(LastScanned)
+}];
+
+export const selectedColumns = (selected = ['Name']) => (
```
bastilian, line 70: This needs to be able to take objects as columns in order to mix in props and allow to adjust the width for example.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -127,24 +113,30 @@ const InventoryTable = ({
...
                     INVENTORY_ACTION_TYPES,
                     mergeWithEntities
                 }) => {
                     getRegistry().register({
                         ...mergeWithEntities(
-                            systemsReducer(
-                                INVENTORY_ACTION_TYPES, columns, showAllSystems, policyId
-                            ))
+                            entitiesReducer(
+                                INVENTORY_ACTION_TYPES, tableColumns
```
bastilian, line 124: The `systemsReducer` didn't use the policyId argument, it didn't seem to matter and `showAllSystems` was used to differentiate the output when consolidating and preparing data.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -127,24 +113,30 @@ const InventoryTable = ({
...
+                            )
+                        )
                     });
                 }}
+                hideFilters={{
```
bastilian, line 129: These need to be enabled where we show all inventory hosts.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -154,10 +146,6 @@ const InventoryTable = ({
                     label: selectedEntities.length > 0 ? `${ selectedEntities.length } Selected` : undefined
                 }}
                 {...!showAllSystems && {
-                    ...pagination,
-                    isLoaded,
-                    items,
-                    total,
```
bastilian, line 160: These props are now handled via `getEntities`

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -1,481 +0,0 @@
-import React, { memo } from 'react';
```
bastilian, line 1: Bye, Bye.

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -12,9 +55,87 @@ const SSGVersion = ({ supported, ssgVersion }) => {
...
+
+FailedRules.propTypes = {
+    id: propTypes.string,
+    testResultProfiles: propTypes.array
+};
+
+const NEVER = 'Never';
+
+export const ComplianceScore = ({ testResultProfiles: profiles }) => {
+    const scoreTotal = profiles.reduce((acc, profile) => acc + profile.score, 0);
```
bastilian, line 92: All this calculation should be in the `ComplianceScore` component.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -82,38 +75,31 @@ const InventoryTable = ({
...
-        } else if (callback) {
-            callback(options);
-        }
+        return {
+            results: systems,
+            perPage,
+            // This is a workaround for a bug in InventoryTable(FEC)
+            per_page: perPage, // eslint-disable-line
+            page,
+            filters: activeFilters,
```
bastilian, line 92: Filters with `getEntities` are not yet properly functioning.

akofink:
When trying to remediate systems from the report details page:

```
Uncaught (in promise) TypeError: Systems array must not be empty
    H index.js:1
    K index.js:1
    xf index.js:1
    xf index.js:1
    e index.js:16
    p runtime.js:63
    _invoke runtime.js:293
    C runtime.js:118
    Babel 6
    openWizard index.js:14
    RemediationButton RemediationButton.js:1
    promise callback*RemediationButton/< RemediationButton.js:1
    React 14
    unstable_runWithPriority scheduler.development.js:646
    React 15
    js bootstrap-dev.js:5
    Webpack 8
index.js:1:3261
```

And the remediations request isn't sent.

akofink (commented):
Looking good! Thanks, @bastilian! I don't really see any weird behavior wrt double refresh or leaked state, but there are some new warnings in the console, like this one:

```
Warning: Failed prop type: Invalid prop `osMajorVersion` of type `number` supplied to `Name`, expected `string`.
Name@https://ci.foo.redhat.com:1337/apps/compliance/js/EditPolicy.js:1805:12
v@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.b5dbd7288d8f7e54b8b7.js:1:308664
div
...
```

src/SmartComponents/SystemsTable/Cells.js
bastilian: @akofink Totally agree!

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -82,38 +75,31 @@ const InventoryTable = ({
...
-    const onRefresh = (options, callback) => {
-        query && fetchSystems(options.per_page, options.page);
-        if (!callback && inventory && inventory.current) {
-            inventory.current.onRefreshData(options);
-        } else if (callback) {
-            callback(options);
-        }
+        return {
+            results: systems,
+            perPage,
```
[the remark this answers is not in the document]
bastilian: Yep, the fix for the issue in the Inventory Table should be released soon, then we should be able to remove both.

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -101,67 +88,8 @@ export const ReportDetails = ({ route }) => {
...
-            width: 10
-        },
-        ...newInventory && {
-            key: 'lastScanned',
-            renderFunc: (lastScanned) => (lastScanned instanceof Date) ?
-                <DateFormat date={Date.parse(lastScanned)} type='relative' />
-                : lastScanned
-        }
-    }];
+    const columns = ['Name', 'ssg', 'Failed rules', 'score', 'Last scanned'];
```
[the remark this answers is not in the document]
bastilian: Yeah, I wasn't really fond of it either. :D I tried this with the RulesTable, but never was too happy with it and I am really not sure why I didn't think of implementing it the way you suggest. It is now jus importing the objects and custom props can be set by just extending the object.

vkrizan:
Hi @bastilian, I've briefly went through out the changes and identified concerns that can be separated:
* Columns refactoring (great one!)
* code split to useFetchSystem hook
* introduction of getEntities
* selection fixes (reduction of state changes, new states, ...)
* Export feature

`useFetchSystem` might be together with `getEntities` (if it cannot be separated).

Please take the above as an example or as an opportunity for exercise.

bastilian:
@vkrizan retroactively splitting up code into commits is not something that can be done easily in this case and the benefits outweigh the cost, in my opinion.

Already the columns refactoring presents an issue that would make it hard almost impossible as it would require, changing and adapting or removing the entities reducer before hand. The entities reducer can't be removed before the getEntities is added. These are just two examples that would require much more work to separate these out. 

And in the end we are squashing them anyways.

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -29,39 +28,8 @@ const DEFAULT_FILTER = 'has_test_results = true or has_policy = true';
...
-            width: 20, isStatic: true
-        },
-        renderFunc: (data, id) => {
-            const { title } = detailsLink({ testResultProfiles: data, id }) || { title: '' };
-            return title;
-        }
-    }];
     const policies = data?.profiles?.edges.map(({ node }) => node);
 
-    useLayoutEffect(() => { dispatch({ type: 'SELECT_ENTITIES', payload: { ids: [] } }); }, []);
```
bastilian, line 63: These are not needed anymore. The `InventoryTable` will dispatch a 'INVENTORY_INIT' which will reset the table to a loading state without selection.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -42,92 +34,53 @@ export const InventoryTable = ({
...
-        debounce(fetchSystems, 800),
-        [conditionalFilter.activeFiltersConfig.filters]
-    );
+    }, []);
 
     useEffect(() => {
-        if (conditionalFilter.activeFiltersConfig.filters) {
-            debounceFetchSystems(pagination.perPage, 1);
-        }
+        inventory.current?.onRefreshData();
```
bastilian, line 81: This will trigger the inventory to call `getEntities` again.

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: ...I should have added eventually. These are the default inventory filters, once our API supports these we can enable them.

src/store/Reducers/SystemStore.js
```diff
@@ -214,7 +58,25 @@ export const systemsReducer = (INVENTORY_ACTION, columns) => applyReducerHash({
...
-    ['SELECT_ENTITIES']: (state, { payload: { ids } }) => ({
-        selectedEntities: ids
+    ['INVENTORY_INIT']: () => ({
+        rows: [],
+        total: 0,
+        columns
     })
 });
+
+export const mapCountOsMinorVersions = (systems) => {
```
bastilian, line 68: Just moved from that chunk above to down here.

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: This doesn't appear to be an issue. The call to getEntities from teh inventory passes `page` and `perPage`, which we can use with our API.

bastilian:
> Export feature

@vkrizan now in https://github.com/RedHatInsights/compliance-frontend/pull/1161. This wasn't supposed to be here, just something i had in my stash.

src/store/Reducers/SystemStore.js
```diff
@@ -214,7 +58,25 @@ export const systemsReducer = (INVENTORY_ACTION, columns) => applyReducerHash({
 
         return newState;
     },
-    ['SELECT_ENTITIES']: (state, { payload: { ids } }) => ({
-        selectedEntities: ids
+    ['INVENTORY_INIT']: () => ({
+        rows: [],
+        total: 0,
+        columns
     })
```
akofink, line 65: This isn't an inventory table action. Perhaps we should add it there instead of here?

src/store/Reducers/SystemStore.js
```diff
@@ -161,43 +34,14 @@ const deselectRow = (state, id) => (
...
         ...state,
-        total: state.systemsCount,
-        rows: state.systems !== undefined ? systemsToRows(state.systems).map((row) => ({
-            ...row, selected: isSelected(row.id, state.selectedEntities)
-        })) : [],
-        columns,
-        loaded: state.systemsCount !== undefined
+        columns: state.total > 0 ? columns : [{ title: '' }]
     }),
     [SELECT_ENTITY]: (state, { payload: { id, selected, clearAll } }) => {
```
akofink, line 46: We should be able to remove this from our constants and use:

```suggestion
    [INVENTORY_ACTION.SELECT_ENTITY]: (state, { payload: { id, selected, clearAll } }) => {
```

src/store/Reducers/SystemStore.js
bastilian: Yeah, i was thinking the same. I'll look into that!

src/store/Reducers/SystemStore.js
akofink: Why do we even need to keep this reducer? If there's extra functionality that would be useful for the inventory table to provide, we should add it there. I agree `SELECT_ENTITY` is a bit cumbersome to use without a select all/none, so let's contribute this back.

bastilian: Unfortunately 'SELECT_ENTITY' isn't exported[0] as far as I can see. :/

> Why do we even need to keep this reducer?

We don't. I was close to ripping it out. (I had a facepalm moment this morning and @vkrizan too) We don't need to interact or use the inventory store at all to keep selections, even now (in master) we could easily put this all into a simple `useState`. The more complex solution would be `useBulkSelect`[1]. I'm not sure anymore, I believe we started doing this with the `selectedEntities` because there was no other way to mark rows as selected (maybe). But we should definitely get rid of this. 

https://github.com/RedHatInsights/frontend-components/blob/39ab58e14990887dff0f2c42ef0c5a20f0f3e11b/packages/inventory/src/redux/action-types.js#L29

vkrizan: That's interesting about the event not being exposed... It is documented here https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory/doc/inventory.md#inventory-actions. How would we get the checkbox event out of the Inventory w/o listening on that event (state change)?

akofink: I really like `useSelector`. It's really clean and doesn't clutter our code with redux stuff.

akofink: If SELECT_ENTITY isn't exported but it'd be useful to be exported, then let's go export it! :wink: Unless the platform UI folks have a reason against it, we can claim we need some functionality not provided. Maybe they'd have an even more elegant solution for us.

akofink: Wait... it is exported, just not in INVENTORY_ACTIONS: https://github.com/RedHatInsights/frontend-components/blob/39ab58e14990887dff0f2c42ef0c5a20f0f3e11b/packages/inventory/src/redux/action-types.js#L35

bastilian: @vkrizan That is interesting! Either the docs are not up to date or the code. :D

> How would we get the checkbox event out of the Inventory w/o listening on that event (state change)?

We should be able to use the plain Patternfly table `onSelect` prop. There is might even an FECInventoryTable specific prop.

bastilian: @akofink We also would need to add the whole inventory package just to get the SELECT_ENTITY constant, i think. 

Again, we can and probably eventually will remove that part. It works for now.

bastilian:
@vkrizan @akofink @skateman I went over it one last time yesterday and couldn't find any odd bits or issues. Please, feel free to  review and test this. All should appear the same as is in `master`.

akofink:
Wow, the filters are so much faster! :clap: So fast, in fact, I'm seeing some kind of race:

image

Typing '97' quickly into the name filter sends 2 queries, in this order (as expected):

`(os_major_version = 7)and (name ~ 9))` 67ms response, 67ms end time
`(os_major_version = 7)and (name ~ 97))` 50ms response, 131ms end time (I'm assuming this means the responses come back in order)

But the table only renders the result of the first query and does not redraw for the result of the second. We could solve this probably with some kind of debounce delay, but the filters seem so much faster now - it'd be awesome if they could stay super fast and just figure out the rendering in order of query.

bastilian:
@akofink Nice! The issue with the filters could be that this is caused by the way updates are manually triggered. I'll be looking into making sure that this moves to the FECInventoryTable component, which should ensure it queries and returns in order and possibly debounce request. Then they'll be fast and reliable!

akofink:
I also noticed that the selected systems do not stay selected in the Create Policy Wizard when going back from step 4 (rules) to step 3 (systems).

akofink:
I see the selections not persisted between filtering as well:

image

I selected a system, then applied the OS filter, then cleared filters.

bastilian:
@akofink Nice catch! Already have a fix, just need to test it.

bastilian:
@akofink Thank you for reviewing and testing! The issues with the selection being reset in the Wizard should now be fixed.

bastilian:
Oh, no wait. there is still the issue with the filter. Looking into that.

akofink:
In the Edit Policy Modal, with a policy that doesn't have hosts, adding hosts does not enable the save button:

image

vkrizan:
> In the Edit Policy Modal, with a policy that doesn't have hosts, adding hosts does not enable the save button:

@akofink this is being fixed via #1177 or #1178

bastilian:
@akofink The issues with filtering (quickly) and it showing unexpected results and with filtering and loosing the selection should be fixed. It is now debouncing the the call to update the inventory table, which mitigates the incorrect results. The manual call to trigger an update is a temporary measure until the inventory supports calling `getEntities` for custom filters. This as well as the initialisation call and updating the row selection are now contained within one hook and will hopefully one day be gone. Till that is the case, I have added some comments to elaborate why these are there. There are also now a handful more tests for the InventoryTable.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -137,55 +87,47 @@ export const InventoryTable = ({
             { emptyStateComponent }
         </StateViewPart>
         <StateViewPart stateKey='noError'>
-            { !!prependComponent && isLoaded && prependComponent }
+            { !!prependComponent && prependComponent }
```
vkrizan, line 90: Regression: This lacks the `isLoaded` mechanism. Is there any replacement of `isLoaded` that would keep the behavior?

An example can be seen in Create Policy when selecting OS version that do not have any systems. Until the table is loaded the text shows to "select systems" and then it slips tho "no systems".

vkrizan (commented):
Great improvement! It much improves the handling of the Inventory table and fixes a bunch of bugs.

There are some small user experience issues that might be looked at.
Thanks.

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: This should be fixed now I re-added the behaviour, just turned it around to be `isLoaded`

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -12,9 +56,85 @@ const SSGVersion = ({ supported, ssgVersion }) => {
 
 SSGVersion.propTypes = {
     supported: propTypes.bool,
-    ssgVersion: propTypes.string
+    ssgVersion: propTypes.string,
+    testResultProfiles: propTypes.array
+};
+
+export const DetailsLink = ({ id, testResultProfiles = [] }) => (
+    testResultProfiles.length > 0 && <SystemLink { ...{ id } }>
```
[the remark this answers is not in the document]
bastilian: Added a fallback for this.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
```diff
@@ -52,35 +53,23 @@ const EditPolicySystemsTab = ({ osMajorVersion, policyOsMinorVersions }) => {
...
-        title: 'Operating system',
-        props: {
-            width: 40, isStatic: true
-        },
-        renderFunc: (osMinorVersion, _id, { osMajorVersion }) => `RHEL ${osMajorVersion}.${osMinorVersion}`
-    }];
-
     return (
         <React.Fragment>
             <InventoryTable
```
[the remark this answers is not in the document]
bastilian: Yes, this can and will be improved. The inventory does actually, briefly show a spinner, which is part of the infamous "double refresh" issue. We previously provided a skeleton table as a fallback. This would make the loading of the inventory table transition from fallback table (via async loader) -> spinner from inventory -> inventory table loading and make it seem the table is being loaded twice. With these changes the loading behaviour is already slightly better, but the rest of the improvements should be done in the FEC InventoryTable.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -137,55 +87,47 @@ export const InventoryTable = ({
...
-                    total,
                     ...conditionalFilter,
                     ...remediationsEnabled && {
                         dedicatedAction: <ComplianceRemediationButton
-                            allSystems={ systemsWithRuleObjectsFailed(selectedEntities) }
+                            allSystems={ systemsWithRuleObjectsFailed(selectedItems) }
                             selectedRules={ [] } />
                     }
                 }}
                 {...enableExport && {
```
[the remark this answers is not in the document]
bastilian: This should now be working again as expected.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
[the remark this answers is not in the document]
vkrizan: Can there be something temporary to mitigate this until we have a proper fix within the FEC?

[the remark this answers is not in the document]
bastilian: @vkrizan Not really. If we'd add a fall back again for loading, it would return to the "double refresh" issue, like it is now in prod.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -29,3 +33,138 @@ export const useOsMinorVersionFilter = (showFilter) => {
...
+            query,
+            fetchResults: true,
+            fetchPolicy: 'no-cache',
+            variables: {
+                filter,
+                perPage,
+                page,
+                ...policyId && { policyId }
+            }
+        }).then(({ data }) => {
```
vkrizan, line 61: Observation: There is no error processing for the query (e.g. on a Network error or 500).

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -44,89 +35,50 @@ export const InventoryTable = ({
...
-            dispatch({
-                type: 'GET_SYSTEMS_FULFILLED',
-                systems: data.systems.edges,
-                systemsCount: data.systems.totalCount
-            });
+        if (emptyStateComponent &&
+            result.meta.totalCount === 0 &&
+            activeFilterValues.length === 0) {
+            setIsEmpty(true);
             setIsLoaded(true);
```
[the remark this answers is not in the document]
bastilian: Moved it up so it is now set on any onComplete from the fetch.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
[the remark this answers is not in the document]
vkrizan: There is a `fallback` prop defined set to empty string. Can we add a `Spinner` there? It seems to work almost seamlessly w/o any double refresh. I only experience couple of re-renders that reset the spinner animation and some slight delay between a spinner and the table. The latter could be due to the fact that it loads some new components, which might not be visible in production.

src/Utilities/Export.js
```diff
@@ -1,3 +1,5 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
bastilian: Yeah, an ongoing battle. :D Some lodash functions aren't as easily replaced with something slimmer. I could have extended the crude deep-fetch, but it might have cost more headache. We can revisit this when we have to implement the export on all other tables.

src/SmartComponents/SystemsTable/hooks.js
bastilian: That is a nice find! I checked in master and it is missing there as well. I'll open a Jira to follow up on that, if there is none already. It behaves the same here and in master and gets stuck in the loading phase.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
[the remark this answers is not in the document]
bastilian: Added a Spinner as the fallback, now it is Spinner -> Spinner -> Loading table. 

The re-rendering might still be some oddities going on with regards to chrome. I see all redux logs twice (in CI) at the moment, for example.

vkrizan (approved):
Works like a charm.
Thank you for this huge effort!

akofink:
Thanks, @bastilian! It looks great. I'm doing some more in-depth testing of this today and noticed a regression. 
When filtering by OS in the wizard and receiving no results back, we get the incorrect empty state:

image

It seems to work as expected on CI.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
```diff
@@ -96,6 +85,7 @@ const EditPolicySystemsTab = ({ osMajorVersion, policyOsMinorVersions }) => {
 };
 
 EditPolicySystemsTab.propTypes = {
+    policy: propTypes.object,
     osMajorVersion: propTypes.string,
```
akofink, line 89: Hmm... isn't this just `policy.osMajorVersion`? Seems like we could remove the prop.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -29,3 +33,138 @@ export const useOsMinorVersionFilter = (showFilter) => {
...
+    const resetPage = () => {
+        Promise.resolve(dispatch({
+            type: 'RESET_PAGE'
+        })).then(() =>
+            inventory?.current?.onRefreshData()
+        );
+    };
+
+    // The debounce is to not have filter updates collide or get out of order.
+    const debounceResetPage = debounce(resetPage, 200);
```
akofink, line 165: I've got to say, it felt better without this debounce :disappointed: 200ms is noticeable. Does it work with something lower, say 50 or 80ms? That should keep it below the perception threshold of users.

src/Utilities/Export.js
```diff
@@ -12,21 +14,11 @@ const linkAndDownload = (data, filename) => {
...
-
-    if (cell.exportValue) {
-        cell = cell.exportValue;
-    }
-
-    if (typeof(cell) === 'object') {
-        cell = getNestedObject(row, key + '_text');
+const textCsvCell = (row, column) => {
+    const { exportKey: key, renderExport } = column;
+    let cell = key ? _.at(row, key)[0] : row;
```
akofink, line 19: This seems like it could become problematic if the API requests don't include the correct attributes. It's also quite a bit more confusing to understand what's happening here (lodash `at` doesn't help). What was the motivation in switching from `exportValue` to `exportKey`?

src/store/Reducers/SystemStore.js
```diff
@@ -1,102 +1,39 @@
...
+    }),
+    ['RESET_PAGE']: (state) => ({
+        ...state,
+        page: 1
+    }),
+    ['SELECT_ENTITIES']: (state, { payload: { selected } }) => ({
+        ...state,
+        selectedEntities: selected,
+        rows: selectRows(state.rows, selected)
+    })
```
akofink, line 34: Still curious why we can't remove these non-inventory actions or move their (clearly useful) functionality to the FECInventoryTable

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -0,0 +1,112 @@
...
+    ...Name,
+    props: {
+        ...Name.props,
+        ...props
+    },
+    renderFunc: renderComponent(NameCell, props)
+});
+
+export const SsgVersion = {
+    exportKey: 'testResultProfiles[0].ssgVersion',
```
akofink, line 37: This is not quite right (not deterministic). It should use the latest testResultProfile, not just the first.

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -0,0 +1,112 @@
...
+    props: {
+        width: 20,
+        ...disableSorting
+    },
+    renderFunc: renderComponent(DetailsLinkCell)
+};
+
+export const FailedRules = {
+    title: 'Failed rules',
+    exportKey: 'testResultProfiles[0].rules.length',
```
akofink, line 65: This is not deterministic, and it returns total rules (not just failed), and it's also exposed on the API:

```
    {         
      "id": "6d9d25aa-c495-45ba-9bfe-e9ddbe5a8e63",
      "type": "host",
      "attributes": {   
        "name": "rhel79-insights-client.virbr0.akofink-laptop",
        "os_major_version": 7,
        "os_minor_version": 9,
        "last_scanned": "2021-04-27T13:09:32Z",
        "rules_passed": 40,
        "rules_failed": 28,
        "has_policy": true,
        "compliant": true
      },
      "relationships": {…}
    }
```

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -0,0 +1,112 @@
...
+    props: {
+        width: 5,
+        ...disableSorting
+    },
+    renderFunc: renderComponent(FailedRulesCell)
+};
+
+export const ComplianceScore = {
+    title: 'Compliance score',
+    exportKey: 'testResultProfiles',
```
akofink, line 75: This doesn't seem right :thinking: It's not used if renderExport is defined, right?

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -0,0 +1,112 @@
...
+    },
+    renderExport: (cell) => (
+        complianceScoreString(complianceScoreData(cell))
+    ),
+    renderFunc: renderComponent(ComplianceScoreCell)
+};
+
+export const LastScanned = {
+    title: 'Last scanned',
+    exportKey: 'testResultProfiles',
```
akofink, line 88: Same here. This isn't used if `renderExport` is defined?

akofink (commented):
It's nearly there! Due to the size of the PR, it seems almost impossible to merge this without introducing regressions, but I guess that's what we'll have to do. Still, we can fix those that we know about before merging. In the future, please (please, I'm begging!) break up your PRs so reviewing is easier, quicker, and less painful. I guarantee you'll get less back-and-forth, and we'll be able to be more confident that we're not introducing regressions. I guess I'll be re-re-re-reviewing this tomorrow :crying_cat_face:

sthirugn:
+1 to @akofink suggestion here, atomic PRs will be better manageable in reviews/getting merged quickly.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
bastilian: It is use. It is for passing in the preselected hosts.

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -12,9 +56,87 @@ const SSGVersion = ({ supported, ssgVersion }) => {
...
+    testResultProfiles: propTypes.array
+};
+
+export const lastScanned = (profiles) => {
+    const dates = profiles.map((profile) => new Date(profile.lastScanned));
+    const last = new Date(Math.max.apply(null, dates.filter((date) => isFinite(date))));
+    const result = (last instanceof Date && isFinite(last)) ? last : NEVER;
+
+    return result;
+};
```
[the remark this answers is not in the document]
bastilian: I opened a Jira to look into these places and replace the calculations with the API attributes.

src/SmartComponents/SystemsTable/Columns.js
bastilian: There are two things related to the export `renderExport` a function to build a string for the export and `exportKey`. The latter allows to define a property to use in `renderExport`, by default it will get the whole object for a system. The reason for using `exportKey` is to not interfere with the FECInventoryTable component, which use that to do something similar for the `renderFunc`. 

The export has become a bit of an eyesore, we can look into improving it more when we look into adding export to all tables in compliance, which was pointed out in the last UI review. 

For now this should work and exports data correctly at least on the systems table.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -1,35 +1,26 @@
...
 import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import useFilterConfig from 'Utilities/hooks/useFilterConfig';
 import { InventoryTable as FECInventoryTable } from '@redhat-cloud-services/frontend-components/Inventory';
-import { useOsMinorVersionFilter } from './hooks';
+import { policyFilter, defaultOnLoad } from './constants';
+import {
+    useFetchSystems, useGetEntities, useOsMinorVersionFilter, useInventoryUtilities, useOnSelect
+} from './hooks';
 
 export const InventoryTable = ({
```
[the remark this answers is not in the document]
bastilian: @akofink Yes! We totally can and we will, but let's do it in a separate PR with the rename and changing the component name throughout compliance back to `SystemsTable`.

src/store/Reducers/SystemStore.js
bastilian: This is part of a "compatibility layer". There are still parts relying on the `selectedEntities` in the inventory store. This adds a `preselectedSystems` and `onSystemSelect` prop, which we can use in those places in the future.

Another reason for it is that some interactions do not properly trigger an update in the Inventory this is to remedy that. Eventually this and the `useInventoryUtilities` hook should not be necessary, maybe even all of the `entitiesReducer`.

src/SmartComponents/SystemsTable/hooks.js
bastilian: It is now down to 50ms, it's less noticeable now, we could try go lower, but the sweet-spot may also vary.

src/SmartComponents/SystemsTable/Columns.js
bastilian: This should now be better, it's doing the same as the table cell and joins multiple, this is also how we do it in some places now, in others the column renderFunction only uses the first profile. This corrects it to always render all in profiles' SSGVersion. We should look into each systems query and make sure the the profile(s) requested is/are the latest and only one, though it can stay an array, maybe.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
akofink: Sorry, what? It's only ever passed in from EditPolicyForm, and it's `policy.osMajorVersion`. The prop is redundant with `policy`.

bastilian: Oh, Sorry, my bad. I though you were talking about addition of `policy`, I wasn't quite sure which prop you were refering to. Yeah, i'll change that!

src/SmartComponents/SystemsTable/Cells.js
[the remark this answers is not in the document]
akofink: RHICOMPL-1760 for reference

bastilian:
@akofink @vkrizan Thank you both for taking the time and reviewing! I'm optimistic (and biased) that if it does introduce regressions, they'll be easier to address now with a better base that does not interfere and rely (much) on the inventory store.

I'm (genuinely) sorry, this is a larger than usual PR. If there ever is another change as impactful as this, and I'm the author, and it becomes difficult to maintain a history of the steps toward the final code, I will at least try to separate them by something, which in other instances I have do, but became impossible to main with these change. But I'm hopeful that it won't require changes like this in the future.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
bastilian: Changed now it is only passed a `policy` prop.

akofink (approved):
Thanks, @bastilian for your huge effort here! :1st_place_medal:  To recap: the plan is to merge this PR soon after our next deployment to production to give QEs (and devs) as much time as possible to work with this before the Next + 1 deployment.

PR #1007, opened 2021-02-18
fix(InventoryTable): RHICOMPL-1442 redux states

akofink:
This change ensures the redux state stays set as intended. This includes
the rendered columns and rows as well as whether the table is loaded or
not.

It is **very important** to realize that the redux actions in our
SystemStore have (basically) no guarantee of any execution order. One
exception is the *_PENDING and *_FULFILLED actions, which always happen in
that order but still may have other actions happen in between.

This adds a GET_SYSTEMS_PENDING action to allow compliance to reset
columns and rows in addition to system information when our API
request begins. This is necessary to ensure old state is not persisted
into a new table, such as when viewing policy systems and then
immediately editing policy systems without refreshing.

Additionally, there was a bug in EditPolicy, which filtered the system list by
policy ID, even though we want to show all systems there.

Hopefully, this will be a good stop-gap measure until the proper InventoryTable
with getEntities is working. Please test filtering, selection, export, etc. and report
any and all weird behavior you see. Note: export seems to only work from
the systems page on master and in this PR.

Also fixes RHICOMPL-1445 in the second commit.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -98,11 +98,9 @@ export const EditPolicy = ({ route }) => {
                         compact
                         showActions={ false }
                         enableExport={ false }
-                        showAllSystems
                         remediationsEnabled={ false }
                         policyId={ policy.id }
                         query={GET_SYSTEMS_WITHOUT_FAILED_RULES}
-                        defaultFilter={`policy_id = ${policy.id}`}
                         columns={[{
                             key: 'facts.compliance.display_name',
                             title: 'Name',
--- a/src/store/Reducers/SystemStore.js
+++ b/src/store/Reducers/SystemStore.js
@@ -19,7 +19,6 @@ import {
     profilesRulesFailed
 } from 'Utilities/ruleHelpers';
 import Truncate from 'react-truncate';
-import { mergeArraysByKey } from '@redhat-cloud-services/frontend-components-utilities/helpers';
 
 const NEVER = 'Never';
 
@@ -196,6 +195,22 @@ export const systemsToInventoryEntities = (systems, entities, showAllSystems, se
     }).filter((value) => (!!value))
 );
 
+const systemsToRows = (systems) => (
+    systems.map(({ node }) => ({
+        ...node,
+        policyNames: policyNames({ policies: node?.policies, testResultProfiles: [] }),
+        rulesPassed: profilesRulesPassed(node.testResultProfiles).length,
+        rulesFailed: profilesRulesFailed(node.testResultProfiles).length,
+        lastScanned: lastScanned(node),
+        compliant: compliant(node),
+        display_name: node.name, // eslint-disable-line camelcase
+        score: score(node),
+        supported: supported(node),
+        ssgVersion: profilesSsgVersions(node),
+        detailsLink: detailsLink(node)
+    }))
+);
+
 const selectRowsByIds = (state, ids) => {
     const rowsToSelect = state.rows.filter((row) => (
         ids.includes(row.id) && !(state.selectedEntities || []).map((e) => (e.id)).includes(row.id)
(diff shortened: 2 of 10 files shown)


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -102,7 +102,6 @@ export const EditPolicy = ({ route }) => {
                         remediationsEnabled={ false }
                         policyId={ policy.id }
                         query={GET_SYSTEMS_WITHOUT_FAILED_RULES}
-                        defaultFilter={`policy_id = ${policy.id}`}
```
akofink, line 105: This is a bug that never manifested itself, since we were still merging results from the inventory API. This PR no longer uses the data returned from the inventory API at all.

src/store/Reducers/SystemStore.js
```diff
@@ -196,6 +195,22 @@ export const systemsToInventoryEntities = (systems, entities, showAllSystems, se
...
 
+const systemsToRows = (systems) => (
+    systems.map(({ node }) => ({
+        ...node,
+        policyNames: policyNames({ policies: node?.policies, testResultProfiles: [] }),
+        rulesPassed: profilesRulesPassed(node.testResultProfiles).length,
+        rulesFailed: profilesRulesFailed(node.testResultProfiles).length,
+        lastScanned: lastScanned(node),
+        compliant: compliant(node),
+        display_name: node.name, // eslint-disable-line camelcase
```
akofink, line 206: This function was copied nearly exactly, except this line, which ensures `display_name` is a part of our node, even if it wasn't requested in the API. If you've ever seen the inventory table render a row without a name, this is why. We could ensure our `columns` always request `name`, but this was a less invasive and more robust change IMO.

src/store/Reducers/SystemStore.js
```diff
@@ -228,35 +243,38 @@ const deselectRow = (state, id) => (
...
-    ['UPDATE_SYSTEMS']: (state, { systems, systemsCount }) => ({
+export const systemsReducer = (INVENTORY_ACTION, columns) => applyReducerHash({
+    ['GET_SYSTEMS_PENDING']: (state) => ({
+        ...state,
+        rows: [],
+        systems: undefined,
+        systemsCount: undefined,
+        columns,
+        loaded: false
+    }),
```
akofink, line 254: When thinking about all these actions, please remember that the order is not guaranteed. When the compliance API request begins, we always need to reset the state with empty rows, undefined systems, etc. or risk some other action thinking that rendering is complete (i.e. double refresh, or incorrect data shown). Also, state can leak from other tables and must be cleared, such as when displaying a policy's systems, then creating a new policy without refreshing.

src/store/Reducers/SystemStore.js
```diff
@@ -228,35 +243,38 @@ const deselectRow = (state, id) => (
...
-                policyNames: policyNames({ policies: node?.policies, testResultProfiles: [] }),
-                rulesPassed: profilesRulesPassed(node.testResultProfiles).length,
-                rulesFailed: profilesRulesFailed(node.testResultProfiles).length,
-                lastScanned: lastScanned(node),
-                compliant: compliant(node),
-                score: score(node),
-                supported: supported(node),
-                ssgVersion: profilesSsgVersions(node),
-                detailsLink: detailsLink(node)
-            }))
```
akofink, line 257: Moved to `systemsToRows`

src/store/Reducers/SystemStore.js
```diff
@@ -228,35 +243,38 @@ const deselectRow = (state, id) => (
...
-        ]),
-        columns: state.total > 0 ? columns : [{ title: '' }]
+        total: state.systemsCount,
+        rows: state.systems !== undefined ? systemsToRows(state.systems).map((row) => ({
+            ...row, selected: isSelected(row.id, state.selectedEntities)
+        })) : [],
```
akofink, line 275: Inventory will try to set `rows` here every time, and we never want their rows. If LOAD_ENTITIES_FULFILLED happens before _or_ after GET_SYSTEMS_FULFILLED, we want the rows from the compliance API. That's why this must match how we set rows in GET_SYSTEMS_FULFILLED.

bastilian:
Awesome! Thank you @akofink for looking into this. It looks good to me as a temporary measure. It does double the amount of requests to GraphQL which could cause issues in certain cases, like filtering in combination with a slow connection. 

There is also an issue with the selecting, which causes the table to indefinitely rerender:

 

https://user-images.githubusercontent.com/7757/108434254-a3b93300-7247-11eb-9135-427c323a1e56.mp4

vkrizan:
@akofink respect :+1:

akofink:
Thank you for the reviews! The select issue (RHICOMPL-1445) and the weird indefinite render issues should be resolved as well. Please re-test. Also, please test remediations work as expected from the report details page and that only supported systems are sent to the remediations API (you can see the request payload in the network tab).

src/store/Reducers/SystemStore.js
vkrizan: The Name filter does not seems to work properly on Edit Policy modal (not sure if this is the cause). It works on the Wizard (Create Policy) and on Systems list page.

(ignore the first host)
Screenshot_2021-02-22 My HIPAA - SCAP policies - Compliance Red Hat Insights(2)

The component sends empty `""` filter to GraphQL, but requests hosts by name through Inventory hosts API
```
https://ci.foo.redhat.com:1337/api/inventory/v1/hosts?hostname_or_id=somename&per_page=50....
                                                      ^^^^^^^^^^^^^^^^^^^^^^^
```

vkrizan (approved):
Works correctly on almost all pages. One bug is within Edit Policy when filtering by Name. See comment...

src/store/Reducers/SystemStore.js
akofink: Ok! This should be fixed now. There's still some logic using `showAllSystems`  that affects the filter for some reason in the InventoryTable that we need to remove completely or replace with a different prop.

PR #1023, opened 2021-03-01
chore(container): Correct webpack binding in container

bastilian:
This corrects the binding behaviour of webpack dev server when it is running within a container.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -6,4 +6,6 @@ if [[ ! -d "/compliance/node_modules/@redhat-cloud-services" ]]; then
     cd /compliance; /usr/bin/npm install;
 fi
 
+export IN_DOCKER='true';
+
 exec "$@"
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
 # Comment out to enable
 DEFAULT_HOST=ci.foo.redhat.com
 HOT_RELOAD='false'
+# WEBPACK_BIND=192.168.0.2
 
 # Enable/uncomment this to run chrome locally
 # CHROME_DIR=../insights-chrome
--- a/config/devserver.config.js
+++ b/config/devserver.config.js
@@ -2,6 +2,14 @@
 /* global module */
 let default_host;
 
+const bindHost = () => {
+    if (process.env.IN_DOCKER && typeof(process.env.WEBPACK_BIND) === 'undefined') {
+        return '0.0.0.0';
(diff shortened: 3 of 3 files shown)


akofink (commented):
As discussed in Slack: this prevents the proxy from being able to access a local API (i.e. localhost:3000 for compliance-backend); however, the dns issue seems to be resolved.

skateman:
I investigated further the issue with the DNS servers and if the host OS uses systemd-resolved, podman copies `/run/systemd/resolve/resolv.conf` for host networked containers, which in my case somehow only contains the DNS servers provided by my ISP. As this file is auto-generated, I tried to mess with the DNS priorities in my `NetworkManager` but that didn't really help me. 

If we're not using host networking for the proxy, the API is basically inaccessible as it is running in a separate pod and podman doesn't support [`host.docker.internal`](https://github.com/RedHatInsights/compliance-frontend/blob/master/config/spandx.config.js#L11) or anything similar like `host.podman.internal` and according to what I found, they are suggesting people to just use `--net=host` :confused:

akofink:
I think we can close this one. As @skateman says, we were hitting issues with new Fedora 33 'features' :laughing:

skateman:
I'm still struggling with the frontend not listening on `0.0.0.0` so I have to keep stashing stuff when git-pulling.

akofink:
Ok, sure. @bastilian could we repurpose this PR to fix just the bind?

bastilian:
@skateman The binding should now work out of the box.

skateman (approved):
The Seal of Approval

entrypoint.sh
```diff
@@ -6,4 +6,6 @@ if [[ ! -d "/compliance/node_modules/@redhat-cloud-services" ]]; then
     cd /compliance; /usr/bin/npm install;
 fi
 
+export IN_DOCKER='true';
```
bastilian, line 9: For some reason the Docker compose documentation is not right. A environment variable defined in the .env file doesn't seem to overwrite the one in the compose file[1].

[1] https://docs.docker.com/compose/environment-variables/

PR #1044, opened 2021-03-12
refactor: RHICOMPL-1355 presentational TabbedRules from PolicyMultiversionRules

vkrizan:
Created presentational component `TabbedRules` out of `PolicyMultiversionRules`

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/PresentationalComponents/TabbedRules/TabbedRules.js
+++ b/src/PresentationalComponents/TabbedRules/TabbedRules.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Tab } from '@patternfly/react-core';
+import SystemRulesTable, {
+    selectColumns as selectRulesTableColumns
+} from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
+import { RoutedTabs } from 'PresentationalComponents';
+
+const eventKey = (id) => (
+    `rules-${id}`
+);
+
+const defaultTab = (tabsData, profileId) => {
+    if (tabsData && tabsData.length > 0) {
+        return eventKey(profileId || tabsData[0].profile.id);
+    }
+};
+
+const TabbedRules = ({ tabsData, defaultProfileId, columns, level, handleSelect, ...rulesTableProps }) => (
+    <RoutedTabs level={ level } defaultTab={ defaultTab(tabsData, defaultProfileId) }>
+        {
+            tabsData?.map(({ title, profile, rules }) => (
+                <Tab
+                    title={ title }
+                    key={ eventKey(profile.id) }
+                    eventKey={ eventKey(profile.id) }>
+                    <SystemRulesTable
+                        remediationsEnabled={false}
+                        columns={ selectRulesTableColumns(columns) }
+                        profileRules={ [{ profile, rules }] }
+                        handleSelect={
+                            handleSelect
+                            && ((selectedRuleRefIds) => handleSelect(profile, selectedRuleRefIds))
+                        }
+                        { ...rulesTableProps } />
+                </Tab>
+            ))
+        }
+    </RoutedTabs>
+);
+
+TabbedRules.propTypes = {
+    tabsData: propTypes.arrayOf(
+        propTypes.shape(
+            {
+                title: propTypes.oneOfType([propTypes.string, propTypes.node]).isRequired,
+                profile: propTypes.object.isRequired,
+                rules: propTypes.arrayOf(propTypes.object).isRequired
+            }
+        )
+    ).isRequired,
+    defaultProfileId: propTypes.string,
+    columns: propTypes.arrayOf(propTypes.string),
+    level: propTypes.number,
+    handleSelect: propTypes.func
+};
+
+TabbedRules.defaultProps = {
+    columns: ['Name', 'Severity', 'Ansible'],
+    level: 0
+};
+
+export default TabbedRules;
--- a/src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
+++ b/src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
@@ -1,38 +1,37 @@
 import React from 'react';
 import propTypes from 'prop-types';
-import { Alert, Label, Tab, TabTitleText, PageSection, PageSectionVariants } from '@patternfly/react-core';
-import SystemRulesTable, {
-    selectColumns as selectRulesTableColumns
-} from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
-import { RoutedTabs } from 'PresentationalComponents';
+import { Alert, Label, TabTitleText, PageSection, PageSectionVariants } from '@patternfly/react-core';
+import { TabbedRules } from 'PresentationalComponents';
+
+const SSGTabTitle = ({ profile: { ssgVersion, rules = [] } }) => (
+    <TabTitleText>
+        <span>SSG { ssgVersion + ' ' }</span>
+        <Label color="blue">{ rules.length }</Label>
+    </TabTitleText>
+);
+
+SSGTabTitle.propTypes = {
+    profile: propTypes.object.isRequired
+};
+
+const mapProfileToTabs = (profiles) => (
+    profiles.map((profile) => (
+        {
+            profile,
+            title: <SSGTabTitle profile={ profile } />,
+            rules: profile.rules
+        }
+    ))
+);
 
 const PolicyMultiversionRules = ({ policy: { policy: { profiles } } }) => (
     <React.Fragment>
         <Alert variant="info" isInline title="Rule editing coming soon" />
         <PageSection variant={ PageSectionVariants.light }>
-            <RoutedTabs level={ 1 } defaultTab={ profiles[0].ssgVersion }>
-                {
-                    profiles.map((profile) => (
-                        <Tab
-                            key={ `ssgversion-tab-${ profile.ssgVersion }` }
-                            title={
-                                <TabTitleText>
-                                    <span>SSG { profile.ssgVersion + ' ' }</span>
-                                    <Label color="blue">{ profile.rules.length }</Label>
-                                </TabTitleText>
-                            }
-                            eventKey={ profile.ssgVersion }>
-                            <SystemRulesTable
-                                remediationsEnabled={false}
-                                columns={ selectRulesTableColumns(['Name', 'Severity', 'Ansible']) }
-                                profileRules={[{
-                                    profile: { ...profile },
-                                    rules: profile.rules
-                                }]} />
-                        </Tab>
-                    ))
-                }
-            </RoutedTabs>
+            <TabbedRules
+                tabsData={ mapProfileToTabs(profiles) }
+                columns={ ['Name', 'Severity', 'Ansible'] }
+                level={ 1 } />
         </PageSection>
     </React.Fragment>
 );
--- a/src/PresentationalComponents/TabbedRules/TabbedRules.test.js
+++ b/src/PresentationalComponents/TabbedRules/TabbedRules.test.js
@@ -1,0 +1,67 @@
+import TabbedRules from './TabbedRules';
+import { policies } from '@/__fixtures__/policies';
+import { TabTitleText } from '@patternfly/react-core';
+
+jest.mock('react-router-dom', () => ({
+    ...jest.requireActual('react-router-dom'),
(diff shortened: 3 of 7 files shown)


src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -1,38 +1,25 @@
...
 import propTypes from 'prop-types';
-import { Alert, Label, Tab, TabTitleText, PageSection, PageSectionVariants } from '@patternfly/react-core';
-import SystemRulesTable, {
-    selectColumns as selectRulesTableColumns
-} from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
-import { RoutedTabs } from 'PresentationalComponents';
+import { Alert, Label, TabTitleText, PageSection, PageSectionVariants } from '@patternfly/react-core';
+import { TabbedRules } from 'PresentationalComponents';
+
+const tabTitle = ({ ssgVersion, rules }) => (
```
bastilian, line 6: This would probably be better to be a real Component in `TabbedRules`

vkrizan: Can you please give an example with a callback?

src/PresentationalComponents/TabbedRules/TabbedRules.js
```diff
@@ -0,0 +1,46 @@
...
+} from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
+import { RoutedTabs } from 'PresentationalComponents';
+
+const TabbedRules = ({ profilesAndRules, tabTitle, defaultProfileId, columns, level, handleSelect, ...rulesTableProps }) => (
+    <RoutedTabs level={ level } defaultTab={ defaultProfileId }>
+        {
+            profilesAndRules?.map(({ profile, rules }) => (
+                <Tab
+                    key={ 'multiversion-rules-' + profile.id }
+                    title={ tabTitle(profile) }
```
bastilian, line 15: Related to the other comment. This should be something like 

```js 
title={ <TabTitle ssgVersion={ profile.ssgVersion } rules={ profile.rules.length } /> } 
```

with a component above like:

```js
const TabTitle = ({ ssgVersion, rules }) => (
    <TabTitleText>
        <span>SSG { ssgVersion + ' ' }</span>
        <Label color="blue">{ rules  }</Label>
    </TabTitleText>
);
```

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -1,38 +1,31 @@
...
-                                    rules: profile.rules
-                                }]} />
-                        </Tab>
-                    ))
-                }
-            </RoutedTabs>
+            <TabbedRules
+                profilesAndRules={ mapProfilesWithRules(profiles) }
+                tabTitle={ tabTitle }
+                defaultProfileId={ profiles[0].id }
```
bastilian, line 26: Could the first profile id not be the default within the `TabbedRules` component and only require setting it when necessary and the first one should not be the default opened?

src/PresentationalComponents/TabbedRules/TabbedRules.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Tab } from '@patternfly/react-core';
+import SystemRulesTable, {
+    selectColumns as selectRulesTableColumns
+} from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
+import { RoutedTabs } from 'PresentationalComponents';
+
+const TabbedRules = ({ profilesAndRules, tabTitle, defaultProfileId, columns, level, handleSelect, ...rulesTableProps }) => (
```
bastilian, line 9: Default props can also be set by just assigning here like:

```js
{ profilesAndRules = [], columns=['Name'], level=1 ...
```

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: That would hardcore the behavior of tab title. Details page shows SSG versions (now), modal would show OS versions. Details page shows number of rules (now), modal would show number of assigned or selected hosts.

I'm targeting to use the callback to construct the title with data that are not directly provided for the `TabbedRules` component. For example number of hosts wouldn't be passed, but the title can still be drawn with it.

It might be changed to some common interface:
```
title={ <tabTitleComponent  profile={profile} /> }
```

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: It's already long enough. Do you know how to format it to several lines?

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
vkrizan: This does not change the opening behavior. The Policy details page does open first tab.

src/PresentationalComponents/TabbedRules/TabbedRules.js
bastilian: In both (all multi OS version rules table) instances the title should be the same and show the RHEL version and assigned hosts, we do not need to preserve the rules count and and SSG version, since it is actually a bug.

src/PresentationalComponents/TabbedRules/TabbedRules.js
bastilian: Press the enter key at convenient locations? :D

vkrizan: It looks terrible. Sticking with `defaultProps` this time.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: Once that happens we can update it. At the moment lets do iterative changes that make the presentational component more robust/open.

As mentioned earlier, with callback it is possible to inject other values that are within the scope of a (anonymous) function.

bastilian: Using a "callback" like that is not a great approach. Passing in the a rendered component rather than a function to be called to render, would be better.

vkrizan: I agree, but cannot think of something better. When this would be utilized within the modals its usability might be more clear or some other approach might arise.

src/PresentationalComponents/TabbedRules/TabbedRules.js
bastilian: It doesn't look too terrible: 

```js
const TabbedRules = ({
    columns = ['Name', 'Severity', 'Ansible'],
    defaultProfileId,
    handleSelect,
    level = 0,
    profilesAndRules,
    tabTitle,
    ...rulesTableProps
}) => {
    // ...
};
```

It also lifts up the defaults in the file and can be seen right away.

src/PresentationalComponents/TabbedRules/TabbedRules.js
```diff
@@ -0,0 +1,46 @@
...
+    <RoutedTabs level={ level } defaultTab={ defaultProfileId }>
```
akofink, line 10: This `level` used to always be 1 before this. I guess it will be 1 in the policy details -> rules view, but it will be 0 in the edit policy modal. Is that correct?

src/PresentationalComponents/TabbedRules/TabbedRules.js
```diff
@@ -0,0 +1,46 @@
...
+);
+
+TabbedRules.propTypes = {
+    profilesAndRules: propTypes.arrayOf(propTypes.object),
+    tabTitle: propTypes.func.isRequired,
+    defaultProfileId: propTypes.string,
+    columns: propTypes.arrayOf(propTypes.string),
+    level: propTypes.number,
+    handleSelect: propTypes.func
+};
```
akofink, line 39: So, the idea of this new component is to use it for policy details, edit policy modal, and create policy wizard, right? Are you planning to add some prop like `allowSelection` or similar to toggle between the rule display vs. rule selection behavior?

akofink (approved):
Thanks, @vkrizan! I only have a couple small questions, but it seems to continue to work as expected, and I think I understand the intended uses for this new component. I'm interested in your vision, though, for how we'd use it (additional props, etc.) in edit/create policy views.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: Yes that's the idea. You can pass `tailoringEnabled` boolean. It's redirected to the SystemsRulesTable.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: Yes, level 1 in details and level 0 in create/edit modal. 

Disabling the routing isn't possible with this component. If that would be needed, we can split it to two components, one with routing one without routing. That's the easiest approach.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: Other way could be to provide titles within the array of data. Current `profilesAndRules` would contain array of `{profile, title, rules}`.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: Sorry, `tailoringEnabled` is not even used, instead the `handleSelect` should be set.

src/PresentationalComponents/TabbedRules/TabbedRules.js
bastilian: Actually, we only need to cover the OS tab, because the SSG versions are behind a feature flag still.

So it should not be necessary at all to provide a function or component from the outside at all.

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
bastilian: What I mean is do we really need to provide a `defaultProfileId`, can we not just set the first item in `profilesAndRules` as the `defaultTab` for the `RoutedTabs` component within `TabbedRules`?

src/PresentationalComponents/TabbedRules/TabbedRules.js
akofink: I think the routing is pretty useful so far. It'd be great to be able to link users directly to the rule set for editing or viewing. At least in this use case, I don't see us needing tabs without routing.

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
vkrizan: Oh, right! We can do have a default for default. :slightly_smiling_face:

vkrizan:
Okay, I've cleaned it up a bit. Changed the title callback to a node provided through `tabsData` and set auto defaulting to first item. Event keys have `rules-` prepended to avoid possible conflicts.

Would you guys please have one last peak?

akofink (approved):
Seems to continue working as expected. Thanks!

src/SmartComponents/PolicyDetails/PolicyMultiversionRules.js
```diff
@@ -1,38 +1,37 @@
...
-import { RoutedTabs } from 'PresentationalComponents';
+import { Alert, Label, TabTitleText, PageSection, PageSectionVariants } from '@patternfly/react-core';
+import { TabbedRules } from 'PresentationalComponents';
+
+const SSGTabTitle = ({ profile: { ssgVersion, rules = [] } }) => (
+    <TabTitleText>
+        <span>SSG { ssgVersion + ' ' }</span>
+        <Label color="blue">{ rules.length }</Label>
+    </TabTitleText>
+);
```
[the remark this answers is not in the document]
vkrizan: Sure that sounds doable. It would likely be flat, like `SSGRulesTabTitle`. This one would eventually go away. We can definitely use this suggestion for the OS tab that'd be used thought the app.

PR #1045, opened 2021-03-12
chore(deps): Bump @patternfly & @redhat-cloud-services packages

bastilian:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


package.json
```diff
@@ -142,6 +142,8 @@
     "travis:build": "NODE_OPTIONS=--max_old_space_size=4096 NODE_ENV=production webpack --config config/prod.webpack.config.js",
     "travis:verify": "npm-run-all travis:build lint test coverage",
     "verify": "npm-run-all build lint test",
-    "nightly": "npm run travis:verify"
+    "nightly": "npm run travis:verify",
+    "bump:redhatservices": "npm install @redhat-cloud-services/frontend-components-notifications@latest @redhat-cloud-services/frontend-components-utilities@latest @redhat-cloud-services/frontend-components-config@latest @redhat-cloud-services/frontend-components@latest @redhat-cloud-services/frontend-components-inventory-compliance@latest",
+    "bump:patternfly": "npm install @patternfly/react-core@latest @patternfly/react-tables@latest @patternfly/react-charts@latest @patternfly/react-tokens@latest @patternfly/react-icons@latest"
```
akofink, line 147: :+1:

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -9,7 +9,7 @@ import { connect } from 'react-redux';
 import useFeature from 'Utilities/hooks/useFeature';
 import { systemName } from 'Store/Reducers/SystemStore';
 
-const EditPolicySystems = ({ change, osMajorVersion, selectedSystemIds }) => {
+export const EditPolicySystems = ({ change, osMajorVersion, selectedSystemIds }) => {
```
akofink, line 12: What causes this change to be required?

akofink (commented):
The code changes seem ok to me, but I'm curious why the tests had to be updated in a few places? For example, InventoryDetails test no longer requires a mock store? Why, some changes with `@redhat-cloud-services/frontend-components-utilities/Registry`?

akofink (approved):
Thanks! Seems to continue working as expected. I guess the tests are also changed to only test our components, not necessarily the versions wrapped with redux. Is that right? Did the package change require us to update our tests in this way? It seems nice to me that we don't have to add those mocks anymore to our tests, but I'm also worried that the test isn't as effective without the redux stuff. I really don't know; I'd like to get your thoughts on this please.

package.json
```diff
@@ -11,12 +11,12 @@
...
-    "@patternfly/react-icons": "^4.9.2",
-    "@patternfly/react-table": "^4.23.2",
-    "@patternfly/react-tokens": "^4.10.2",
-    "@redhat-cloud-services/frontend-components": "^3.0.10",
+    "@patternfly/react-charts": "^6.14.6",
+    "@patternfly/react-core": "^4.101.3",
+    "@patternfly/react-icons": "^4.9.5",
+    "@patternfly/react-table": "^4.23.14",
+    "@patternfly/react-tokens": "^4.10.5",
+    "@redhat-cloud-services/frontend-components": "3.0.10",
```
bastilian, line 19: There is a fix in progress to be able and unpin it and use the latest version.

akofink (approved):
Thanks!

bastilian:
@akofink I dug into the changes to the tests again and reverted them. It seems the previous lockfile had updated some dependencies that caused issues and prompted the updated to move to only use shallow, like we do in other tests. I tried reproducing these issues again with the new lockfile, but it appears to not require any changes and tests pass, and I can't remember exactly what the issue was and why I chose to only use `shallow`

> It seems nice to me that we don't have to add those mocks anymore to our tests, but I'm also worried that the test isn't as effective without the redux stuff. I really don't know; I'd like to get your thoughts on this please.

Using `shallow` is less effective from a certain perspective, since it goes further in the React rendering and invokes third party code. However in this case it made no big difference as the test only verified that it renders with the props and conditions given. With `mount` the conditions are also setup by dependencies, while with `shallow` you'd actually have to mock what dependencies would provide.

package.json
bastilian: After insights-chrome v2

PR #1049, opened 2021-03-16
feat(Hooks): Add useCollection and REST API client

bastilian:
This adds a new hook for interacting with the Compliance REST API.

Example usage: 

```js
const Component = () => {
  const { collection: supportedSSG } = useCollection('supported_ssgs');
  
  return supportedSSG;
};
```

Example return:

```js
{ 
  collection: normalized,
  meta: json.meta,
  total: json.meta.total,
  json
}
```

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


package.json
```diff
@@ -26,6 +26,7 @@
     "classnames": "^2.2.6",
     "graphql": "^15.4.0",
     "graphql-tag": "^2.11.0",
+    "json-api-normalizer": "^1.0.1",
```
bastilian, line 29: The latest version `1.0.3` causes an missing module error. 🤷

src/Utilities/hooks/api/useCollection.js
```diff
@@ -0,0 +1,61 @@
...
+            includeAttributes(normalizedJson?.[entity.type]?.[entity.id])
+        )).filter((v) => (!!v));
+    });
+    delete entity.relationships;
+    return { ...entity, ...relationships };
+};
+
+const normalizeData = (json, type) => {
+    const jsonNormalized = normalize(json);
+    return Object.values(jsonNormalized[type] || {})?.map((entity) => (
```
bastilian, line 30: This makes sure that if a resource has a relation ship, that this relation ship is accessible via a prop.

src/Utilities/ApiClient.js
```diff
@@ -0,0 +1,107 @@
...
+        return fetch(
+            url, params
+        );
+    }
+
+    checkForEmptyResponse(response) {
+        return response.status === 204 ? { json: () => ({}) } : response;
+    }
+
+    checkForErrors(response, options = {}) {
```
bastilian, line 60: We should check how this behaves within the Compliance frontend and with using it within a hook.

bastilian:
@vkrizan @akofink Here's a first version of an API client, for now it's just `useCollection`, which seems to be enough to cover our first cases where we'd want to use it. I'm not 100% sure about how the hook is implemented and meant to be used.

src/Utilities/hooks/api/useCollection.js
```diff
@@ -0,0 +1,61 @@
...
+
+    return {
+        collection: normalized,
+        meta: json.meta,
+        total: json.meta.total,
+        json
+    };
+};
+
+const useCollection = (collection, options = {}) => {
```
bastilian, line 47: We could move the API client setup into a `useFetch` hook and/or swap out that part with a more robust client, ...like `useFetch`[0]. This hook then would only do the normalising and preparing the data for better use in components.

[0] https://use-http.com/#/

package.json
akofink: Does this change result in the -20k lines removed in package-lock.json?

bastilian: That might not be right, I think this may be due to the lockfile being updated with a different versions of npm than what was used to built our current lock file. I'll check to make sure the changes to the lockfile are proper.

akofink (approved):
Thanks, @bastilian. IMO we should go ahead and merge this even (without the package-lock.json changes) if we're not going to use it right away. Hopefully including it here will help with future REST-related changes. The code seems reasonable to me but a bit difficult to thoroughly test, but if we find issues later, we can fix them later. :)

vkrizan (approved):
Awesome stuff! I guess this deserves to be a library.

Would you mind adding more tests?

bastilian:
@akofink @vkrizan Thank you for taking a look!

Yeah, this will hopefully help us get started to use the REST API in the frontend. In the long run it would be great to be able to move to the javascript-clients[0] as a replacement for the `ApiClient` class, if we can.

Based on that it will make more sense to move it to the frontend-components-utilities package for other applications to use.

I'll add some tests for the hooks and also split out a generic `useFetch` hook that can be used without the normalisation shenanigans.

[0] https://github.com/RedHatInsights/javascript-clients

package.json
bastilian: We seem to have at some point bumped the lockfileVersion. This change is now the same lockfileVersion, but it still updates quite a lot, most notably the apollo-client (a dependency of the apollo packages). And it was needed to use the `--legacy-deps` flag.

It doesn't seem to cause any issues though, but we should still look into it a bit. I'll take a look at https://github.com/RedHatInsights/compliance-frontend/pull/1045 again and get it mergable. 
It's possible that recent config package updates could have solved the issue with the inventory there and reduce the change-set here to only the normaliser package related changes.

akofink:
> In the long run it would be great to be able to move to the javascript-clients[0] as a replacement for the ApiClient class, if we can.

Our openapi.json probably needs to be compliant with OpenAPI v3 for this to work. QE also wants a compliant openapi.json. It may be time to just write it ourselves instead of waiting any longer for rswag to be fixed.

package.json
bastilian: The package PR[0] now downgrades the lockfile to version 1.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1045

src/Utilities/hooks/api/utils/normalize.test.js
```diff
@@ -0,0 +1,20 @@
+import supportedSsgs from '../__fixtures__/supported_ssgs.json';
```
bastilian, line 1: Tests moved and adjusted from the "hooks" project: https://github.com/RedHatInsights/hooks-frontend/blob/master/src/store/reducers/reducerHelper.test.js#L56

src/Utilities/hooks/api/utils/ApiClient.test.js
```diff
@@ -0,0 +1,138 @@
+import { default as ApiClientClass } from './ApiClient';
```
bastilian, line 1: Teste moved and adjusted from "hooks" project: https://github.com/RedHatInsights/hooks-frontend/blob/master/src/Utilities/notificationsBackendAPI.test.js

src/Utilities/hooks/api/utils/normalize.test.js
```diff
@@ -0,0 +1,20 @@
...
+import { normalizeData } from './normalize';
+
+describe('normalizeData', () => {
+    it('normalizes API data', () => {
+        expect(normalizeData(supportedSsgs, 'supportedSsg', '/supported_ssgs')).toMatchSnapshot();
+    });
+
+//     it('sorts it when sortBy is given', () => {
```
bastilian, line 9: These tests pass locally, but for some reason the sorting acts differently on CI. 🤷

bastilian:
@akofink The package lock file updates are now right here.
@vkrizan I moved some tests from the Hooks project over to here. Some are failing on CI due to some odd issue that causes the sorting to be different. I commented those out for now.

There is now a `useAPI` to use the API more directly.

Unless there are any objections, I think we can merge this now.

akofink:
:+1: let's do it! Hopefully those sorting differences can be figured out and not forgotten, but since this code won't be in use yet, I can't see how it would hurt.

PR #1050, opened 2021-03-16
feat(CreatePolicy): RHICOMPL-1521 gather selected OS minor versions

vkrizan:
Gathers OS minor versions of selected systems within the PolicyCreate wizard using `countOsMinorVersions` reducer. The result is sorted naturally.

Adds a PoC within the Rules step in the wizard to see the result (requires `?multiversionTabs=enable`). This PoC would likely be fully replaced/reverted with implementation of the tabs for each OS version. One step at a time...

PoC ("Tailoring for ...")
Screenshot_2021-03-16 SCAP policies - Compliance Red Hat Insights

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicyRules.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyRules.js
@@ -14,6 +14,7 @@ import { useQuery } from '@apollo/react-hooks';
 import {
     StateViewWithError, StateViewPart
 } from 'PresentationalComponents';
+import useFeature from 'Utilities/hooks/useFeature';
 
 const QUERY = gql`
 query benchmarkAndProfile($benchmarkId: String!, $profileId: String!){
@@ -42,7 +43,8 @@ query benchmarkAndProfile($benchmarkId: String!, $profileId: String!){
 }
 `;
 
-export const EditPolicyRules = ({ profileId, benchmarkId, selectedRuleRefIds, change }) => {
+export const EditPolicyRules = ({ profileId, benchmarkId, osMajorVersion, osMinorVersionCounts, selectedRuleRefIds, change }) => {
+    const multiversionRules = useFeature('multiversionTabs');
     const columns = selectRulesTableColumns(['Name', 'Severity', 'Ansible']);
     const { data, error, loading } = useQuery(QUERY, { variables: { profileId, benchmarkId } });
     const [defaultSelection, setDefaultSelection] = useState(null);
@@ -90,6 +92,13 @@ export const EditPolicyRules = ({ profileId, benchmarkId, selectedRuleRefIds, ch
                         </Button>
                     }
                 </Text>
+                { multiversionRules && osMinorVersionCounts && osMinorVersionCounts.length > 0 &&
+                    <Text>
+                        Tailoring for{' '}
+                        { osMinorVersionCounts.map(({ osMinorVersion, count }) =>
+                            `RHEL ${osMajorVersion}.${osMinorVersion} (${count} systems)`).join(', ') }
+                    </Text>
+                }
             </TextContent>
             <SystemRulesTable
                 remediationsEnabled={ false }
@@ -109,6 +118,11 @@ EditPolicyRules.propTypes = {
     profileId: propTypes.string,
     benchmarkId: propTypes.string,
     change: reduxFormPropTypes.change,
+    osMajorVersion: propTypes.string,
+    osMinorVersionCounts: propTypes.arrayOf(propTypes.shape({
+        osMinorVersion: propTypes.number,
+        count: propTypes.number
+    })),
     selectedRuleRefIds: propTypes.array
 };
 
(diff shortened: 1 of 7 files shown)


src/store/Reducers/SystemStore.js
```diff
@@ -116,6 +117,20 @@ const profilesSsgVersions = ({ testResultProfiles: profiles = [] }) => (
...
+
+    let counted = systems.reduce((acc, { osMinorVersion }) => {
+        if (osMinorVersion) {
+            (acc[osMinorVersion] = acc[osMinorVersion] || { osMinorVersion, count: 0 }).count++;
+        }
+
+        return acc;
+    }, {});
+    counted = Object.values(counted);
+    return counted.sort((a, b) => naturalSorting(a.osMinorVersion, b.osMinorVersion));
```
vkrizan, line 131: I've later found out that the `osMinorVersion` is a number (not a string), however the `naturalSorting` still works correctly for numbers. I've kept it as is, otherwise there would have to be another sorting mechanism just for numbers of a selected property.

src/SmartComponents/CreatePolicy/EditPolicyRules.js
```diff
@@ -90,6 +92,13 @@ export const EditPolicyRules = ({ profileId, benchmarkId, selectedRuleRefIds, ch
                         </Button>
                     }
                 </Text>
+                { multiversionRules && osMinorVersionCounts && osMinorVersionCounts.length > 0 &&
+                    <Text>
+                        Tailoring for{' '}
+                        { osMinorVersionCounts.map(({ osMinorVersion, count }) =>
+                            `RHEL ${osMajorVersion}.${osMinorVersion} (${count} systems)`).join(', ') }
```
vkrizan, line 99: PoC mostly for testing purposes to verify it in action. Please set `?multiversionTabs=enable`.

src/store/Reducers/SystemStore.js
vkrizan: Taking it back. It does not work with zeros nor empty strings.

skateman: Shouldn't we need a test for numbers as well then?

src/store/Reducers/SystemStore.js
```diff
@@ -116,6 +117,20 @@ const profilesSsgVersions = ({ testResultProfiles: profiles = [] }) => (
...
+    if (!systems) { return []; }
+
+    let counted = systems.reduce((acc, { osMinorVersion }) => {
+        if (osMinorVersion !== undefined && osMinorVersion !== null) {
+            (acc[osMinorVersion] = acc[osMinorVersion] || { osMinorVersion, count: 0 }).count++;
+        }
+
+        return acc;
+    }, {});
+    counted = Object.values(counted);
```
skateman, line 130: I'm not sure how much are we forcing the use of immutable stuff, but here it's kinda easy to avoid the `let` and the value reassignment. AFAIK you can just call `values` directly on an object and you can prevent the mutation of the `acc` variable as well.

src/Utilities/helpers.js
```diff
@@ -1 +1,15 @@
 export const uniq = (collection) => [...new Set(collection)];
+
+export const sortingByProp = (propName) => (
```
bastilian, line 3: A similar function is going to be introduced with the REST API client[0]. If this function here would also implement a directional argument it could be used in the REST API client PR as well. 

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1049/files#diff-782407e360d7925f7133d239de9109d43bdf505d08de3542078de388da2d0896R64

akofink:
@vkrizan would you mind please fixing the linting issues so we can build+test without disabling eslint?

vkrizan:
> would you mind please fixing the linting issues so we can build+test without disabling eslint?

Done. Sorry about that.

src/Utilities/helpers.js
vkrizan: Added.

Example:
```
array.sort(sortingByProp('propName', 'desc'))
```
There is also a built-in [`reverse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse).

src/SmartComponents/CreatePolicy/EditPolicyRules.js
akofink: Nice :) Seems to work great

akofink (approved):
Seems to work as expected, and the code makes sense. :+1:

src/store/Reducers/SystemStore.js
skateman: Yeah, my bad :see_no_evil: doing too much Ruby these days. But the immutability can be still done.

vkrizan: Done.

PR #1057, opened 2021-03-19
feat(CreatePolicy): RHICOMPL-1546 display OS version for systems

skateman:
The Policies column has been replaced with the OS version which has been extracted from beneath the display name.

Just for the sake of consistency, should this column also be displayed for the non-wizard when editing a policy?

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -25,6 +25,33 @@ export const EditPolicy = ({ route }) => {
     const selectedEntities = useSelector((state) => (state?.entities?.selectedEntities));
     const saveEnabled = updatedPolicy && !updatedPolicy.complianceThresholdValid;
 
+    const columns = [
+        {
+            key: 'facts.compliance.display_name',
+            title: 'Name',
+            props: {
+                width: 40, isStatic: true
+            },
+            ...newInventory && {
+                key: 'display_name',
+                renderFunc: (displayName, id, extra) => {
+                    return extra?.lastScanned ? systemName(displayName, id, { name: extra?.name }) : displayName;
+                }
+            }
+        },
+        {
+            key: 'facts.compliance.osMinorVersion',
+            title: 'Operating system',
+            props: {
+                width: 40, isStatic: true
+            },
+            ...newInventory && {
+                key: 'osMinorVersion',
+                renderFunc: (osMinorVersion, _id, { osMajorVersion }) => `RHEL ${osMajorVersion}.${osMinorVersion}`
+            }
+        }
+    ];
+
     const linkToBackgroundWithHash = () => {
         newInventory && dispatch({
             type: 'SELECT_ENTITIES',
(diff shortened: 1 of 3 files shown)


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -25,6 +25,33 @@ export const EditPolicy = ({ route }) => {
     const selectedEntities = useSelector((state) => (state?.entities?.selectedEntities));
     const saveEnabled = updatedPolicy && !updatedPolicy.complianceThresholdValid;
 
+    const columns = [
+        {
+            key: 'facts.compliance.display_name',
+            title: 'Namee',
```
akofink, line 31:
```suggestion
            title: 'Name',
```

src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -25,6 +25,33 @@ export const EditPolicy = ({ route }) => {
     const selectedEntities = useSelector((state) => (state?.entities?.selectedEntities));
     const saveEnabled = updatedPolicy && !updatedPolicy.complianceThresholdValid;
 
+    const columns = [
```
akofink, line 28: This is also defined identically in `EditPolicySystems`. Maybe it'd make sense to consolidate them.

akofink (approved):
Works as expected, thanks! Just one typo.

akofink:
Also, I believe we need to add filters here for OS, but that can wait for a future PR.

src/SmartComponents/EditPolicy/EditPolicy.js
skateman: I was trying to find the right file :see_no_evil: :rofl:

src/SmartComponents/EditPolicy/EditPolicy.js
skateman: Umm, where exactly?

akofink: `src/SmartComponents/CreatePolicy/EditPolicySystems.js` This is the systems table shown in the policy create wizard (the component name is a tad confusing)

akofink: Maybe @bastilian can recommend a location for these columns, like perhaps `SystemsTable/constants.js`?

bastilian: For now this is fine and we can move these together somewhere later. I agree that the definitions are the the same and should be consolidated, but the definition of columns, as in what columns are in the array, should still be possible to see in here. So the individual column definitions should be importable from a columns module, to be put into the array, and possibly extend, the same as what we are working towards with the way we want to define columns for the `SystemRulesTable` component.

There is also already some work towards having this for the `SystemTable`/`InventoryTable`  here[0], which we could move over to here, but I'm fine with just changing how the operating system is shown in this PR.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1005/files#diff-990cfe23e4e9e522e0e9e5f3e72349f5b95ce50bf0afe392ddefded4fd9a882e

akofink:
@skateman feel free to rebase with master and merge

PR #1062, opened 2021-03-23
feat(CreatePolicy): RHICOMPL-1548 validate >0 systems selection

skateman:
Added a validation that prevents the wizard to continue from the systems step if no systems are selected. 

Is the snapshot sufficient for testing here, or should I add something else and if yes, where and what?
Should this validation also happen when editing a policy?

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreatePolicy.test.js
+++ b/src/SmartComponents/CreatePolicy/CreatePolicy.test.js
@@ -17,4 +17,16 @@ describe('CreatePolicy', () => {
         );
         expect(toJson(wrapper.find('Wizard'))).toMatchSnapshot();
     });
+
+    it('expect to render the wizard with enableNext on systems', () => {
+        useQuery.mockImplementation(() => ({
+            data: { latestBenchmarks: benchmarksQuery },
+            error: false,
+            loading: false
+        }));
+        const wrapper = shallow(
+            <CreatePolicy systemIds={['123', '456']}/>
+        );
+        expect(wrapper.find('Wizard').prop('steps').find(({ name }) => name === 'Systems').enableNext).toMatchSnapshot();
+    });
 });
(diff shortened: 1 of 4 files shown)


bastilian:
@skateman Nice! Thank you! It would definitely help to add a test. You'd need to duplicate the test that is already in the test file for the component[0] with a `systemIds` prop set. According to that the `enableNext` property should be `true` in the snapshot. That should do it.

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/CreatePolicy.test.js#L9

src/SmartComponents/CreatePolicy/CreatePolicy.test.js
```diff
@@ -17,4 +17,17 @@ describe('CreatePolicy', () => {
...
+        useQuery.mockImplementation(() => ({
+            data: { latestBenchmarks: benchmarksQuery },
+            error: false,
+            loading: false
+        }));
+        const wrapper = shallow(
+            <CreatePolicy systemIds={['123', '456']}/>
+        );
+        expect(toJson(wrapper.find('Wizard'))).toMatchSnapshot();
+        expect(wrapper.find('Wizard').prop('steps').find(({ name }) => name === 'Systems').enableNext).toBe(true);
```
bastilian, line 31: This is not really necessary, as the snapshots already keep track of the value in the object.

You could write a hybrid version to limit the snapshot to specific part of the object like:
```js
expect(wrapper.find('Wizard').prop('steps').find(({ name }) => name === 'Systems').enableNext).toMatchSnapshot();
```

bastilian (approved):
Thank you @skateman!

PR #1067, opened 2021-03-24
feat(CreatePolicy): RHICOMPL-1547 implemented empty state for systems

skateman:
Added optional `emptyStateComponent` and `titleComponent` attributes to the `InventoryTable` component that are invoked from the `SystemInventory` in the create policy wizard. The goal is to show an informative message if there are no systems available with the selected OS major version and provide a button to go back two steps to reselect the OS major.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.js
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
 import { propTypes as reduxFormPropTypes, reduxForm, formValueSelector } from 'redux-form';
-import { Form, FormGroup, Text, TextContent, TextVariants } from '@patternfly/react-core';
+import { Button, Form, FormGroup, Text, TextContent, TextVariants, WizardContextConsumer } from '@patternfly/react-core';
 import { InventoryTable, SystemsTable } from 'SmartComponents';
 import { GET_SYSTEMS_WITHOUT_FAILED_RULES } from '../SystemsTable/constants';
 import { compose } from 'redux';
@@ -11,6 +11,8 @@ import { systemName, countOsMinorVersions } from 'Store/Reducers/SystemStore';
 
 const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selectedSystemIds }) => {
     const newInventory = useFeature('newInventory');
+    const multiversionRules = useFeature('multiversionTabs');
+
     const columns = [{
         key: 'facts.compliance.display_name',
         title: 'Name',
@@ -43,6 +45,33 @@ const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selec
         }
     }, [selectedSystemIds, osMinorVersionCounts, change]);
 
+    const emptyStateComponent = (<React.Fragment>
+        <TextContent className="pf-u-mb-md">
+            <Text>
+                You do not have any <b>RHEL { osMajorVersion }</b> systems connected to Insights and enabled for Compliance.<br/>
+                Policies must be created with at least one system.
+            </Text>
+        </TextContent>
+        <TextContent className="pf-u-mb-md">
+            <Text>
+                Choose a different operating system, or connect RHEL { osMajorVersion } systems to Insights.
+            </Text>
+        </TextContent>
+        <WizardContextConsumer>
+            { ({ goToStepById }) => <Button onClick={() => goToStepById(1)}>Choose a different operating system</Button> }
+        </WizardContextConsumer>
+    </React.Fragment>);
+
+    const prependComponent = (<React.Fragment>
+        <TextContent className="pf-u-mb-md">
+            <Text>
+                Select which of your <b>RHEL { osMajorVersion }</b> systems should be included
+                in this policy.<br />
+                Systems can be added or removed at any time.
+            </Text>
+        </TextContent>
+    </React.Fragment>);
+
     const InvCmp = newInventory ? InventoryTable : SystemsTable;
 
     return (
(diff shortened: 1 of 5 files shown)


src/SmartComponents/SystemsTable/InventoryTable.test.js
```diff
@@ -0,0 +1,109 @@
...
+
+    it('expect to render a title component', () => {
+        const wrapper = shallow(
+            <InventoryTable { ...defaultProps } titleComponent="Title" />
+        );
+
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it('expect to render an empty state component', () => {
```
skateman, line 82: I guess I'm not doing this properly :see_no_evil: at least the shallow mount isn't doing much difference between the 3 cases. Should I go just with a single case where I'm just verifying if all props are set? In that case how should I actually test the empty state?

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -205,7 +221,9 @@ InventoryTable.propTypes = {
     defaultFilter: PropTypes.string,
     systemProps: PropTypes.shape({
         isFullView: PropTypes.bool
-    })
+    }),
+    emptyStateComponent: PropTypes.any,
+    titleComponent: PropTypes.any
```
skateman, line 226: What's *The Right Way&trade;* here? Probably not the `any` :confused:

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -43,23 +43,48 @@ const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selec
...
-                    Select which of your <b>RHEL { osMajorVersion }</b> systems should be included
-                    in this policy.<br />
-                    Systems can be added or removed at any time.
-                </Text>
-            </TextContent>
             <Form>
                 <FormGroup>
                     <InvCmp
+                        titleComponent={titleComponent}
+                        emptyStateComponent={emptyStateComponent}
```
bastilian, line 87: Mind changing this behind to be behind the `newInventory` Feature flag. You could do it like so:

```js
{ ...multiversionRules ? { titleComponent: titleComponent, emptyStateComponent: emptyStateComponent } : {} }
```

and define the constant for the flag at the top of the Component 

```js
const multiversionRules = useFeature('multiversionTabs');
```

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: The proper propType is PropTypes.node for React components. :)

src/SmartComponents/CreatePolicy/EditPolicySystems.js
skateman: That would be probably not what we want, the `titleComponent` should be always passed as I moved it from the parent component. But I'll do the change accordingly.

skateman: @bastilian do we really need the feature flag here? It would significantly reduce the complexity if we don't need it and as far as I understand it's not reducing any functionality if we not have it.

bastilian: @skateman Why does the title component need to be passed in? Can it not be here just above the `InvCmp`?

skateman: According to the mockup, the title component should not be shown in the empty state.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -115,73 +123,81 @@ const InventoryTable = ({
         }
     };
 
-    return <StateView stateValues={{ error, noError: error === undefined }}>
+    return <StateView stateValues={{ error, noError: error === undefined && !empty, empty }}>
         <StateViewPart stateKey='error'>
-            <ErrorPage error={error}/>
+            <React.Fragment>
+                { !!titleComponent && titleComponent }
```
vkrizan, line 129: The title used within Create Policy wizard mentions that you "can select systems" while it loads and then it swaps the content to saying that you have no systems. UX-wise, this might be seen as unexpected.

A better variant could be to show only the heading ("Systems") and the loader without any other text content. Once its loaded and the state is known it could show the text content, depending on the conditions.

I'd suggest to look at whether the (h1) heading can be hardcoded to `<h1>Systems</h1>` within the component. Alternatively the heading variant can be as a param, or, the whole heading can be passed within a new param.

```
<InventoryTable 
    prependComponent={<p>Select systems</p>}
    emptyStateComponent={<p>No systems</p>}
    
    titleComponent={<h1>Systems</h1>} />
    # or
    titleVariant={TextVariants.h1} 
/>
```

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -115,73 +123,81 @@ const InventoryTable = ({
...
-                        onSelect: (_, format) => exportFromState(store.getState()?.entities, format)
-                    }
-                }}
-                {...showActions && {
-                    actions: [{
-                        title: 'View in inventory',
-                        onClick: (_event, _index, { id }) => {
-                            const beta = window.location.pathname.split('/')[1] === 'beta';
-                            window.location.href = `${window.location.origin}${beta ? '/beta' : ''}/insights/inventory/${id}`;
+            <React.Fragment>
```
vkrizan, line 137: Was there some parse error that lead to the Fragment wrapper?

vkrizan (commented):
Works as coded. The UX might need some care.  :slightly_smiling_face:

src/SmartComponents/SystemsTable/InventoryTable.js
skateman: Well, the titleComponent below makes it necessary.

src/SmartComponents/SystemsTable/InventoryTable.js
skateman: Hardcoding is not an option as there are pages where the component is used differently. The rest is doable...

skateman: It's not that good idea to hardcode the title as it should be in the same `TextContent` as the text beneath to not have styling glitches...

vkrizan: The `TextContent` is rendered as `<div>`. We might keep the heading within that context, and recompose what would go through the `prependComponent`.

vkrizan (approved):
Looks good and works as expected.

PR #1101, opened 2021-04-07
fix(PolicyCreate): RHICOMPL-484 editing Name and Description

vkrizan:
Name and Description were not editable unless the other fields were
edited first.  The useEffect was called on each Name/Desc. field value
change resetting it to original profile value(s).

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicyDetails.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyDetails.js
@@ -7,12 +7,14 @@ import { Form, FormGroup, Text, TextContent, TextVariants } from '@patternfly/re
 import { ReduxFormTextInput, ReduxFormTextArea } from 'PresentationalComponents/ReduxFormWrappers/ReduxFormWrappers';
 import { ProfileThresholdField } from 'PresentationalComponents';
 
-const EditPolicyDetails = ({ change, policy }) => {
+const EditPolicyDetails = ({ change, policy, refId }) => {
 
     useEffect(() => {
-        change('name', `${policy.name}`);
-        change('refId', `${policy.refId}`);
-        change('description', `${policy.description}`);
+        if (policy && policy.refId !== refId) {
+            change('name', `${policy.name}`);
+            change('refId', `${policy.refId}`);
+            change('description', `${policy.description}`);
+        }
     }, [policy]);
 
     return (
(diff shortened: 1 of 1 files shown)


src/SmartComponents/CreatePolicy/EditPolicyDetails.js
```diff
@@ -7,12 +7,14 @@ import { Form, FormGroup, Text, TextContent, TextVariants } from '@patternfly/re
...
     useEffect(() => {
-        change('name', `${policy.name}`);
-        change('refId', `${policy.refId}`);
-        change('description', `${policy.description}`);
+        if (policy && policy.refId !== refId) {
+            change('name', `${policy.name}`);
+            change('refId', `${policy.refId}`);
+            change('description', `${policy.description}`);
+        }
     }, [policy]);
```
akofink, line 18: Idk what the best practice is, but it seems weird to me that any update to the policy will `change` all three of these fields. Isn't it better to have each one `useEffect`?

```js
    useEffect(() => {
        change('name', `${policy.name}`);
    }, [policy?.name]);

    useEffect(() => {
        change('refId', `${policy.refId}`);
    }, [policy?.refId]);

    useEffect(() => {
        change('description', `${policy.description}`);
    }, [policy?.description]);
```

It seems to work correctly both ways. I'm also confused about the refId comparison - when does the policy refId not match the prop refId?

akofink (approved):
Works as expected. Thanks!

src/SmartComponents/CreatePolicy/EditPolicyDetails.js
vkrizan: Interesting. I guess that we can keep one `useEffect` but list all of the attributes. Splitting is not necessary as all values change after policy type reselection. 

The `refId` was there for comparison of a policy type change. I haven't thought about putting the attributes in the trigger array.

Thanks for the suggestion. I'll update it...

vkrizan: > when does the policy refId not match the prop refId?

* the `refId` comes from the redux selector, which is the value that is in the form
* the `policy.refId` comes from the JSON profile blob out of the form radio button (from selection of Policy Type)

I've kept it with one `useEffect` and just moved the condition to React.

vkrizan:
@akofink reverted it back to the original. The trigger `[policy?.refId]` did not worked as expected.

PR #1114, opened 2021-04-09
fix(PolicyEdit): RHICOMPL-1650 empty policy rules on initial load

vkrizan:
The rationale is that even-though the `SELECT_ENTITIES` is called to
select the systems, the `selectedEntities` variable stays undefined.
This probably due to the lack of `state.entities.selectedEntities`
within the redux, which is probably filled out only when the Systems are
drawn.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -72,6 +72,16 @@ query Profile($policyId: String!){
 }
 `;
 
+const profilesToOsMinorMap = (profiles, hosts) => (
+    (profiles || []).reduce((acc, profile) => {
+        if (profile.osMinorVersion !== '') {
+            acc[profile.osMinorVersion] ||= { osMinorVersion: profile.osMinorVersion, count: 0 };
+        }
+
+        return acc;
+    }, mapCountOsMinorVersions(hosts || []))
+);
+
 export const EditPolicy = ({ route }) => {
     const { policy_id: policyId } = useParams();
     let { data } = useQuery(MULTIVERSION_QUERY, {
@@ -143,20 +153,13 @@ export const EditPolicy = ({ route }) => {
         updateSelectedRuleRefIds();
 
         setOsMinorVersionCounts(
-            (policy?.policy?.profiles || []).reduce((acc, profile) => {
-                if (profile.osMinorVersion !== '') {
-                    acc[profile.osMinorVersion] ||= { osMinorVersion: profile.osMinorVersion, count: 0 };
-                }
-
-                return acc;
-            }, mapCountOsMinorVersions(selectedEntities))
+            profilesToOsMinorMap(policy?.policy?.profiles, selectedEntities)
         );
     }, [selectedEntities]);
 
     useEffect(() => setUpdatedPolicy({ ...updatedPolicy, selectedRuleRefIds }), [selectedRuleRefIds]);
 
     useEffect(() => {
-
         if (policy) {
             const complianceThresholdValid =
                 (policy.complianceThreshold < 101 && policy.complianceThreshold > 0);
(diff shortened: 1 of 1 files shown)


akofink (changes requested):
I'm not able to verify if it fixes the original issue, since I haven't been able to reproduce it. However, this seems to break when all systems are not selected.

Scenario:
1. Edit a policy
2. Add a new osMinorVersion system
3. See the new rule set in the rules tab
4. De-select all systems in the systems tab
5. See the new rule set is still in the rules tab, expected: the new rule set disappears

vkrizan:
@akofink Thank you for testing it. It should be fixed now.

My expectation that the `selectedEntities` would be an empty array if deselecting all systems was wrong.

src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -165,10 +168,14 @@ export const EditPolicy = ({ route }) => {
...
             });
             updateSelectedRuleRefIds();
+
             dispatch({
                 type: 'SELECT_ENTITIES',
                 payload: { ids: policy?.hosts || [] }
             });
+            setOsMinorVersionCounts(
+                profilesToOsMinorMap(policy.policy.profiles, policy.hosts)
+            );
```
akofink, line 178: So, I guess the idea is that, if SELECT_ENTITIES takes a while to update the redux state, this will set osMinorVersionCounts first, but then SELECT_ENTITIES will update osMinorVersionCounts again, but nothing will have changed. Seems reasonable to me. If it happens in the reverse order, it also should work. :+1:

akofink (approved):
Thanks! Seems to continue working as expected. Again, I'm not able to verify the bug fix, but the code makes sense, and I can understand how it might break if SELECT_ENTITIES takes a while to update the redux state.

src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -165,10 +168,14 @@ export const EditPolicy = ({ route }) => {
                 complianceThresholdValid
             });
             updateSelectedRuleRefIds();
+
             dispatch({
                 type: 'SELECT_ENTITIES',
                 payload: { ids: policy?.hosts || [] }
             });
+            setOsMinorVersionCounts(
+                profilesToOsMinorMap(policy.policy.profiles, policy.hosts)
```
akofink, line 177: Should this be `policy?.policy?.profiles` just in case?

src/SmartComponents/EditPolicy/EditPolicy.js
vkrizan: The `SELECT_ENTITIES` does not actually update anything, or better to say it does not seems to create `entities.selectedEntities` on the redux state. I guess that it only happens if it has the data loaded, hence the InventoryTable must be loaded (?). The useEffect on `selectedEntities` isn't called is it is `undefined` up to the point of the systems table render.

src/SmartComponents/EditPolicy/EditPolicy.js
vkrizan: Sure, can add it. I've trusted the backend to return the correct stuff. :slightly_smiling_face:

src/SmartComponents/EditPolicy/EditPolicy.js
akofink: `entities.selectedEntities` is set by our own reducer whenever SELECT_ENTITIES is fired: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/store/Reducers/SystemStore.js#L314

It's run `onLoad` here in the FECInventoryTable: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/InventoryTable.js#L149, so that seems to track with your understanding.

vkrizan: Thanks. Good to know. When the `onLoad` is being fired might be the issue that we are facing.

PR #1123, opened 2021-04-12
feat(CreatePolicy): RHICOMPL-1630 Empty state for missing policy type

akofink:
image

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
@@ -5,7 +5,10 @@ import {
 } from '@redhat-cloud-services/frontend-components-inventory-compliance/SystemRulesTable';
 import EmptyTable from '@redhat-cloud-services/frontend-components/EmptyTable';
 import Spinner from '@redhat-cloud-services/frontend-components/Spinner';
-import { Text, TextContent, TextVariants } from '@patternfly/react-core';
+import {
+    Title, Text, TextContent, TextVariants,
+    EmptyState, EmptyStateBody
+} from '@patternfly/react-core';
 import gql from 'graphql-tag';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -119,26 +122,27 @@ export const EditPolicyProfilesRules = ({ policy, selectedRuleRefIds, change, os
     tabsData = tabsData.filter(({ profile }) => !!profile);
 
     const filter = `${ (profileIds || []).map((i) => (`id = ${ i }`)).join(' OR ') }`;
+    const skipProfilesQuery = benchmarksLoading || filter.length === 0;
     const { data: profilesData, error, loading } = useQuery(PROFILES_QUERY, {
         variables: {
             filter
         },
-        skip: filter.length === 0
+        skip: skipProfilesQuery
     });
     const dataState = ((profileIds?.length > 0) ? profilesData : undefined);
     const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const noRuleSets = !error && !loadingState && profileIds?.length === 0;
+    const profiles = skipProfilesQuery ? [] : profilesData?.profiles.edges.map((p) => (p.node));
 
     useLayoutEffect(() => {
-        if (profilesData) {
-            const profiles = profilesData?.profiles.edges.map((p) => (p.node));
-
+        if (!loadingState) {
             change('selectedRuleRefIds', profiles.map((profile) => ({
                 id: profile.id,
                 ruleRefIds: selectedRuleRefIds?.find(({ id }) => id === profile.id)?.ruleRefIds ||
                             profile.rules.map((rule) => (rule.refId))
             })));
         }
-    }, [profilesData]);
+    }, [profiles, loadingState]);
 
     return <React.Fragment>
         <TextContent className='pf-u-pb-md'>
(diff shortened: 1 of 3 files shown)


src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
```diff
@@ -127,6 +130,7 @@ export const EditPolicyProfilesRules = ({ policy, selectedRuleRefIds, change, os
     });
     const dataState = ((profileIds?.length > 0) ? profilesData : undefined);
     const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const noRules = !error && !dataState && !loadingState;
```
akofink, line 133: Probably the weakest part of this PR. It's a very simple way to figure out this case, but in the future when we display messages for which systems are supported for the given policy type (COMP-E-178), we'll need to check the discrepancies between `osMinorVersionCounts` and `benchmarks` returned from the API.

src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
```diff
@@ -127,6 +130,7 @@ export const EditPolicyProfilesRules = ({ policy, selectedRuleRefIds, change, os
     });
     const dataState = ((profileIds?.length > 0) ? profilesData : undefined);
     const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const noRuleSets = !error && !dataState && !loadingState;
```
akofink, line 133: Probably the weakest part of this PR. It's a very simple way to figure out this case, but in the future when we display messages for which systems are supported for the given policy type (COMP-E-178), we'll need to check the discrepancies between `osMinorVersionCounts` and `benchmarks` returned from the API.

skateman:
I'm concerned that this might introduce a way to submit invalid data to our server similarly to 1636, but I can't verify it as my local setup is somehow not responding :confused:

vkrizan:
@skateman Indeed the "Next" button is enabled even if there are no rules to draw.

skateman:
> @skateman Indeed the "Next" button is enabled even if there are no rules to draw.

That's another issue, I'm talking about selecting a system with a "valid" OS, getting to the review page, then going back to the systems page, deselecting the "valid" system, selecting an "invalid" one and then returning to the review page.

vkrizan:
> I'm talking about selecting a system with a "valid" OS, getting to the review page, then going back to the systems page, deselecting the "valid" system, selecting an "invalid" one and then returning to the review page.

That is more broader problem. The review page will display system counts regardless of the tailoring or profile availability for that OS version. I guess that we will tackle this later within a different feature. Even if a user adds this "invalid" system, such system would not be able to scan, or if it would it would use the canonical set.

akofink:
There is no issue to have systems assigned to a policy which don't have corresponding rule sets. As Viliam said, the client will just skip over those policies. It's a scenario that can happen even if we try restricting assigning systems to a policy based on its OS, since OS versions can be upgraded or downgraded, so we must handle it.

That being said, it doesn't make sense for the wizard to allow users to create policies with no valid systems, so I'll work on updating the next/jumpto buttons.

akofink:
I spent some time trying to get the wizard to prevent users from going to steps they're not ready for. It works mostly, but there are a few ways I've found to bypass these restrictions. Of course, if a (nefarious) user really wants to bypass steps, it's not difficult, nor will it cause any major issues.

One shortcoming with the current implementation is that changes to step 3, Systems, have to wait until step 4, Rules, renders so the frontend can request the benchmarks and figure out if there are any OS versions missing. We could go ahead and set `selectedRuleRefIds` to `[]` to prevent this, but then going back to 3 to add systems and forward to 4 would clear any rules the user has already selected.

Let me know what you think. Thanks!

skateman:
What if we don't set the `selectedRuleRefIds˙ to `[]` but update it based on the systems selection, i.e. delete the ones that are coming from a minor OS version that is no longer in the list of selected systems?

src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
```diff
@@ -119,26 +122,27 @@ export const EditPolicyProfilesRules = ({ policy, selectedRuleRefIds, change, os
...
     const dataState = ((profileIds?.length > 0) ? profilesData : undefined);
     const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const noRuleSets = !error && !loadingState && profileIds?.length === 0;
+    const profiles = skipProfilesQuery ? [] : profilesData?.profiles.edges.map((p) => (p.node));
 
     useLayoutEffect(() => {
-        if (profilesData) {
-            const profiles = profilesData?.profiles.edges.map((p) => (p.node));
-
+        if (!loadingState) {
```
vkrizan, line 138: Observation: `loadingState` covers more and is not included within the effect's dependencies. 

How about `if (profiles)`? That should cover the loading state when the profiles are not yet loaded (`undefined`). And I think it will cover the case of empty array `[]` that would reset the `selectedRuleRefIds` state.

vkrizan (approved):
Works well. Thank you.

akofink:
> What if we don't set the selectedRuleRefIds˙ to []` but update it based on the systems selection, i.e. delete the ones that are coming from a minor OS version that is no longer in the list of selected systems?

This would be ideal but requires the benchmarks query from the rules tab. The system selection by itself isn't enough information to determine if selectedRuleRefIds needs to be changed. There's room for improvement here that we might want to tackle later.

src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
akofink: This doesn't allow setting rules, going back to step 3 to (un)assign systems, and moving forward to step 4 again. In this scenario, I expect rules already tailored to still be set. Adding `loadingState` to the dependencies doesn't seem to cause any issues, so I've added it to this hook.

PR #1124, opened 2021-04-13
fix(CreatePolicy): RHICOMPL-1621 reset anchor on a wizard step

vkrizan:
The Rules step sets an anchor when a tab is selected.  The anchor
persists between steps creating an issue when user deselects a system
and the tab gets removed.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management


src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -74,11 +81,13 @@ export const CreatePolicy = ({
             <Wizard
                 isOpen
                 width={ 1220 }
+                onNext={ onNext }
+                onGoToStep={ resetAnchor }
+                onBack={ resetAnchor }
```
skateman, line 86: Do we really need to reset the anchor on any step? Wouldn't be enough to do so if the selected systems change?

vkrizan: Sure, we can reset it only on a system change. I did not wanted to have it persistent between wizard steps.

skateman: I would only reset it if the active tab's affected system is removed. Otherwise it makes sense to remember the tab, but I'm not sure if that's doable...

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -16,9 +16,16 @@ export const CreatePolicy = ({
     benchmark, osMajorVersion, complianceThreshold, name, profile, refId, selectedRuleRefIds, systemIds
 }) => {
     const history = useHistory();
+    const location = useLocation();
```
akofink, line 19: I'm not sure I really understand react-router-dom, but `location` is provided by `useHistory`:

```suggestion
    const { push, location } = useHistory();
```

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -16,9 +16,16 @@ export const CreatePolicy = ({
...
     const [stepIdReached, setStepIdReached] = useState(1);
+
+    const resetAnchor = () => {
+        history.push({ state: location.state, hash: '' });
```
akofink, line 23: Are we sure `state` and `hash` are the only things required here? We could play it safer and spread everything in:

```suggestion
        history.push({ ...location, hash: '' });
```

akofink (approved):
Works as expected, fixing the bug! Thanks, @vkrizan!

I'm a bit confused why the achor isn't pre-set when visiting the rules step. We display the latest OS tab by default, why isn't the tag set in the URL?

src/SmartComponents/CreatePolicy/CreatePolicy.js
vkrizan: Make sense!

vkrizan:
> I'm a bit confused why the anchor isn't pre-set when visiting the rules step. We display the latest OS tab by default, why isn't the tag set in the URL?

It initially renders a the first tab as default w/o any changes to the anchor. I don't see much point of setting the anchor initially when the default is in effect.

src/SmartComponents/CreatePolicy/CreatePolicy.js
vkrizan: It would work up to the point if other routed tabs would come into play on other steps (if there were any). I'll think about it, and revisit this tomorrow.

vkrizan: This has to be reset on a policy type change of the within the same OS major version, as the system selection stays intact. Playing it safe to reset it every on every step. :wink:

PR #1131, opened 2021-04-14
feat(InventoryTable): Implement OS Major & Minor version filter

bastilian:
This adds a filter to the policy wizard and edit for the OS minor versions grouped by major version:

Screenshot 2021-04-15 at 14 17 22

**Note:** This requires https://github.com/RedHatInsights/frontend-components/pull/1073 to support the grouped filter

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.js
@@ -84,7 +84,8 @@ const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selec
             <Form>
                 <FormGroup>
                     <InvCmp
-                        prependComponent={prependComponent}
+                        showOsMinorVersionFilter={ [osMajorVersion] }
+                        prependComponent={ prependComponent }
                         emptyStateComponent={multiversionRules ? emptyStateComponent : undefined}
                         columns={columns}
                         remediationsEnabled={false}
--- a/src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
+++ b/src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
@@ -9,7 +9,7 @@ import { useHistory } from 'react-router-dom';
 const EditPolicySystemsTab = ({ osMajorVersion, policyOsMinorVersions }) => {
     const { push, location } = useHistory();
     const selectedSystemOsMinorVersions = useSelector(state => (
-        state?.entities?.selectedEntities?.map(entity => `${entity.osMinorVersion}`)
+        state?.entities?.selectedEntities?.map((entity) => (`${entity.osMinorVersion}`))
     ));
 
     const newOsMinorVersions = () => (
@@ -59,9 +59,10 @@ const EditPolicySystemsTab = ({ osMajorVersion, policyOsMinorVersions }) => {
     return (
         <React.Fragment>
             <InventoryTable
-                prependComponent={prependComponent}
-                emptyStateComponent={emptyStateComponent}
-                columns={columns}
+                showOsMinorVersionFilter={ [osMajorVersion] }
+                prependComponent={ prependComponent }
+                emptyStateComponent={ emptyStateComponent }
+                columns={ columns }
                 compact
                 showActions={ false }
                 query={ GET_SYSTEMS_WITHOUT_FAILED_RULES }
--- a/src/SmartComponents/SystemsTable/InventoryTable.js
+++ b/src/SmartComponents/SystemsTable/InventoryTable.js
@@ -22,6 +22,7 @@ import ComplianceRemediationButton from '@redhat-cloud-services/frontend-compone
 import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import useFilterConfig from 'Utilities/hooks/useFilterConfig';
 import { InventoryTable as FECInventoryTable } from '@redhat-cloud-services/frontend-components/Inventory';
+import { useOsMinorVersionFilter } from './hooks';
 
 export const InventoryTable = ({
     columns,
@@ -42,7 +43,8 @@ export const InventoryTable = ({
     systemProps,
     defaultFilter,
     emptyStateComponent,
-    prependComponent
+    prependComponent,
+    showOsMinorVersionFilter
 }) => {
     const store = useStore();
     const dispatch = useDispatch();
@@ -53,11 +55,14 @@ export const InventoryTable = ({
     });
     const [isLoaded, setIsLoaded] = useState(false);
     const [isEmpty, setIsEmpty] = useState(false);
+    const osMinorVersionFilter = useOsMinorVersionFilter(showOsMinorVersionFilter);
     const { conditionalFilter, activeFilters, buildFilterString } = useFilterConfig([
         ...DEFAULT_SYSTEMS_FILTER_CONFIGURATION,
         ...(compliantFilter ? COMPLIANT_SYSTEMS_FILTER_CONFIGURATION : []),
-        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : [])
+        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : []),
+        ...osMinorVersionFilter
     ]);
+
     const total = useSelector(({ entities }) => entities?.systemsCount) || 0;
     const items = useSelector(({ entities } = {}) => (entities?.systems?.map((system) => (
         system?.node?.id
@@ -219,7 +224,11 @@ InventoryTable.propTypes = {
         isFullView: PropTypes.bool
     }),
     emptyStateComponent: PropTypes.node,
-    prependComponent: PropTypes.node
+    prependComponent: PropTypes.node,
+    showOsMinorVersionFilter: PropTypes.oneOfType([
+        PropTypes.bool,
+        PropTypes.arrayOf(PropTypes.number)
+    ])
 };
 
 InventoryTable.defaultProps = {
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -1,0 +1,31 @@
+import useCollection from 'Utilities/hooks/api/useCollection';
+import { osMinorVersionFilter } from './constants';
+
+const groupByMajorVersion = (versions = [], showFilter) => {
+    const showVersion = (version) => {
+        if (showFilter.length > 0) {
+            return Array(showFilter).map(String).includes(String(version));
+        } else {
+            return true;
+        }
+    };
+
+    return versions.reduce((acc, currentValue) => {
+        if (showVersion(currentValue.osMajorVersion)) {
+            acc[String(currentValue.osMajorVersion)] = [...new Set(
+                [...acc[currentValue.osMajorVersion] || [], currentValue.osMinorVersion]
+            )];
+        }
+
+        return acc;
+    }, []);
+};
+
+export const useOsMinorVersionFilter = (showFilter) => {
+    const { data: supportedSsgs } = useCollection('supported_ssgs', {
+        type: 'supportedSsg',
+        skip: !showFilter
+    });
+
+    return osMinorVersionFilter(groupByMajorVersion(supportedSsgs?.collection, showFilter));
+};
--- a/src/constants.js
+++ b/src/constants.js
@@ -48,6 +48,36 @@ export const systemsOsFilterConfiguration = (policies) => ([{
     }))
 }]);
 
+const toSystemsOsMinorFilterConfigurationItem = (osVersions) => (
+    (majorVersion) => ({
+        label: `RHEL ${ majorVersion }`,
+        value: majorVersion,
+        items: osVersions[majorVersion].map((minorVersion) => ({
+            label: `RHEL ${ majorVersion }.${ minorVersion }`,
+            value: minorVersion
+        }))
+    })
+);
+
+export const systemsOsMinorFilterConfiguration = (osMajorVersions) => {
+    const filterString = (value) => ([
+        Object.keys(value).flatMap((majorVersion) => (
+            Object.keys(value[majorVersion]).map((minorVersion) => (
+                value[majorVersion][minorVersion] &&
+                    `(os_major_version = ${ majorVersion } AND os_minor_version = ${ minorVersion })`
+            ))
+        )).filter((v) => (!!v)).join(' OR ')
+    ]);
+    const items = Object.keys(osMajorVersions).map(toSystemsOsMinorFilterConfigurationItem(osMajorVersions));
+
+    return [{
+        type: conditionalFilterType.group,
+        label: 'Operating System',
+        filterString,
+        items
+    }];
+};
+
 export const COMPLIANT_SYSTEMS_FILTER_CONFIGURATION = [
     {
         type: conditionalFilterType.checkbox,
--- a/package-lock.json
+++ b/package-lock.json
@@ -2239,9 +2239,9 @@
       }
     },
     "@redhat-cloud-services/frontend-components-inventory-compliance": {
-      "version": "3.2.3",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components-inventory-compliance/-/frontend-components-inventory-compliance-3.2.3.tgz",
-      "integrity": "sha512-omQvPI4UA6tlcKX+58nKg3eoQmOtWZQg4lFz8oGIRQPhTLJxE771sq88K60I5vj4b/NMgG4ZP7ACqfYC1U2D3Q==",
+      "version": "3.2.4",
+      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components-inventory-compliance/-/frontend-components-inventory-compliance-3.2.4.tgz",
+      "integrity": "sha512-+h2sTLw47xw/sYzeET6uPUeQCFutD9nUROpd6l7ZpODBIV6PPgrQ9dQ42WVij4VINRVwSTGXCbTeSXwkgo7Cvg==",
       "requires": {
         "@apollo/react-hooks": "^3.1.5",
         "@redhat-cloud-services/frontend-components": ">=3.0.0",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@patternfly/react-table": "^4.23.14",
     "@patternfly/react-tokens": "^4.10.5",
     "@redhat-cloud-services/frontend-components": "3.0.10",
-    "@redhat-cloud-services/frontend-components-inventory-compliance": "^3.2.3",
+    "@redhat-cloud-services/frontend-components-inventory-compliance": "^3.2.4",
     "@redhat-cloud-services/frontend-components-notifications": "^3.1.0",
     "@redhat-cloud-services/frontend-components-utilities": "^3.0.3",
     "@redhat-cloud-services/host-inventory-client": "^1.0.89",
--- a/src/SmartComponents/CreatePolicy/__snapshots__/EditPolicyDetails.test.js.snap
+++ b/src/SmartComponents/CreatePolicy/__snapshots__/EditPolicyDetails.test.js.snap
@@ -35,6 +35,12 @@ exports[`EditPolicyDetails expect to render without error 1`] = `
   <EditPolicyDetails
     policy={
       Object {
+        "benchmark": Object {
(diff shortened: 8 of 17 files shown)


src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -84,7 +86,9 @@ const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selec
             <Form>
                 <FormGroup>
                     <InvCmp
-                        prependComponent={prependComponent}
+                        osMajorVersion={ osMajorVersion }
+                        osMinorVersions={ latestSupportedOsMinorVersions }
```
vkrizan, line 90: Why are we passing OS minor versions from the benchmarks and not reading the versions that are coming from the host list? The Inventory component can read what OS versions it gets and use that for the filter. If we go with server side pagination, such information can be read from the metadata.

src/constants.js
```diff
@@ -48,6 +48,24 @@ export const systemsOsFilterConfiguration = (policies) => ([{
     }))
 }]);
 
+export const systemsOsMinorFilterConfiguration = (osMajorVersion, osMinorVersions) => ([{
+    type: conditionalFilterType.group,
+    label: 'Operating System',
+    filterString: (value) => ([Object.keys(value).map((majorVersion) => (
```
vkrizan, line 54: Observation: This can be extracted to a separate function to reduce the complexity.

src/constants.js
```diff
@@ -48,6 +48,24 @@ export const systemsOsFilterConfiguration = (policies) => ([{
     }))
 }]);
 
+export const systemsOsMinorFilterConfiguration = (osMajorVersion, osMinorVersions) => ([{
```
vkrizan, line 51: This passes major and minor versions separately. This would not work for general cases, like listing all systems (on Systems page). It is too specific for the policy creation and edit. Could this be more robust and filter by OS version (as whole)?

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -56,8 +58,10 @@ export const InventoryTable = ({
     const { conditionalFilter, activeFilters, buildFilterString } = useFilterConfig([
         ...DEFAULT_SYSTEMS_FILTER_CONFIGURATION,
         ...(compliantFilter ? COMPLIANT_SYSTEMS_FILTER_CONFIGURATION : []),
-        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : [])
+        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : []),
+        ...(osMajorVersion && osMinorVersions?.length > 0 ? osMinorVersionFilter(osMajorVersion, osMinorVersions) : [])
```
vkrizan, line 62: I'm getting `Uncaught TypeError: e is undefined` and an infinite loop on the wizard. I have two systems: one 7.9 and one 7.7.

vkrizan (commented):
The implementation is too tied to the benchmarks with their latest supported version and not the the systems data. It is a bit unfortunate as the filter cannot be used on Systems page.

bastilian:
@vkrizan There is a chicken/egg problem. We can not filter systems data by their data yet. We can dynamically add filters, so using systems data loaded from the backend can then be used to construct them, but again, only with data loaded, and there is currently no other way to get some kind of meta data of a systems list. If we would have such a feature in our API then we could tie the filters to data from the systems it shows, but for now we will need to construct them from the benchmarks.

src/constants.js
bastilian: Yes, we could, of course. But for now as a first iteration of RHICOMPL-1476 to just fix RHICOMPL-1655 this should work.

We would just need to allow passing in a similar format that the `ConditionalFilter` component already uses for grouped filters.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
bastilian: > If we go with server side pagination, such information can be read from the metadata.

Do we have this metadata available (now)?

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: Hm. Is there more to the error? I have two hosts with different minor versions and same major version as well and can not reproduce any issue with an infinite loop, also not when filtering?

akofink:
Could we please put the jira issue ID in the commit?

src/constants.js
```diff
@@ -48,6 +48,24 @@ export const systemsOsFilterConfiguration = (policies) => ([{
...
+    type: conditionalFilterType.group,
+    label: 'Operating System',
+    filterString: (value) => ([Object.keys(value).map((majorVersion) => (
+        Object.keys(value[majorVersion]).map((minorVersion) => (
+            `os_major_version = ${ majorVersion } AND os_minor_version = ${ minorVersion }`
+        ))
+    )).flat().join(' OR ')]),
+    items: [osMajorVersion].map((majorVersion) => ({
+        label: `RHEL ${ majorVersion }.0`,
```
akofink, line 60: It doesn't really make sense to show RHEL 7.0 and then each RHEL 7.Y version under that.

```suggestion
        label: `RHEL ${ majorVersion }`,
```

akofink:
Removing a filter using the filter dropdown doesn't seem to work. The chips show what the filter actually is, but it should have removed 7.7:

image

akofink:
Could we also add this to the policy details -> systems table, since we now show Operating system there?

akofink:
This filter does not work at all for policies without OS version tabs. This is essentially all policies today in production and represents the first time a user opens the new edit policy modal after we deploy 133. To reproduce:

1. Create a policy with some systems, ideally two different OS versions (but not required)
2. Remove the OS minor versions using the Rails console: `policy.profiles.update_all(os_minor_version: '')`
3. Open the edit modal for that policy

image

I agree with Viliam that this filter should not need OS versions passed in. We need to use the response from the API to build the content of this filter. We do this in the rules table with `updatePolicyFilterConfig` for the policy filter in `componentDidMount` (obviously, could useEffect).

akofink:
This doesn't work properly in the policy create wizard:

image

bastilian:
> I agree with Viliam that this filter should not need OS versions passed in. We need to use the response from the API to build the content of this filter. We do this in the rules table with updatePolicyFilterConfig for the policy filter in componentDidMount (obviously, could useEffect).

With `updatePolicyFilterConfig` this works, because all rules are loaded, for the systems table the same is not the case and we would only be able to build filters based on the data we have requested from the API and unless we have a way to "summaries" certain details of **all** systems in a view we can not use that data to build the filters.

bastilian:
> This doesn't work properly in the policy create wizard

@akofink Can you provide steps to reproduce this? I'm not really sure how to get to that. What Policy and profile is this?

akofink:
> Can you provide steps to reproduce this?

1. Have some systems that include non-latest OS version (RHEL < 7.9)
2. Begin to create any new policy

Expected: all OS versions from systems selected show as a filter
Actual: Only the latest OS version shows (i.e. RHEL 7.9)

akofink:
> With updatePolicyFilterConfig this works, because all rules are loaded, for the systems table the same is not the case and we would only be able to build filters based on the data we have requested from the API and unless we have a way to "summaries" certain details of all systems in a view (meaning not only those on the current page, but all in an account/profile/etc) we can not use that data to build the filters.

Yes, I understand this issue and how it's very different than the rules table. I'm not really sure of a good way to solve it, tbh, since the API needs to be able to return all OS versions for some filter. Using some other unrelated query result and passing it in, though, seems less than ideal and definitely not portable - whatever logic or extra queries need to be made for this filter should be self-contained within the InventoryTable.

bastilian:
@akofink We do something similar on the `ComplianceSystems` component, where we request profiles in order to build the profiles filter.

akofink:
Hmm, true :thinking: ok :laughing: it does match that filter fairly well. We definitely cannot rely on `latestSupportedOsMinorVersions`, though. We need the entire set of OS versions in our DB, not just those supported on the latest SSG versions. I don't think that information is anywhere in any of our existing queries.

I still think it'd be more ideal to have this self-contained within the InventoryTable. Perhaps we could simply show all OS versions for now, even those that don't have any systems. We could use the supported_ssgs API endpoint, perhaps.

bastilian:
@akofink @vkrizan It now uses the OS version provided by the `supported_ssgs` endpoint, but I think it would make sense to limit them at least by major version until we have a more accurate way to request a list of operating systems to filter.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
```diff
@@ -9,7 +9,7 @@ import { useHistory } from 'react-router-dom';
 const EditPolicySystemsTab = ({ osMajorVersion, policyOsMinorVersions }) => {
     const { push, location } = useHistory();
     const selectedSystemOsMinorVersions = useSelector(state => (
-        state?.entities?.selectedEntities?.map(entity => `${entity.osMinorVersion}`)
+        state?.entities?.selectedEntities?.map((entity) => (entity.osMinorVersion))
```
vkrizan, line 12: Observation: `osMinorVersion` coming from systems is an Integer. Previously it was converted to string here.

What's the intent of this change? Could this break in some other places where this information is used?
Note that keys of an Object are always strings.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -84,7 +84,8 @@ const EditPolicySystems = ({ change, osMajorVersion, osMinorVersionCounts, selec
             <Form>
                 <FormGroup>
                     <InvCmp
-                        prependComponent={prependComponent}
+                        showOsMinorVersionFilter
```
vkrizan, line 87: > I think it would make sense to limit them at least by major version until we have a more accurate way to request a list of operating systems to filter.

We can pass the configuration of the filter here.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -0,0 +1,19 @@
...
+        return acc;
+    }, [])
+);
+
+export const useOsMinorVersionFilter = (showFilter) => {
+    const { data: supportedSsgs } = useCollection('supported_ssgs', {
+        type: 'supportedSsg',
+        skip: !showFilter
+    });
+    return osMinorVersionFilter(groupByMajorVersion(supportedSsgs?.collection));
```
vkrizan, line 18: `groupByMajorVersion` does not seems to check for `undefined`.

vkrizan:
The infinite loop is still happening in wizard as well as in Policy Details - Systems. 

Here is the traceback:
```
Uncaught TypeError: e is undefined
    T index.js:2
    dropdownItems index.js:2
    value index.js:2
    React 12
    unstable_runWithPriority scheduler.development.js:646
    React 5
    t index.js:2
   ...

The above error occurred in the <n> component:

n@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.57dac3b7d578085a88a2.js:1:198798
div
ToolbarItem@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Toolbar/ToolbarItem.js:31:105
div
ToolbarGroupWithRef@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Toolbar/ToolbarGroup.js:23:1
div
div
ToolbarContent@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Toolbar/ToolbarContent.js:22:9
div
Toolbar@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Toolbar/Toolbar.js:24:9
n@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.57dac3b7d578085a88a2.js:1:212433
S@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.57dac3b7d578085a88a2.js:1:290199
2759/e.exports</h<@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.57dac3b7d578085a88a2.js:1:287434
l@https://ci.foo.redhat.com:1337/apps/chrome/js/vendors~inventory.57dac3b7d578085a88a2.js:1:277365
AsyncInventory@webpack-internal:///./node_modules/@redhat-cloud-services/frontend-components/esm/Inventory/AsyncInventory.js:40:18
BaseScalprumComponent@webpack-internal:///./node_modules/@scalprum/react-core/dist/esm/scalprum-component.js:115:28
Suspense
BaseInvTable@webpack-internal:///./node_modules/@redhat-cloud-services/frontend-components/esm/Inventory/InventoryTable.js:32:77
StateViewPart@webpack-internal:///./src/PresentationalComponents/StateView/StateView.js:16:18
StateView@webpack-internal:///./src/PresentationalComponents/StateView/StateView.js:24:21
InventoryTable@webpack-internal:///./src/SmartComponents/SystemsTable/InventoryTable.js:71:17
ApolloConsumer
WithApollo@webpack-internal:///./node_modules/@apollo/client/react/hoc/withApollo.js:26:32
PolicySystemsTab@webpack-internal:///./src/SmartComponents/PolicyDetails/PolicySystemsTab.js:36:16
ContentTab@webpack-internal:///./src/PresentationalComponents/TabSwitcher/TabSwitcher.js:29:18
RoutedTabSwitcher@webpack-internal:///./src/PresentationalComponents/TabSwitcher/TabSwitcher.js:56:18
section
Main@webpack-internal:///./node_modules/@redhat-cloud-services/frontend-components/esm/Main/Main.js:73:81
ConnectFunction@webpack-internal:///./node_modules/react-redux/es/components/connectAdvanced.js:232:68
StateViewPart@webpack-internal:///./src/PresentationalComponents/StateView/StateView.js:16:18
StateView@webpack-internal:///./src/PresentationalComponents/StateView/StateView.js:24:21
StateViewWithError@webpack-internal:///./src/PresentationalComponents/StateView/StateView.js:34:21
PolicyDetails@webpack-internal:///./src/SmartComponents/PolicyDetails/PolicyDetails.js:60:15
Route@webpack-internal:///./node_modules/react-router/esm/react-router.js:465:29
Switch@webpack-internal:///./node_modules/react-router/esm/react-router.js:671:29
Suspense
Router@webpack-internal:///./src/Utilities/Router.js:71:16
Routes
App@webpack-internal:///./src/App.js:43:81
RouterParams@webpack-internal:///./node_modules/@redhat-cloud-services/frontend-components-utilities/esm/RouterParams/RouterParams.js:48:83
ConnectFunction@webpack-internal:///./node_modules/react-redux/es/components/connectAdvanced.js:232:68
C@webpack-internal:///./node_modules/react-router/esm/react-router.js:726:31
ApolloProvider@webpack-internal:///./node_modules/@apollo/client/react/context/ApolloProvider.js:13:18
Router@webpack-internal:///./node_modules/react-router/esm/react-router.js:94:30
BrowserRouter@webpack-internal:///./node_modules/react-router-dom/esm/react-router-dom.js:60:35
IntlProvider@webpack-internal:///./node_modules/react-intl/lib/src/components/provider.js:93:47
Provider@webpack-internal:///./node_modules/react-redux/es/components/Provider.js:17:15
AppEntry@webpack-internal:///./src/AppEntry.js:42:16

React will try to recreate this component tree from scratch using the error boundary you provided, BaseScalprumComponent.
```

vkrizan (commented):
The use of supported SSGs seems like a better option... 
I'm still not able to test this as every page has the infinite loop (also in Chromium).

vkrizan:
> do you have the frontend-components changes checkout and linked?

Oh, missed that. Sorry.

src/SmartComponents/SystemsTable/hooks.js
bastilian: Because it has a default for the versions argument.

src/constants.js
```diff
@@ -48,6 +48,36 @@ export const systemsOsFilterConfiguration = (policies) => ([{
...
+export const systemsOsMinorFilterConfiguration = (osMajorVersions) => {
+    const filterString = (value) => ([
+        Object.keys(value).flatMap((majorVersion) => (
+            Object.keys(value[majorVersion]).map((minorVersion) => (
+                value[majorVersion][minorVersion] &&
+                    `os_major_version = ${ majorVersion } AND os_minor_version = ${ minorVersion }`
+            ))
+        )).filter((v) => (!!v)).join(' OR ')
+    ]);
+    const items = Object.keys(osMajorVersions).map(toSystemsOsMinorFilterConfigurationItem(osMajorVersions));
```
vkrizan, line 71: There are too many items that it overflows. Limiting it to one major version could mitigate this. Can the filter options have a max height and show scrollbar when it overflows?

Notice the scrollbar:
Screenshot_2021-04-16 C2S for Red Hat Enterprise Linux 7 - SCAP policies - Compliance Red Hat Insights

src/constants.js
```diff
@@ -48,6 +48,36 @@ export const systemsOsFilterConfiguration = (policies) => ([{
...
+        }))
+    })
+);
+
+export const systemsOsMinorFilterConfiguration = (osMajorVersions) => {
+    const filterString = (value) => ([
+        Object.keys(value).flatMap((majorVersion) => (
+            Object.keys(value[majorVersion]).map((minorVersion) => (
+                value[majorVersion][minorVersion] &&
+                    `os_major_version = ${ majorVersion } AND os_minor_version = ${ minorVersion }`
```
vkrizan, line 67: The filter does not work when combing different major versions. For example RHEL 6.8 and 7.9.

Here is the `search` sent to the request:
```
os_major_version = 6 AND os_minor_version = 8 OR os_major_version = 7 AND os_minor_version = 9
```

Try wrapping each group with parenthesis.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
bastilian: @vkrizan I reverted it. It didn't appear to be necessary to convert it and it kinda is.

`selectedSystemOsMinorVersions` is only used below to check for any new item compared to the policy versions passed in, regardless of what it is or what type.

vkrizan: I guess this leads to always showing the alert "You selected a system that has a release version previously not included in this policy.".

src/SmartComponents/CreatePolicy/EditPolicySystems.js
bastilian: @vkrizan You mean to limit the versions?

src/constants.js
bastilian: Yeah, it is a big list. We should at least for now limit it to a major version.

Regarding the issue with them being hidden, i'll investigate and maybe try to see if anyone in the patternfly slack has an answer.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -53,11 +55,14 @@ export const InventoryTable = ({
     });
     const [isLoaded, setIsLoaded] = useState(false);
     const [isEmpty, setIsEmpty] = useState(false);
+    const osMinorVersionFilter = useOsMinorVersionFilter(showOsMinorVersionFilter);
     const { conditionalFilter, activeFilters, buildFilterString } = useFilterConfig([
         ...DEFAULT_SYSTEMS_FILTER_CONFIGURATION,
         ...(compliantFilter ? COMPLIANT_SYSTEMS_FILTER_CONFIGURATION : []),
-        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : [])
+        ...(policies?.length > 0 ? policyFilter(policies, showOsFilter) : []),
+        ...osMinorVersionFilter
```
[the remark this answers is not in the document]
bastilian: Nice catch! This was an issue in the frontend components part. It has been fixed now and filters are now properly removed from group filters via the chip.

src/constants.js
bastilian: @vkrizan Each version combination is now in parenthesis, so this should work now with different major OS versions.

bastilian:
@akofink @vkrizan It is now changed to only show minor OS versions matching the appropriate major version.

akofink (approved):
It seems to work as expected! There is a bit of strange behavior regarding the overflow due to the filter being so long, but I don't think we should block this work on that issue. Thanks, @bastilian, for tackling this tricky feature!

PR #1134, opened 2021-04-14
fix(EditPolicy): RHICOMPL-1651 rules loading states 

vkrizan:
* fixes showing spinner in the edit modal early, especially if the policy does not have tabs configured yet
  * with added `skip` option to `useCollection`
* fixes `useCollection` to reset to default on effects variable change, which fixes showing spinner right after user selects a system from a new OS minor group
* fixes showing spinner only if the new data are being still being loaded (old were shown right under the spinner)

Note, I haven't experienced any issues with the Create Policy Wizard regarding a missing loading state.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
+++ b/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
@@ -112,8 +112,8 @@ const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, osMinorV
         },
         skip: filter.length === 0
     });
-    const dataState = ((tabsData?.length > 0) ? profilesData : undefined);
     const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const dataState = ((!loadingState && tabsData?.length > 0) ? profilesData : undefined);
 
     useLayoutEffect(() => {
         if (profilesData) {
--- a/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
+++ b/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
@@ -59,13 +59,15 @@ const getBenchmarkProfile = (benchmark, profileRefId) => (
 
 const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, osMinorVersionCounts }) => {
     const osMajorVersion = policy?.osMajorVersion;
+    const osMinorVersions = Object.keys(osMinorVersionCounts).sort();
     const benchmarkSearch = `os_major_version = ${ osMajorVersion } ` +
-        `and latest_supported_os_minor_version ^ (${ Object.keys(osMinorVersionCounts).sort().join(',') })`;
+        `and latest_supported_os_minor_version ^ (${ osMinorVersions.join(',') })`;
 
     const { data: benchmarks, loading: benchmarksLoading } = useCollection('benchmarks', {
         type: 'benchmark',
         include: ['profiles'],
-        params: { search: benchmarkSearch }
+        params: { search: benchmarkSearch },
+        skip: osMinorVersions.length === 0
     }, [benchmarkSearch]);
 
     let profileIds = [];
--- a/src/Utilities/hooks/api/useCollection.js
+++ b/src/Utilities/hooks/api/useCollection.js
@@ -60,10 +60,16 @@ const useCollection = (collection, options = {}, effects = []) => {
     };
 
     useEffect(() => {
+        setCollectionState({
+            data: undefined,
+            loading: true,
+            error: undefined
+        });
+
         fetchCollection(apiClient, collection, params, options).then((data) => {
             setCollectionState({
                 data,
-                loading: undefined,
+                loading: false,
                 error: undefined
             });
         });
--- a/src/Utilities/hooks/api/useCollection.js
+++ b/src/Utilities/hooks/api/useCollection.js
@@ -48,7 +48,7 @@ const fetchCollection = async (apiClient, collection, params = {}, options = {})
 const useCollection = (collection, options = {}, effects = []) => {
     const [collectionState, setCollectionState] = useState({
         data: undefined,
-        loading: true,
+        loading: !options?.skip,
         error: undefined
     });
     const apiClient = useApi({
@@ -60,6 +60,16 @@ const useCollection = (collection, options = {}, effects = []) => {
     };
 
     useEffect(() => {
+        if (options?.skip) {
+            setCollectionState({
+                data: undefined,
+                loading: false,
+                error: undefined
+            });
+
+            return;
+        }
+
         setCollectionState({
             data: undefined,
             loading: true,
@@ -73,7 +83,7 @@ const useCollection = (collection, options = {}, effects = []) => {
                 error: undefined
             });
         });
-    }, effects);
+    }, [...effects, options?.skip]);
 
     return collectionState;
 };


src/Utilities/hooks/api/useCollection.js
```diff
@@ -60,10 +60,16 @@ const useCollection = (collection, options = {}, effects = []) => {
     };
 
     useEffect(() => {
+        setCollectionState({
```
vkrizan, line 63: @bastilian do you happen to know how to test to trigger the useEffect?

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -110,8 +110,8 @@ const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, osMinorV
         },
         skip: filter.length === 0
     });
-    const dataState = ((tabsData?.length > 0) ? profilesData : undefined);
-    const loadingState = ((loading || benchmarksLoading) ? true : undefined);
+    const loadingState = ((loading || benchmarksLoading || tabsData.length === 0) ? true : undefined);
```
akofink, line 113: tabsData.length can be 0 even after loading has completed for both API requests.

src/Utilities/hooks/api/useCollection.js
```diff
@@ -60,10 +60,16 @@ const useCollection = (collection, options = {}, effects = []) => {
...
+            data: undefined,
+            loading: true,
+            error: undefined
+        });
+
         fetchCollection(apiClient, collection, params, options).then((data) => {
             setCollectionState({
                 data,
-                loading: undefined,
+                loading: false,
```
bastilian, line 72: This should be `undefined`. useCollection mimics `useQuery`, which will return undefined rather than false.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
vkrizan: Yeah, that would be a corner case where it would show the loader indefinitely. It should not happen with the new policies where there has to be at least one OS minor tab configured.

Do you have an idea what else to put there to fix the situation?

src/Utilities/hooks/api/useCollection.js
vkrizan: `useQuery` did returned `false` for me. See [Apollo docs](https://www.apollographql.com/docs/react/api/react/hooks/#result) where it states boolean. We should avoid using `undefined` explicitly where a `false` is more appropriate.

bastilian: @vkrizan All components that use `useQuery` (or in a few instances with `useCollection`) are built on there only one of `data`, `loading` or `error` being defined, and when one of these is defined it is the type stated in the Apollo docs.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
vkrizan: We have these cases:
* no systems, no OS profile
* some systems, no canonical profile found for them

The later is what I was labeling as a corner case.

src/Utilities/hooks/api/useCollection.js
vkrizan: Here is the result signature:
```
export interface QueryResult<TData = any, TVariables = OperationVariables>
  extends ObservableQueryFields<TData, TVariables> {
  client: ApolloClient<any>;
  data: TData | undefined;
  previousData?: TData;
  error?: ApolloError;
  loading: boolean;
  networkStatus: NetworkStatus;
  called: true;
}
```

Only `data`, `previousData` and `error` are optional (can be undefined).

bastilian: @vkrizan Oh, ok. In the past we definitely had loading `undefined`, otherwise we would not have changed all instances where we use it.

If we change this convention we then (eventually) should also update other places and make sure loading is a boolean.

src/Utilities/hooks/api/useCollection.js
bastilian: Why is it needed to set the initial state again?

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
vkrizan: Okay, I've restructured it differently. The tabsData are not in the play here.

Instead, I've added a skip option to `useCollection`. Please review and test. Thank you.

src/Utilities/hooks/api/useCollection.js
```diff
@@ -60,14 +60,30 @@ const useCollection = (collection, options = {}, effects = []) => {
...
 
     useEffect(() => {
+        if (options?.skip) {
+            setCollectionState({
+                data: undefined,
+                loading: false,
+                error: undefined
+            });
+
+            return;
```
bastilian, line 70: Shouldn't an effect only return with a cleanup function and otherwise not at all?

akofink: > and otherwise not at all

All functions return. Implicit return is the same as this; both return `undefined`. This is just a way to exit early without a large `else` condition wrapping the rest of the function.

src/Utilities/hooks/api/useCollection.js
```diff
@@ -48,7 +48,7 @@ const fetchCollection = async (apiClient, collection, params = {}, options = {})
 const useCollection = (collection, options = {}, effects = []) => {
     const [collectionState, setCollectionState] = useState({
         data: undefined,
-        loading: true,
+        loading: !options?.skip,
```
akofink, line 51: Can we write a test for this (or report a jira to add tests here)?

akofink (approved):
Seems to work as expected! Thanks, @vkrizan!

src/Utilities/hooks/api/useCollection.js
bastilian: I'm a bit confused why the skip option dictates the loading state.

bastilian: Or is this meant to be "skip loading state"?

vkrizan: Sure, I can try to figure one out.

akofink: It means "skip the request" - so loading state doesn't make sense in that case. We're never loading if there's no request or response to wait for.

src/Utilities/hooks/api/useCollection.js
```diff
@@ -48,7 +48,7 @@ const fetchCollection = async (apiClient, collection, params = {}, options = {})
 const useCollection = (collection, options = {}, effects = []) => {
```
bastilian, line 48: How about something like this:

```suggestion
const useCollection = (collection, options = {}, effects = []) => {
    const [collectionState, setCollectionState] = useState({
        data: undefined,
        loading: true,
        error: undefined
    });
    const apiClient = useApi({
        apiBase: COMPLIANCE_API_ROOT
    });
    const params = {
        ...(options?.params || {}),
        include: (options?.include || [])
    };

    useEffect(() => {
        Promise.resolve(setCollectionState({
            data: undefined,
            loading: true,
            error: undefined
        })).then(() => {
            fetchCollection(apiClient, collection, params, options).then((data) => {
                setCollectionState({
                    data,
                    loading: false,
                    error: undefined
                });
            });
        });
    }, effects);

    return collectionState;
};
```

Having the initial loading state `false` feels more appropriate as data is not being requested yet.
The above also reflects the transitions between states nicely.
I think the second `setCollectionState` call might not be needed to be wrapped into a promise, but it should make things flow nicer. I'm not sure how the `skip` option would play into this.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -59,13 +59,15 @@ const getBenchmarkProfile = (benchmark, profileRefId) => (
...
     const benchmarkSearch = `os_major_version = ${ osMajorVersion } ` +
-        `and latest_supported_os_minor_version ^ (${ Object.keys(osMinorVersionCounts).sort().join(',') })`;
+        `and latest_supported_os_minor_version ^ (${ osMinorVersions.join(',') })`;
 
     const { data: benchmarks, loading: benchmarksLoading } = useCollection('benchmarks', {
         type: 'benchmark',
         include: ['profiles'],
-        params: { search: benchmarkSearch }
+        params: { search: benchmarkSearch },
+        skip: osMinorVersions.length === 0
```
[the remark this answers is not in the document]
vkrizan: I guess to have this consistent with the behavior with Apollo's `useQuery` the data should be `undefined` if skipped, as there are none. I'll double check if that's happening correctly with the introduction of the test.

The problem here is deeper. The `skip` mitigates it and relies on the user not changing OS minor versions too quickly whilst switching back and forth from Rules and Systems tab. The issue is race condition between request.

This is how it behaved w/o the skip:
```
* benchmarkSearch is changed and first request is made
* loading is true
* benchmarkSearch is changed and second request is made
* first request finishes and sets the loading to false    <--- here is where the spinner disappeared
* second request finishes
```
I'm not sure if this is handled within Apollo, but I think so it is. It needs canceling of callbacks of previous request(s).

src/Utilities/hooks/api/useCollection.js
vkrizan: This could help with a case when race condition between setting the first state and the second one. I doubt that it would happen especially if the setStates are put into queue. It definitely helps with the flow. Let me try something out. Is there any other benefit that I've missed?

I'm more worried about the case I've described at https://github.com/RedHatInsights/compliance-frontend/pull/1134#discussion_r614185234.

vkrizan: `Promise.resolve(setCollectionState())` does not wait for the set state, as the setter is async AFAIK.

src/Utilities/hooks/api/useCollection.js
vkrizan: There are as many requests as there are changes to the dependencies for the `useEffect` (from the `effects` variable).

src/Utilities/hooks/api/useCollection.js
vkrizan: The tests for some reason do not work within jest for some reason. I'm getting `Invalid hook call.` on the first `useState`. The existing test that is there for `useCollection` is currently being skipped. Lets revisit this issue later. I'll find or file a new Jira.

bastilian: We should add `renderHook` from the react testing library that can help testing hooks. See here for an example: https://github.com/RedHatInsights/frontend-components/pull/919/files#diff-ad9471069e8109b82757d9ae0e14abab2d4aa8e4461b03278eff01ffbb3c5919

src/Utilities/hooks/api/useCollection.js
vkrizan: `undefined` is functionally the same as `false`. Updating all old occurrences is not necessary (cost to benefit).

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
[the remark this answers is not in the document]
akofink: Wait, so you're saying if the API returns no results (`data: []`) then `benchmarks` is `undefined`? I don't think so... it'd be `[]` in that case, right?

src/Utilities/hooks/api/useCollection.js
bastilian: @vkrizan That doesn't explain why the initial state needs to be set again. It's not obvious to me why this would need to happen.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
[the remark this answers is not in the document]
vkrizan: > it'd be [] in that case, right?
Of course, if API returns empty collection the result is an empty array.

The `undefined` is only when there is no request being made when `skip` is in effect.

src/Utilities/hooks/api/useCollection.js
vkrizan: Assume there is no skip... The first call to `useCollection` sets the initial `collectionState` with `loading` true. Request finishes and sets the data and `loading` to false. Then there is a change within dependencies from `effects` that triggers `useEffect` to be run second time. If it wouldn't reset the `loading` to true then there would be no loading information between first and second request.

bastilian: @vkrizan The initial loading can either be true or false due to the skip option, or not?

bastilian: Why not something like this: 

```js
const useCollection = (collection, options = {}, dependencies = []) => {
    const [collectionState, setCollectionState] = useState({
        data: undefined,
        loading: false,
        error: undefined
    });
    const apiClient = useApi({
        apiBase: COMPLIANCE_API_ROOT
    });
    const params = {
        ...(options?.params || {}),
        include: (options?.include || [])
    };

    useEffect(() => {
        if (!options?.skip) {
            setCollectionState({
                data: undefined,
                loading: true,
                error: undefined
            });

            fetchCollection(apiClient, collection, params, options).then((data) => {
                setCollectionState({
                    data,
                    loading: false,
                    error: undefined
                });
            });
        }
    }, [...dependencies, options?.skip]);

    return collectionState;
};
```

Here it is more obvious what all states are, and what the skip option does.

vkrizan: It does one less state setting, but has one caveat that when it transitions from skip false to skip true and the request is pending it does not transition to loading false immediately.

PR #1142, opened 2021-04-16
fix(EditPolicySystemsTab): RHICOMPL-1697 new rules alert

akofink:
The new rules alert should only show when new tabs appear. This PR fixes
an issue where the alert appeared for new systems, despite not having
a valid policy type due to the OS version. This solution highlights the
performance issue with the rules tab requesting all benchmark rules.
This request is required to determine if the newly added system OS
versions have this policy type, so the alert is quite sluggish, on the
order of 1-2 seconds, to update when new systems are selected. I'm
hopeful we'll optimize the rules tab queries, which will also fix this
sluggish alert.

I've reported this API sluggishness as RHICOMPL-1706.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
+++ b/src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
@@ -120,7 +120,7 @@ export const toTabsData = (policy, osMinorVersionCounts, benchmarks, selectedRul
     }).filter(({ profile, newOsMinorVersion }) => !!profile && newOsMinorVersion)
 );
 
-export const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, osMinorVersionCounts }) => {
+export const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, osMinorVersionCounts, setNewRuleTabs }) => {
     const osMajorVersion = policy?.osMajorVersion;
     const osMinorVersions = Object.keys(osMinorVersionCounts).sort();
     const benchmarkSearch = `os_major_version = ${ osMajorVersion } ` +
@@ -152,6 +152,14 @@ export const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, o
     });
     const loadingState = ((profilesLoading || benchmarksLoading) ? true : undefined);
     const dataState = ((!loadingState && tabsData?.length > 0) ? profilesData : undefined);
+
+    if (!loadingState) {
+        setNewRuleTabs(!!tabsData.find(tab => (
+            policy.policy.profiles.find(profile => (
+                profile.osMinorVersion !== tab.newOsMinorVersion
+            ))
+        )));
+    }
 
     useLayoutEffect(() => {
         if (profilesData) {
@@ -198,6 +206,7 @@ export const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, o
 
 EditPolicyRulesTab.propTypes = {
     handleSelect: propTypes.func,
+    setNewRuleTabs: propTypes.func,
     policy: propTypes.object,
     osMinorVersionCounts: propTypes.shape({
         osMinorVersion: propTypes.shape({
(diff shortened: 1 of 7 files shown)


vkrizan (approved):
Works. 

The alert delay is terrible :smile:. Are we sure that we want this to be fixed. I'd accept this caveat of showing the warning even if there is no tab to edit. It is smaller price to pay over the delay. Even with GraphQL there would be some delay.

akofink:
I think https://github.com/RedHatInsights/compliance-frontend/pull/1151 will solve our problems on this sluggish alert, though I haven't tested them together yet to confirm.

akofink:
Updated after the big feature flag removal/refactor.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -89,7 +89,8 @@ export const toTabsData = (policy, osMinorVersionCounts, benchmarks, selectedRul
                         ...benchmarkProfile,
                         benchmark: benchmarkProfile.relationships?.benchmark?.data,
                         rules: benchmarkProfile.relationships?.rules?.data,
-                        ...profile
+                        ...profile,
+                        newTab: true
```
akofink, line 93: @bastilian moved the tabsData building to this function, so I had to get kind of creative about how to pass this information back to the calling scope.

src/SmartComponents/EditPolicy/EditPolicySystemsTab.test.js
```diff
@@ -41,4 +41,11 @@ describe('EditPolicySystemsTab', () => {
         );
         expect(toJson(wrapper)).toMatchSnapshot();
     });
+
+    it('expect to render with new tabs alert', async () => {
```
akofink, line 45: Added a test that renders the alert as well.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
bastilian: @akofink It is a bit odd. Could we pass in the updatedPolicy and compare the profiles count there to the number of "tab profiles" are returned rather than find this prop for `setNewRuleTabs`?

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -153,6 +153,14 @@ export const EditPolicyRulesTab = ({ handleSelect, policy, selectedRuleRefIds, o
     const loadingState = ((profilesLoading || benchmarksLoading) ? true : undefined);
     const dataState = ((!loadingState && tabsData?.length > 0) ? profilesData : undefined);
 
+    if (!loadingState) {
+        setNewRuleTabs(!!tabsData.find(tab => (
+            policy.policy.profiles.find(profile => (
+                profile.osMinorVersion !== tab.newOsMinorVersion
+            ))
+        )));
+    }
```
akofink, line 162: Ok, moved from using `newTab` within `tabsData` to detecting it by comparing `tabsData` to `policy.policy.profiles`. Sorry for the delay!

bastilian: Hm. Won't this evaluate to `true` all the time?

The finds return an array and regardless of wether it is filled or empty it seems to evaluate to true:
Screenshot 2021-04-30 at 09 03 03

Maybe this could use `.some()`?

akofink: `find` does not return an array https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

bastilian: Oh, sorry, my bad. I always mix up filter/find.

vkrizan (approved):
Works as expected. Thank you.

There is still some delay due the requests. If the inventory table displays more than 10 items per page the alert can be seen only by scrolling to bottom.

PR #1155, opened 2021-04-21
Refactoring notes - 04/2021

bastilian:
This PR only serves as a place to mark and keep track of places in the compliance frontend that could use some refactoring.
Feel free to also push notes and discuss the proposed spots to refactor.


akofink:
Maybe we should just report jiras for this instead so we can plan it.

sthirugn:
+1 to creating jira, otherwise we tend to forget or not plan time

PR #1170, opened 2021-04-22
fix(PolicyDetails): RHICOMPL-1253 - Restore no systems assigned warning

bastilian:
Brings back the "No systems" warning to the policy details:

Screenshot 2021-04-22 at 22 54 40

⚠️ Can be tested with LOCAL_CHROME and the latest frontend-components linked to include  https://github.com/RedHatInsights/frontend-components/pull/1087 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


src/SmartComponents/PolicyDetails/PolicySystemsTab.js
```diff
@@ -46,7 +48,8 @@ PolicySystemsTab.propTypes = {
     policy: propTypes.shape({
         id: propTypes.string.isRequired,
         complianceThreshold: propTypes.number.isRequired,
-        majorOsVersion: propTypes.string.isRequired
+        majorOsVersion: propTypes.number.isRequired,
+        hosts: propTypes.array
```
akofink, line 52: When does this component not require `hosts`? Can we make this `isRequired`?

akofink (commented):
The code looks good. I've been unable so far to test this. I'm using local chrome (via our docker-compose). I don't need FC linked as well, right?

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
bastilian: Good point! Updated it to have `isRequired`.

bastilian:
@akofink It actually is the chrome that needs to have the `frontend-components-inventory` package linked. You can also try and bump the version of `frontend-components-inventory` in the `insights-chrome` directory/package.json to version `3.1.14`. Then you shouldn't need to link the package.

akofink:
@bastilian I've tried again with local chrome and updated `frontend-components-inventory`. The systems tables then started querying the inventory API again instead of the compliance API :confused: It also was using some new query params; the inventory API was returning 400s for those queries. Any idea if I need anything else besides local chrome with the updated `frontend-components-inventory`?

bastilian:
@akofink There may be issues in the latest frontend-components-inventory package. (I've already consulted with the platform/inventory team regarding this and will open a PR with the changes needed, if needed).

As for this PR, to test it, I've updated the `@redhat-cloud-services/frontend-components-inventory` package only to version `3.1.14` in the `LOCAL_CHROME` `insights-chrome` directory. This is the release including the custom `noSystemsTable` prop, but not the columns/default-columns changes.

This package update in chrome should be done in its repository as well, for the next planned chrome update. We can wait until then to verify and test it as well.It is dependent on it anyways. 

However, I've recorded how it will work once the changes have fallen into place: 


https://user-images.githubusercontent.com/7757/116989283-d8c71480-acd1-11eb-9fa8-3632f51875e6.mp4

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
```diff
@@ -20,14 +21,16 @@ const PolicySystemsTab = ({ policy }) => (
         defaultFilter={`policy_id = ${policy.id}`}
         showActions={ false }
         remediationsEnabled={ false }
+        noSystemsTable={ policy?.hosts.length === 0 && <NoSystemsTableWithWarning /> }
```
akofink, line 24: This makes me think `policy` could be `undefined`, and then we're calling `length`, which would raise an uncaught TypeError.

akofink (approved):
Code makes sense. I'll try retesting this later (I tried with LOCAL_CHROME and `"@redhat-cloud-services/frontend-components-inventory": "^3.1.14"` before), or especially when chrome is updated.

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
bastilian: Yeah, you may be right. I'll check that!

bastilian: I added another `?` that should prevent any issues and also a test to verify. I don't think it would happen actually, but just to be sure nothing blows up.

bastilian:
@akofink Meanwhile these changes have been deployed to ci-beta, so you should be able to verify without any `LOCAL_CHROME` but you could use the `ci-beta` chrome.

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
akofink: The case when it could happen, I guess, is if the `policy` query didn't include `hosts`.

akofink (approved):
Seems to work fine on ci-beta, though there are a bunch of other, non-related breakages there as well. Still, the change is fairly straightforward :+1:

PR #1178, opened 2021-04-23
fix(EditPolicy): RHICOMPL-1742 save button disabled on empty systems (split)

vkrizan:
The updatedPolicy has been set more than once in EditPolicyForm whithin
three different useEffect callbacks.  Because the effects did not have it
as a dependency (for triggering reasons) they do not get the updated
content from the previous useEffect.  Since these useEffects can fire
all at the same time, one would not get the update from the other.  This
fixes this by splitting it to three separate state variables.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -77,6 +77,8 @@ export const EditPolicy = ({ route }) => {
     const dispatch = useDispatch();
     const anchor = useAnchor();
     const [updatedPolicy, setUpdatedPolicy] = useState(null);
+    const [selectedRuleRefIds, setSelectedRuleRefIds] = useState([]);
+    const [selectedHosts, setSelectedHosts] = useState();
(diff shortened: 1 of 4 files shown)


akofink:
This seems to jive with what is recommended in the docs: https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables
>If you miss automatic merging, you could write a custom useLegacyState Hook that merges object state updates. However, we recommend to split state into multiple state variables based on which values tend to change together.

akofink (approved):
This seems to be the recommended way to manage larger states with useState :+1: I think it makes it more clear as well which data are used by which components. Also, it fixes the bug! Thanks, @vkrizan!

PR #1200, opened 2021-04-30
refactor(usePolicy): RHICOMPL-1624 refactor and add tests 

vkrizan:
Refactored usePolicy hook and its related parts to separate modules into a band new location `src/Mutation/hooks`.
Added unit tests.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management


akofink (commented):
Awesome refactor! The tests are looking great, too! :100:

config/alias.webpack.config.js
```diff
@@ -6,6 +6,7 @@ const aliases = {
     'react-intl': resolve(__dirname, '../node_modules/react-intl'),
     PresentationalComponents: resolve(__dirname, '../src/PresentationalComponents'),
     SmartComponents: resolve(__dirname, '../src/SmartComponents'),
+    Mutations: resolve(__dirname, '../src/Mutations'),
```
bastilian, line 9: Could we not use a more overarching term like `GraphQL` that also in the future could hold all the queries, not only mutations?

src/Mutations/hooks/useAssociateRules.test.js
```diff
@@ -0,0 +1,83 @@
+import { useMutation } from '@apollo/client';
+import useAssociateRules from './useAssociateRules';
+jest.mock('@apollo/client');
+
+afterEach(() => {
```
bastilian, line 5: All other tests are wrapped in at least one `describe` block, please wrap these in one too.

vkrizan: There is exactly one component to test, so the describe seemed redundant. I also can switch the calls from `it()` to `test()` if that'd be acceptable.

bastilian: This is just about consistency. Any test file in this repository (and neighbouring ones) have their tests wrapped in a describe block, regardless of it just testing one function/component or more.

vkrizan: Okay.

config/alias.webpack.config.js
vkrizan: Any suggestions?

My assumption is that that non-mutating queries or API calls would not require special handling as opposed mutating ones.

bastilian: The suggestion would be `GraphQL`. Queries may not need much special treatment and their own hook, but we do have a number of queries in constants spread out over the application. It would be great to consolidate those in one place at some point and maybe even see if it is possible to combine and reduce them.

bastilian (approved):
Nice job! Thank you @vkrizan for refactoring these parts and adding tests!

The `Mutations` as the "category" is fine for now, and we can rename it if/when we want to consolidate all GraphQL related functions.

PR #1205, opened 2021-05-04
Migrate compliance system detail over from FEC repository

karelhala:
## Use federated modules for hosting system detail

Since we have now Webpack's module federation in all places, we can use them to host system detail. This way we don't have to open PRs to update compliance detail.

## Usage

```JSX
import React from 'react';
import { useRouteMatch } from 'react-router-dom';
import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
import fallback from '../SpinnerFallback';

const ComplianceTab = () => {
    const { params } = useRouteMatch('/:inventoryId');

    return <AsyncComponent
        appName="compliance"
        module="./SystemDetail"
        fallback={fallback}
        inventoryId={ params.inventoryId }
        customItnl
        intlProps={{
            locale: navigator.language
        }}
    />;
};
```


bastilian:
@karelhala The new implementation for the `RulesTable` is now merged[0]. This means you should be able to leave out `SystemRulesTable` and all related code and instead use `RulesTable` in the Presentational components. Furthermore there is no need to migrate any places that already use the new implementation, which should leave the `SystemsDetails` as the only place where this needs to be done.

The `FilterConfig..` related code has also been moved in under the `useTableTools` and can also be left out of moving and already use the one in compliance-frontend.

With this the PR should become significantly smaller.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1208

bastilian (commented):
Thank you @karelhala! I've tested it and everything appears to still be working as expected 🎉 .

I also looked over it and there are a few more lines I think we can get rid off.
Could you take a look at those? Then this would be ready to merge from my perspective. 

The `PresentationslComponents/PresentationalComponents` is a bit meh, but we can shuffle that around at another point when we clean up other path and filenames.

@akofink @skateman @vkrizan It would probably make sense to give this a look from another set of eyes. Maybe there is something I missed that could be omitted as well.

package.json
```diff
@@ -115,6 +116,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.6.3",
     "jest-canvas-mock": "^2.3.0",
+    "jest-fetch-mock": "^3.0.3",
```
bastilian, line 119: I don't think we are using this actually.

akofink (commented):
@bastilian it seems like this PR breaks the severity sort on the system details rules table. Can you confirm? I'm comparing against master. Other than that, I haven't found any other regressions.

bastilian:
@akofink I can confirm that sorting by severity is broken, but it appears the issue is not introduced with this PR. I'm currently investigating what the issue is.

akofink:
> it appears the issue is not introduced with this PR

So it's also broken for you with master? I see it working fine with master for some reason.

src/PresentationalComponents/SupportedSSGVersionsLink/__snapshots__/SupportedSSGVersionsLink.test.js.snap
```diff
@@ -2,7 +2,6 @@
 
 exports[`SupportedSSGVersionsLink expect to render without error 1`] = `
 <a
-  href="https://access.redhat.com/documentation/en-us/red_hat_insights/2021/html/assessing_and_monitoring_security_policy_compliance_of_rhel_systems/index"
```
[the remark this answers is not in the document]
karelhala: Good point! This let me to removing additional file which was copied by mistake.

bastilian:
@karelhala Hej, thanks for updating the PR! I've tested it once more and noticed that we were missing support for the grouped filter we recently added in https://github.com/RedHatInsights/frontend-components/pull/1073. I've opened a PR here[0] to add the same to the `useFilterConfig` TableTools hook and allow migrating the two places that rely on `inventory-compliance` away before. 

After that is merged and this is rebased all should be good to merge this as well.

Another minor thing that changed meanwhile is the link[1], so I think that `2020...` link doesn't need to be copied, but I'm not 100% sure which one is the right one yet, so we can iron that out after merging. 

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1297
[1] https://github.com/RedHatInsights/frontend-components/pull/1131

src/PresentationalComponents/PoliciesTable/PoliciesTable.js
```diff
@@ -6,7 +6,7 @@ import { Button, Pagination, PaginationVariant, ToolbarItem, TextContent } from
 import PrimaryToolbar from '@redhat-cloud-services/frontend-components/PrimaryToolbar';
 import TableToolbar from '@redhat-cloud-services/frontend-components/TableToolbar';
 import { conditionalFilterType } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
-import { FilterConfigBuilder } from '@redhat-cloud-services/frontend-components-inventory-compliance/Utilities';
+import FilterConfigBuilder from '../../Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder';
```
bastilian, line 9: We also have an alias for Utilities, so this could be written without the `../../` like
```js
import FilterConfigBuilder from 'Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder';
```

src/Utilities/hooks/useFilterConfig.js
```diff
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { FilterConfigBuilder } from '@redhat-cloud-services/frontend-components-inventory-compliance/Utilities';
+import FilterConfigBuilder from './useTableTools/FilterConfigBuilder/FilterConfigBuilder';
```
bastilian, line 2: With https://github.com/RedHatInsights/compliance-frontend/pull/1297 merged and rebased on top of it, this won't be needed anymore

bastilian:
@karelhala Apologies that this took so long, but https://github.com/RedHatInsights/compliance-frontend/pull/1297 is finally merged! If this is rebased on top of master it should only require to move in the Compliance Systems Details and the Remediations button.

If it is needed, let me know, and I can also take care of rebasing this PR and finishing it up.

bastilian (approved):
Thank you @karelhala! Everything is now looking good!

PR #1209, opened 2021-05-06
refactor(TabbedRules): RHICOMPL-1831 centralize rule selection

vkrizan:
Moves common handling of `selectedRuleRefIds` into `TabbedRules` component. Adds an utility function for transforming profiles with rules into the selection format.

This is a preparation for fixing the RHEL 6 cases where one canonical profile can share more than one OS minor version.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management


vkrizan:
>  when creating a new policy (RHEL 7 CCP), moving from the systems (one RHEL 7.9, one RHEL 7.7) step to rules step redirects me to the first wizard step. Are you also able to reproduce this?

I wasn't able to reproduced it. I can only wonder what has happened there. Maybe some unexpected state variable resets. Can you retry and observe what is happening within the redux form?

src/PresentationalComponents/TabbedRules/TabbedRules.js
```diff
@@ -18,12 +18,47 @@ const defaultTab = (tabsData, profileId) => {
...
+const selectedRuleRefIdsForTab = (selectedRuleRefIds, { id }) => {
+    const tabSelection = (selectedRuleRefIds || []).find((selection) =>
+        selection.id === id
+    );
+    return tabSelection?.ruleRefIds || [];
+};
+
+export const profilesWithRulesToSelection = (profiles, prevSelection = [], options = {}) => {
+    const { only } = options;
+    const additionalSelection = profiles.map((profile) => {
```
akofink, line 30: Do you think we need to enforce somehow that profiles are passed in with rules?  Otherwise, we could forget to request them in the API query and receive a TypeError when calling `profile.rules.map`. At the very least, safe navigation on line 35?

akofink (approved):
Seems to work as expected! Thanks, @vkrizan! I'm not sure what was going on the first time I tried it out - sorry about that. Just one minor concern inline.

src/PresentationalComponents/TabbedRules/TabbedRules.js
vkrizan: I guess we can place a `console.error` and continue with no rule selection if that'd be a better option.

akofink: I'm not super worried about it - we'll almost certainly find this sort of error in development and not in deployed environments (It'd be a pretty obvious breakage). That's why I've ACKed the PR and leave it up to you to leave it as is or handle it.

vkrizan: Added. We can add more checks within next upcoming changes.

PR #1218, opened 2021-05-14
feat(Systems):  RHICOMPL-1716 - Enable OS minor version filter for policy OS versions

bastilian:
This enables the OS minor version filter for the `ComplianceSystems` component  and enables the group selection option (to be) introduced for the group filter[0].

https://user-images.githubusercontent.com/7757/118270604-93b89480-b4c0-11eb-81d9-18985cca2d61.mp4

⚠️ For the group selection to work the `frontend-components` package needs to be linked in `LOCAL_CHROME`, not compliance-frontend.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

[0] https://github.com/RedHatInsights/frontend-components/pull/1112

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -7,10 +7,10 @@ import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import { osMinorVersionFilter } from './constants';
 import useExport from 'Utilities/hooks/useTableTools/useExport';
 
-const groupByMajorVersion = (versions = [], showFilter) => {
+const groupByMajorVersion = (versions = [], showFilter = []) => {
     const showVersion = (version) => {
         if (showFilter.length > 0) {
-            return Array(showFilter).map(String).includes(String(version));
+            return showFilter.map(String).includes(String(version));
         } else {
             return true;
         }
--- a/src/constants.js
+++ b/src/constants.js
@@ -52,6 +52,7 @@ const toSystemsOsMinorFilterConfigurationItem = (osVersions) => (
     (majorVersion) => ({
         label: `RHEL ${ majorVersion }`,
         value: majorVersion,
+        groupSelectable: true,
         items: osVersions[majorVersion].map((minorVersion) => ({
             label: `RHEL ${ majorVersion }.${ minorVersion }`,
             value: minorVersion
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -52,7 +52,7 @@ export const ComplianceSystems = () => {
                             systemProps={{
                                 isFullView: true
                             }}
-                            showOsFilter
+                            showOsMinorVersionFilter={ policies.map((policy) => (policy.majorOsVersion)) }
                             showComplianceSystemsInfo
                             enableEditPolicy={ false }
                             remediationsEnabled={ false }


src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -8,7 +8,7 @@ import { osMinorVersionFilter } from './constants';
 const groupByMajorVersion = (versions = [], showFilter) => {
     const showVersion = (version) => {
         if (showFilter.length > 0) {
-            return Array(showFilter).map(String).includes(String(version));
+            return Array(...new Set(showFilter)).map(String).includes(String(version));
```
bastilian, line 11: For some very odd reason, here, without making sure each string is only in the array once, the `includes` will return `false`. 🤷

vkrizan: The `map` there concats the strings:
```
> Array(['a', 'a', 'a']).map(String)
Array [ "a,a,a" ]
```

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -8,7 +8,7 @@ import { osMinorVersionFilter } from './constants';
 const groupByMajorVersion = (versions = [], showFilter) => {
```
vkrizan, line 8: Observation: The `showFilter` seems like a boolean, but it acts as a filter for OS major version. Naming could be improved.

src/constants.js
```diff
@@ -52,6 +52,7 @@ const toSystemsOsMinorFilterConfigurationItem = (osVersions) => (
     (majorVersion) => ({
         label: `RHEL ${ majorVersion }`,
         value: majorVersion,
+        groupSelectable: true,
```
vkrizan, line 55: There seems to be no group checkbox:
Screenshot_2021-05-19 Systems - Compliance Red Hat Insights

bastilian: 🤦 damn. my bad, i was still testing with `LOCAL_CHROME` when trying to check if it's deployed. And it seems it is not yet deployed, so it still requires `LOCAL_CHROME` with the latest `frontend-components` in `insights-chrome` linked.

src/SmartComponents/SystemsTable/hooks.js
bastilian: It was actually just the `Array(['a', 'a', 'a'])`. Not even sure why I went all fancy here.

src/SmartComponents/SystemsTable/hooks.js
bastilian: `showFilter` is basically a shortened version of `showOsMinorVersionFilter` prop. I added an array default to indicate it is used as an array in this function. For `useOsMinorVersionFilter` it should be obvious that it is used as a boolean.

vkrizan: Neither the `showOsMinorVersionFilter` says that it expects/holds the configuration for the filter to limit the selection. Here in this function it could be `osMajorVersion` or `options = {}`.

Within the prop on the InventoryTable the `showOsMinorVersionFilter` can be hash of options to configure the filter. In that case it could be `osMinorVersionFilter`. The `showFilter` might as well be an option there (with default true). That would separate the configuration and enable/disable option.
Example:
```
osMinorVersionFilter={ { osMajorVersions: [...] } }
```

vkrizan (approved):
Thanks. 

I don't have local Chrome, but I trust that it would work as expect.

akofink:
So, before merge, this will need a chrome release that includes the updated FC package, right? We will be able to test it without local chrome at that point, right?

bastilian:
@akofink Yes, this will still require a package bump in `insights-chrome` of the `frontend-components` package to at least `3.1.12` to include the necessary change. A new chrome release to CI/QA environments is planned in the next few weeks, which should include an update of that package.

https://github.com/RedHatInsights/insights-chrome/blob/master/package.json#L154

bastilian:
insights-chrome (master) has the dependencies now updated and it should be possible to test this with just the latest chrome as `LOCAL_CHROME` and without having to link  the packages.

@vkrizan @akofink Feel free to take another look to verify this.

PR #1221, opened 2021-05-18
fix(Inventory): RHICOMPL-1839 remediation of supported systems 

vkrizan:
First this fixes the data that are passed to `selectedSystems` state variable. Previously it appended tabe row data instead of the original entity data.

Mainly this fixes Remediation button on Systems Table (e.g. on Report Details page). This is done by passing preprocessed data (esp. `supported`) from `selectedSystems` into the remediation button.

## Secure Coding Checklist
- [X] Input Validation
- [X] Output Encoding
- [X] Session Management
- [X] Access Control
- [X] Cryptographic Practices
- [X] Data Protection
- [X] Communication Security
- [X] System Configuration
- [X] Database Security
- [X] File Management
- [X] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -3,6 +3,7 @@ import { useApolloClient } from '@apollo/client';
 import { useDispatch } from 'react-redux';
 import debounce from '@redhat-cloud-services/frontend-components-utilities/debounce';
 import useCollection from 'Utilities/hooks/api/useCollection';
+import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import { osMinorVersionFilter } from './constants';
 
 const groupByMajorVersion = (versions = [], showFilter) => {
@@ -59,8 +60,10 @@ export const useFetchSystems = (
                 ...policyId && { policyId }
             }
         }).then(({ data }) => {
+            const systems = data?.systems?.edges?.map((e) => e.node) || [];
+            const entities = systemsWithRuleObjectsFailed(systems);
             const result = {
-                entities: data?.systems?.edges?.map((e) => e.node) || [],
+                entities,
                 meta: {
                     totalCount: data?.systems?.totalCount || 0
                 }
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -99,9 +99,15 @@ export const useOnSelect = (onSelectProp, items, preselectedSystems, total) => {
     );
 
     const onSelect = (_event, select, _index, row) => {
-        const selected = select ?
-            [...selectedSystems, row] :
-            selectedSystems.filter((selected) => (selected.id !== row.id));
+        const system = items.find(({ id }) => id === row.id);
+        if (!system) {
+            console.error(`System identifed as ${row.id} not found in items for selection!`);
+            return;
+        }
+
+        const selected = select
+            ? [...selectedSystems, system]
+            : selectedSystems.filter((selected) => (selected.id !== row.id));
 
         setSelected(selected);
         onSelectCallback(selected);
--- a/src/SmartComponents/SystemsTable/Cells.js
+++ b/src/SmartComponents/SystemsTable/Cells.js
@@ -4,7 +4,9 @@ import { Link } from 'react-router-dom';
 import { Text, TextContent, TextVariants } from '@patternfly/react-core';
 import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat';
 import { UnsupportedSSGVersion, ComplianceScore as complianceScore } from 'PresentationalComponents';
-import { profilesRulesPassed, profilesRulesFailed } from 'Utilities/ruleHelpers';
+import {
+    profilesRulesPassed, profilesRulesFailed, systemSupportedByProfiles
+} from 'Utilities/ruleHelpers';
 
 const SystemLink = ({ id, children }) => (
     <Link to={{ pathname: `/systems/${id}` }}>{ children }</Link>
@@ -113,14 +115,13 @@ export const complianceScoreData = (profiles) => {
     const compliant = profiles.every(profile => (
         profile.lastScanned === NEVER || profile.compliant === true
     ));
-    const supported = (profiles || []).reduce((acc, profile) => acc && profile.supported, true);
 
     return {
         score,
         rulesPassed,
         rulesFailed,
         compliant,
-        supported
+        supported: systemSupportedByProfiles(profiles)
     };
 };
 
--- a/src/SmartComponents/SystemsTable/InventoryTable.js
+++ b/src/SmartComponents/SystemsTable/InventoryTable.js
@@ -7,7 +7,6 @@ import ComplianceRemediationButton from '@redhat-cloud-services/frontend-compone
 import { exportItems } from 'Utilities/Export';
 import { DEFAULT_SYSTEMS_FILTER_CONFIGURATION, COMPLIANT_SYSTEMS_FILTER_CONFIGURATION } from '@/constants';
 import { ErrorPage, StateView, StateViewPart } from 'PresentationalComponents';
-import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import useFilterConfig from 'Utilities/hooks/useFilterConfig';
 import { InventoryTable as FECInventoryTable } from '@redhat-cloud-services/frontend-components/Inventory';
 import { policyFilter, defaultOnLoad } from './constants';
@@ -48,8 +47,6 @@ export const InventoryTable = ({
     const {
         onSelect, onBulkSelect, selectedSystems, isPageSelected
     } = useOnSelect(onSelectProp, items, preselectedSystems, total);
-    const selectedIds = selectedSystems?.map((system) => (system.id));
-    const selectedItems = items.filter((item) => (selectedIds.includes(item.id)));
     const selectedCount = selectedSystems.length;
 
     const osMinorVersionFilter = useOsMinorVersionFilter(showOsMinorVersionFilter);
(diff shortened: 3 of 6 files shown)


src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -99,9 +102,15 @@ export const useOnSelect = (onSelectProp, items, preselectedSystems, total) => {
     );
 
     const onSelect = (_event, select, _index, row) => {
-        const selected = select ?
-            [...selectedSystems, row] :
-            selectedSystems.filter((selected) => (selected.id !== row.id));
+        const system = items.find(({ id }) => id === row.id);
```
bastilian, line 105: You can't rely on `item` since it only includes the systems on the current page.

vkrizan: I had the same concern. I've reviewed the bulk select, and it pretty much depends on what entities are used to draw the page... When would it happen that a selected row isn't selected on the currently viewed page?

bastilian: It can happen at any point that a selected system is not on the current page, for example if a user has 100 systems and the perPage is 50 and they select one or two of the first 50, go to page 2 and the `selectedSystems` would not be on the current page anymore and the selection would be lost.

Is the change to `systemsWithRuleObjectsFailed` not enough?

vkrizan: This block of code is triggered for an `onChange` event from the row checkbox. If you select from page 1 and then go to page 2 and select another one, the selection from the previous state and page is kept. This only appends new item from the current view to the `selectedSystems`. Where do you see it being lost?

bastilian: Oh, i see. Still doesn't seem necessary to change this at all, or?

vkrizan: I've refactored/fixed it, because it combined two concepts (original data and row data). It would have worked with row data, however it changes the original data which might lead to bugs (if it were to override a property). For example there is a `name` injected. For systems this is okay atm (`name` vs `display_name`) but in contrast it would not be okay for policies where `name` is a regular property. I guess change a name of the column could lead to an issue.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -59,8 +60,10 @@ export const useFetchSystems = (
                 ...policyId && { policyId }
             }
         }).then(({ data }) => {
+            const systems = data?.systems?.edges?.map((e) => e.node) || [];
+            const entities = systemsWithRuleObjectsFailed(systems);
```
bastilian, line 64: Why does this need to be here? 

It should be sufficient to change the allSystems attribute of the `ComplianceRemediationButton` to `systemsWithRuleObjectsFailed(selectedSystems)`, is it not?

src/SmartComponents/SystemsTable/hooks.js
bastilian: > because it combined two concepts (original data and row data)

I'm not really sure I know what you mean by that. There is or shouldn't be any difference in what is passed in to the `FECInventoryTable` via `getEntities` and what the `onSelect` callback receives as a "row".

vkrizan: There is a difference, hence this refactoring. `entity !== row`. I am happy to explain in person.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: This is there so the reshape is done once on loaded data as opposed to reshaping it on every selection change. Do you see a some complications with this approach?

src/SmartComponents/SystemsTable/hooks.js
bastilian: The only difference is that within the InventoryStore they are called "rows" once the items are loaded[0].
When they are selected they are returned as "entity"[1] and in between they aren't modified, as far as I can see, which I see as being the same and not being different.

[0] https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory/src/redux/entities.js#L116
[1] https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory/src/redux/entities.js#L135

src/SmartComponents/SystemsTable/hooks.js
bastilian: There shouldn't be any issue at all with "reshaping" it on select, this is what we do now and it should actually be better, since it will only have to process selected systems as opposed to all loaded systems, would it not?

src/SmartComponents/SystemsTable/hooks.js
vkrizan: Here is what's returned within `row`:
```
Object { id: "4a972d07-7e61-4d5e-8e96-f2a2b0ec54d9", name: {…}, osMajorVersion: 7, osMinorVersion: 9, testResultProfiles: (1) […], policies: (1) […], __typename: "System", ruleObjectsFailed: (95) […], supported: true, profiles: (1) […], … }
​__typename: "System"
​actionProps: Object { "data-ouia-component-id": "4a972d07-7e61-4d5e-8e96-f2a2b0ec54d9-actions-kebab" }
​cells: Array(5) [ {…}, {…}, {…}, … ]
​"compliance-score": Object { title: undefined, formatters: [], props: {…} }
​"failed-rules": Object { title: undefined, formatters: [], props: {…} }
​id: "4a972d07-7e61-4d5e-8e96-f2a2b0ec54d9"
​isExpanded: undefined
​isFirst: true
​isFirstVisible: true
​isHeightAuto: false
​isLast: true
​isLastVisible: true
​"last-scanned": Object { title: undefined, formatters: [], props: {…} }
​name: Object { title: undefined, formatters: [], props: {…} }
​osMajorVersion: 7
​osMinorVersion: 9
​policies: Array [ {…} ]
​profiles: Array [ {…} ]
​ruleObjectsFailed: Array(95) [ {…}, {…}, {…}, … ]
​secretTableRowKeyId: "4a972d07-7e61-4d5e-8e96-f2a2b0ec54d9"
​selected: false
​"ssg-version": Object { title: undefined, formatters: [], props: {…} }
​supported: true
​testResultProfiles: Array [ {…} ]
​<prototype>: Object { … }
hooks.js:106:16
```
a bunch of internal table data.

bastilian: Oh, i overlooked the createRows, sorry. 

I'm still not sure about this change. There is no advantage in querying  `items` for a system, the original properties are still preserved as is when returned onSelect and eventually we'll only want to have system ids as a selection and not have any more data in the selection or any part rely on data within the selection. This move to just using ids will likely happen when enabling a proper BulkSelect. The whole `useOnSelect` hook is meant to be a temporary solution till we can do that.

There are a couple of places that currently rely on data of the selected items like the `RemdediationsButton`, the export, as well as the Edit Policy and Create Wizard, maybe some other place as well. Which makes me a bit more worried about changing what is in the selection and where it comes from. We will hopefully remove the need for these places to require additional data in the selection in the foreseeable future, as part of performance improvements and when updating the systems export to use the API rather than the selectedSystems. Other parts may take longer or we provide them with the data, but not via selectedSystems. And eventually we can remove the whole `useOnSelect` hook.

Please, let's leave the `onSelect` as is.

bastilian: Looking at the `onBulkSelect` it already does use items, however, it shouldn't, there might even be bugs already that we haven't discovered yet specifically with this `onBulkSelect`. It works for now, but it is not well done and probably error prone. I'd rather kick them both then try to refactor them.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: There is a trade-off. One of the reasons I moved it it here is that the data shows a spinner on loading as opposed to selecting a row on a table which doesn't. From the user experience perspective is better to have it under the spinner vs. hanging the page on checkbox (un)check.

bastilian: The hanging checkbox is probably more likely due to the amount of data we put into the selected state and other factors, rather than the state being prepared. The Button itself does some evaluating of the data passed in before it updates the enable/disable state, which adds another delay.

If we wanna have it "under a spinner" we should move the `supported` attribute to the backend and expose it in the API so we don't have to do any data preparations on the client and can just pass along the API result.

Even better would be an attribute like `remediatable` which indicates if a system is supported, has failed rules and these rules have remediations, this would remove the need for any data inspection on the client to determine if the button should be enabled. The RefIds for rules themselves could then be requested when the user actually opens the remediations wizard. This would also remove the need for the Remediations Button requiring complex objects in the selected state.

akofink (approved):
Seems to work as expected. I tried testing with multiple pages of hosts - the inventory table seems pretty broken right now with bulk select (rows in other pages aren't selected, even though the total selected shows the correct number), but all systems are still sent to the remediations API regardless. Pending @bastilian's code concerns :)

https://slack-files.com/T026NJJ6Z-F0239T1RSQ0-eda33438fd

src/SmartComponents/SystemsTable/hooks.js
vkrizan: Indeed, there is a room for improvement.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: There are implication of not using objects within the selection. There is stuff mapped, like osMinorVersions to render the tabs, or the rules and supported for the remediation... I like the spirit for optimization, hopefully we'll get there.

I believe that this change improves what data are present within the selection. We've tested the selection on various of contexts. This change is small enough to find a problem if there are any bugs uncovered.

bastilian: An improvement would be to remove the logic of querying `items` from the `onBulkSelect`. This introduces the same wrong logic into another place. This then would also fix the issues that @akofink is seeing with it and multiple pages.

All places that rely on data in the selection shouldn't, because they can't really control what is going to end up in there and maintaining it to satisfy the need of every place is and has been a pain. So the goal should be to go towards removing the need for these places to use that data to eventually be able and just have IDs as a selection. And not add more data for them use.

PR #1257, opened 2021-06-02
feat(SystemsTable): RHICOMPL-789 - Enable export of all systems

bastilian:
This turns our existing export into a TableTools hook and integrates it into the Inventory Table.
The export will still be compiled on the client to support JSON as well as CSV and load systems in batches.

I was initially hoping that we could literally just open a link like a normal download for the export, but CSV would have still needed to be built somewhere, and the backend doesn’t support CSV as a format, yet. Maybe we can move that some day to the backend.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


src/Utilities/hooks/useTableTools/useExport.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    return cell;
+};
+
+export const csvForItems = ({ items, columns }) => {
+    const header = columns.map((column) => (column.title)).join(CSV_DELIMITER);
+    const csvRows = [header, ...items.map((row) => (
+        columns.map((column) => (
+            textForCell(row, column)
```
vkrizan, line 39: Previously string values have been quoted in CSV. Does this still happen?

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -180,3 +188,55 @@ export const useInventoryUtilities = (inventory, selectedSystems, activeFilters)
...
+        const pages = Math.floor(total / BATCH_SIZE) + 1;
+        return Promise.all([...new Array(pages)].map((_, pageIdx) => (
+            fetchSystems(BATCH_SIZE, pageIdx + 1, filter)
+        )));
+    };
+
+    const selectedFilter = () => {
+        if (selected?.length > 0) {
+            return selected.map((id) => (
+                `id = ${ id }`
```
vkrizan, line 223: I wonder whether we can use the caret `^` symbol to do an IN operator. 
```
id ^ "UUID1,UUID2,.."
```

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -180,3 +188,55 @@ export const useInventoryUtilities = (inventory, selectedSystems, activeFilters)
...
+        const BATCH_SIZE = 100;
+        const pages = Math.floor(total / BATCH_SIZE) + 1;
+        return Promise.all([...new Array(pages)].map((_, pageIdx) => (
+            fetchSystems(BATCH_SIZE, pageIdx + 1, filter)
+        )));
+    };
+
+    const selectedFilter = () => {
+        if (selected?.length > 0) {
+            return selected.map((id) => (
```
vkrizan, line 222: Observation: This repeats what `toIdFilter` already seems to do (stored in `selectionFilter` variable).

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -10,14 +10,19 @@ import {
...
 export const Name = {
     title: 'Name',
     props: {
         width: 40,
         ...disableSorting
     },
-    exportKey: 'name',
-    renderExport: (name) => (String(name)),
+    renderExport: (system) => (
+        `${ system.name } (${ operatingSystemString(system) })`
```
vkrizan, line 24: OS information could deserve it's own column on the export.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -180,3 +188,55 @@ export const useInventoryUtilities = (inventory, selectedSystems, activeFilters)
...
+            filter: selectionFilter ?
+                `${ filter } and (${ selectionFilter })` : filter,
+            ...policyId && { policyId }
+        }
+    });
+
+    const fetchBatched = (total, filter) => {
+        const BATCH_SIZE = 100;
+        const pages = Math.floor(total / BATCH_SIZE) + 1;
+        return Promise.all([...new Array(pages)].map((_, pageIdx) => (
```
vkrizan, line 215: How are the pages reconstructed? What is the final order of the items?

vkrizan (commented):
The export works well. Couple of things to consider are below.

Thank you for splitting the commits. It helped much!

Also I've noticed that the positioning of the export icon on the toolbar seems weird on Report details. Not sure if you'd want to fix it here.
Screenshot_2021-06-03 Report Health Insurance Portability and Accountability Act (HIPAA) - Reports - Compliance Red Hat Ins

src/Utilities/hooks/useFilterConfig.js
```diff
@@ -65,7 +65,7 @@ const useFilterConfig = (initialConfig, arrayToFilter) => {
         filtered: filteredArray,
         activeFilters,
         activeFilterValues,
-        buildFilterString: () => filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters)
+        filterString: filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters)
```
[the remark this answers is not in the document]
bastilian: Yep, because they are not the same. Eventually we will remove this one and use only the one in TableTools

src/SmartComponents/SystemsTable/Columns.js
bastilian: Yeah, that is not yet possible. I tried to think about how to do that, but not sure how we can implement that neatly.

src/SmartComponents/SystemsTable/hooks.js
bastilian: We use this only for the export, so there aren't actually pages, it's just the wording, it could also be "batch". We can not guarantee the same order of items until we implement sorting properly. For now it should not matter, as long as all items are fetched.

src/Utilities/hooks/useTableTools/useExport.js
bastilian: Good catch, must have snuck out somehow.

src/SmartComponents/SystemsTable/hooks.js
bastilian: I'll try that! I was not aware of the `^` operator.

bastilian:
@vkrizan Thanks for taking a look! I'll address the comments. I'll take a look at the button position, I think i know what's happening there.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: That's the one we had a problem with the CDN, so be aware. :slightly_smiling_face:

src/SmartComponents/SystemsTable/hooks.js
vkrizan: Isn't it a page in the end, I mean it ends up in the API call to retrieve a limited set. Can the resulting set order the batches? The API returns items in some "order". We can keep that order. Otherwise the exported items would have a random order that does not follow the rendered pages on the table view.

bastilian: For now, if the API keeps some order between requests, with just the page parameter changed, we should get all items in the order the API returned them. However, if the API, for some reason, changes the order in between requests or new items are added to the set on the server, the order and accuracy of the resulting export is off.

To get a 100% accurate export it should really just be a link to an API result or an export endpoint and all of the exporting happens on the server. We'd also need to consider (very) large lists, in case we have a limit in the API and to get all the user would need to paginate too.

I think with this, for now, we are as accurate as we can be on the client and it is only limited to exporting.

vkrizan: The concern is around the client (fronted) doing parallel requests and then combining the results from all of them in the order of their response. Could it respect at least the page/batch order within the final result/set?

Of course we cannot guarantee that page/batch crawl would give 100% correct response if there was a change in the middle of it, as there is no transaction. What we can do is a best effort.

bastilian: There should not be any issue here. Once all promises resolve the returned array of promise results should be in the same order as they were called mapping them.

vkrizan: Oh, nice. In that case it is all good. Thanks.

src/Utilities/hooks/useTableTools/useExport.js
```diff
@@ -46,9 +46,10 @@ export const csvForItems = ({ items, columns }) => {
 export const jsonForItems = ({ items, columns }) => {
     const result = items.map((row) => (
         columns.reduce((object, column) => {
-            const key = _.snakecase(column.title);
-            object[key] = textForCell(row, column);
+            const key = camelCase(column.title);
```
vkrizan, line 49: Nit: This one changes the SSG Version to sSGVersion. :slightly_smiling_face:

vkrizan (approved):
Great one. Thanks!

bastilian:
@vkrizan All comments should be addressed. I've looked at the issue with the shifted export button, but a fix would be more appropriate in the frontend-components either in the inventory or the `PrimaryToolbar`. I'll look further into that.

src/Utilities/hooks/useTableTools/useExport.js
bastilian: LOL. I didn't think about that. i think we can prevent that, i'll take a look.

PR #1259, opened 2021-06-02
chore(deps): RHICOMPL-1784 - Bump frontend-components and update to npm 7

bastilian:
`frontend-componetns` was recently updated to npm v7, this updates these packages to that version.
It also updates `compliance-frontend`s package lockfile to npm v7.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


vkrizan (approved):
Works nominally. I had to clear the `node_modules/` dir to make compiler happy.

The npm 7.6.0 adds empty `"requires": {}` to some of the dependencies in the package-lock file for some reason.

Dockerfile
```diff
@@ -15,8 +15,9 @@ RUN dnf -y --disableplugin=subscription-manager module enable nodejs:10 && \
        make gcc-c++ git && \
     dnf --disableplugin=subscription-manager clean all
 
+RUN npm install -g npm@latest
```
akofink, line 18: ubi8/nodejs-14 has npm 6.14, but maybe we should consider using that image anyway? I'm not too concerned, though, given this doesn't run in a container in prod.

bastilian: Hopefully there will be a `ubi8/nodejs-15` that should then use npm v7. Till then I think this is ok. I did just see a difference with using node v10 vs v15 when running tests.

We could install `nvm` in the container and use it to install node and npm, that way we can set the desired node and npm versions.

bastilian:
@akofink @vkrizan I think we should add `nvm` to our frontend container. We do not deploy or use this image anywhere critical, since it is only for development. `nvm` would allow us to align all `node` and `npm` versions with other places where it needs to run or build.

We should also make sure that we all can use the same or compatible versions to avoid any "it runs here, but not there" issues.
I've used  node v15.14.0  and npm v7.7.6 (via `nvm`), if these aren't possible to install easily on any platform you work on we can see what latest node version is we can use. `npm` should be possible to install via it's update mechanism and not depend on node.

Dockerfile
vkrizan: If this update mechanism works for you then we don't need to invest time with the `nvm`, yet.

bastilian: @vkrizan Well, right now we seem to be able to get to node v14 and npm 7.7.6 within the container on ubi, which seems to work. If this is ok for others, i'm fine with it. I wanna make sure everyone can run the same version of node and npm to avoid any environment differences. For example if you install and run the test in this branch with node v10 they will fail due to snapshot differences.

akofink:
I'm just going to run whatever container is included with this repo and don't really care how it's built :) I suspect UBI will _always_ be behind what Insights wants to use, unfortunately :disappointed: For the backend, I think I'm pretty much always going to be against trying to add more recent versions of ruby/rubygems, etc. on top of the UBI base image, since performance and security are important considerations the UBI team tries to solve for (and we run it in prod, of course).

PR #1265, opened 2021-06-04
feat(SystemsTable): RHICOMPL-1657 - Pass filters and orderBy to API

bastilian:
This adds sorting to the SystemsTable. For now only for the name and operating system column, and only in the Wizard, but we can gradually add more columns when we implement the backend part for them.

https://user-images.githubusercontent.com/7757/122219630-e9150680-ceaf-11eb-9e7a-703117929705.mp4

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.js
@@ -67,10 +67,17 @@ export const EditPolicySystems = ({ change, osMajorVersion, selectedSystems }) =
                         showOsMinorVersionFilter={ [osMajorVersion] }
                         prependComponent={ <PrependComponent osMajorVersion={ osMajorVersion } /> }
                         emptyStateComponent={ <EmptyState osMajorVersion={ osMajorVersion } /> }
-                        columns={[
-                            Columns.Name,
-                            Columns.OperatingSystem
-                        ]}
+                        columns={[{
+                            ...Columns.Name,
+                            props: {
+                                width: 40
+                            },
+                            sortBy: ['name']
+                        }, {
+                            ...Columns.OperatingSystem,
+                            props: {},
+                            sortBy: ['osMajorVersion', 'osMinorVersion']
+                        }]}
                         remediationsEnabled={false}
                         compact
                         showActions={ false }
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -94,8 +94,7 @@ export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
     );
 
     return async (_ids, { page = 1, per_page: perPage, orderBy, orderDirection, filters }) => {
-        const columnKey = orderBy.split(' ')[0];
-        const sortableColumn = findColumnByKey(columnKey);
+        const sortableColumn = findColumnByKey(orderBy);
         const sortBy = sortableColumn && sortableColumn.sortBy ?
             appendDirection(sortableColumn.sortBy, orderDirection) : undefined;
 
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -57,7 +57,7 @@ export const useFetchSystems = ({
 }) => {
     const client = useApolloClient();
 
-    return (perPage, page) => (
+    return (perPage, page, requestVariables = {}) => (
         client.query({
             query,
             fetchResults: true,
@@ -65,7 +65,8 @@ export const useFetchSystems = ({
             variables: {
                 perPage,
                 page,
-                ...variables
+                ...variables,
+                ...requestVariables
             }
         }).then(({ data }) => {
             const systems = data?.systems?.edges?.map((e) => e.node) || [];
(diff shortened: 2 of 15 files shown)


src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -84,8 +85,11 @@ export const useFetchSystems = ({
 };
 
 export const useGetEntities = (fetchEntities, { selected }) => (
-    async (_ids, { page = 1, per_page: perPage }) => {
-        const fetchedEntities = await fetchEntities(perPage, page);
+    async (_ids, { page = 1, per_page: perPage, orderBy, orderDirection, filters }) => {
+        const fetchedEntities = await fetchEntities(perPage, page, {
+            filters,
```
bastilian, line 90: We would not need this, and it doesn't do anything yet, but while we are here.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -84,8 +85,11 @@ export const useFetchSystems = ({
...
+            orderBy: `${orderBy} ${orderDirection}`
```
bastilian, line 91: We can also split them into two variables.

akofink: Shouldn't this use the colon syntax, @skateman?

```js
            orderBy: `${orderBy}:${orderDirection}`
```

bastilian: It does. This is a comment on an old outdated commit. :) I tried deleting it but GitHub apparently didn't let me.

skateman: Dfeinitely a colon

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -83,20 +84,38 @@ export const useFetchSystems = ({
...
+    const appendDirection = (attributes, direction) => (
+        attributes.map((attribute) => `${attribute}:${direction}`)
+    );
+
+    const findColumnByKey = (key) => (
+        (columns || []).find((column) => column.key === key)
+    );
+
+    return async (_ids, { page = 1, per_page: perPage, orderBy, orderDirection, filters }) => {
+        const columnKey = orderBy.split(' ')[0];
```
vkrizan, line 97: Can you please elaborate on the split by a space and taking the first element?

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -67,10 +67,17 @@ export const EditPolicySystems = ({ change, osMajorVersion, selectedSystems }) =
...
-                        columns={[
-                            Columns.Name,
-                            Columns.OperatingSystem
-                        ]}
+                        columns={[{
+                            ...Columns.Name,
+                            props: {
+                                width: 40
+                            },
+                            sortBy: ['name']
```
vkrizan, line 75: Can you also please add this column configuration to the EditPolicy modal? It is possible to click there to sort, but no sorting is happening.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -83,20 +84,38 @@ export const useFetchSystems = ({
...
+        const sortableColumn = findColumnByKey(columnKey);
+        const sortBy = sortableColumn && sortableColumn.sortBy ?
+            appendDirection(sortableColumn.sortBy, orderDirection) : undefined;
+
+        const fetchedEntities = await fetchEntities(perPage, page, {
+            filters,
+            sortBy
```
vkrizan, line 104: Once an order is changed, the whole table goes to preload state and hides the table header and page information for a brief moment. It jumps around. I'd expect only the table contents to be preloaded keeping the table header and page stable.

(line unrelated)

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -67,10 +67,17 @@ export const EditPolicySystems = ({ change, osMajorVersion, selectedSystems }) =
...
+                        }, {
+                            ...Columns.OperatingSystem,
+                            props: {},
+                            sortBy: ['osMajorVersion', 'osMinorVersion']
```
vkrizan, line 79: Observation: DRY.The sorting array for common coulumns would likely be repeated wherever the column is used.

vkrizan (commented):
So far so good.

The sorting works only in CreatePolicy, in EditPolicy it only gives the ability, and other system lists on Systems page and Policy's systems do not have any sorting enabled at all.

src/SmartComponents/SystemsTable/hooks.js
bastilian: Remnant from earlier development.

src/SmartComponents/SystemsTable/hooks.js
bastilian: That is the default behaviour of the Inventory table. We can bring it up with UX and the platform team.

https://user-images.githubusercontent.com/7757/122388071-cf3af880-cf6f-11eb-93f1-555c78991bec.mov

bastilian:
@vkrizan Thank you for taking a look and reviewing this! RHICOMPL-1657 is only about adding sorting for the SystemsTable in the Wizard, that it was showing up on the EditPolicy was a bug with the columns being cached in the Inventory store. We will be adding sorting to the other SystemsTables and for all columns too, but in time. For the EditPolicy we also would want to be able and not only filter but also sort selected systems first.

akofink (approved):
Excellent! Works as expected. Any reason we shouldn't enable this on EditPolicy, PolicyDetails, and the Systems page as well?

bastilian:
@akofink Thanks for taking a look as well! We can consider adding it to the Edit Policy, maybe the PolicyDetails, since they have the same columns and all columns in the table would be sortable. On the other tables I left them out to avoid a mix of sortable and none-sortable columns, as we are for now only able to sort by name and OS versions, while columns using other attributes would still be static.

akofink:
Ah, makes sense! Yes, it'd be confusing if only some columns were sortable.

PR #1269, opened 2021-06-07
fix(FilterConfig): Only call buildFilterString without array

bastilian:
The filter string should only be attempted if no array is set.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/Utilities/hooks/useFilterConfig.js
+++ b/src/Utilities/hooks/useFilterConfig.js
@@ -65,7 +65,7 @@ const useFilterConfig = (initialConfig, arrayToFilter) => {
         filtered: filteredArray,
         activeFilters,
         activeFilterValues,
-        filterString: filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters)
+        filterString: !arrayToFilter ? filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters) : ''
     };
 };
 


src/Utilities/hooks/useFilterConfig.js
```diff
@@ -65,7 +65,7 @@ const useFilterConfig = (initialConfig, arrayToFilter) => {
         filtered: filteredArray,
         activeFilters,
         activeFilterValues,
-        filterString: filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters)
+        filterString: !arrayToFilter ? filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters) : ''
```
vkrizan, line 68: Can you please elaborate on how the `arrayToFilter` interplays with this?

bastilian: If `useFilterConfig` is given an array it can apply active filters to it, if not it can return a string the is compiled of all the `filterString` function returns provided by the filter configuration and this we do not need when an array is given, as the filters in the configuration won't have a `filterString`, but a `filter` function.

vkrizan (approved):
Thanks.

PR #1271, opened 2021-06-08
fix(TableTools): RHICOMPL-942 - Sort items not orderArray

bastilian:
Sorting the `orderArray` instead of the array with items, apparently didn't work. 
I'm unsure why it doesn't work anymore or why it worked before. 🤷 

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/Utilities/helpers.js
+++ b/src/Utilities/helpers.js
@@ -40,11 +40,12 @@ export const orderArrayByProp = (property, objects, direction) => (
     })
 );
 
-export const orderByArray = (objectArray, orderProp, orderArray, direction) => (
-    (direction !== 'asc' ? orderArray.reverse() : orderArray).flatMap((orderKey) => (
+export const orderByArray = (objectArray, orderProp, orderArray, direction) => {
+    const sortedObjectArray = orderArray.flatMap((orderKey) => (
         objectArray.filter((item) => (item[orderProp] === orderKey))
-    ))
-);
+    ));
+    return (direction !== 'asc' ? sortedObjectArray.reverse() : sortedObjectArray);
+};
 
 export const getProperty = (obj, path, fallback) => {
     const parts = path.split('.');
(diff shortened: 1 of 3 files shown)


src/Utilities/helpers.js
```diff
@@ -40,11 +40,12 @@ export const orderArrayByProp = (property, objects, direction) => (
...
-export const orderByArray = (objectArray, orderProp, orderArray, direction) => (
-    (direction !== 'asc' ? orderArray.reverse() : orderArray).flatMap((orderKey) => (
+export const orderByArray = (objectArray, orderProp, orderArray, direction) => {
+    const sortedObjectArray = orderArray.flatMap((orderKey) => (
         objectArray.filter((item) => (item[orderProp] === orderKey))
-    ))
-);
+    ));
+    return (direction !== 'asc' ? sortedObjectArray.reverse() : sortedObjectArray);
+};
```
akofink, line 48: This looks like a pure refactor. Mind pointing out what the difference is? :smiley_cat:

akofink (commented):
Is this supposed to fix the severity sort in the system details rules table? It seems to work fine for me already on master :confused:

src/Utilities/helpers.js
vkrizan: The `orderArray` is getting the items already reversed, hence the reason why the sorting keeps the same (the reversing zeros it out).

Also a note on the sorting implementation: This does N passes to filter out items in order of what comes from the `orderArray`. Can it do one pass using sort leveraging index?

bastilian: > The orderArray is getting the items already reversed

Where do you see the reversion happen outside of `orderByArray`? 
Both `useTableSort` implementations just pass whatever is configured in the column for `sortByArray` without touching it.

vkrizan: The `reverse` is in-place, so it might come from the previous function call.

vkrizan (approved):
Works as expected.

akofink:
> this is what I see (in FireFox)

Ah! I also see this issue within the TabbedRules component. I was looking at the system details page, which seems to work as expected on master.

akofink (approved):
Confirmed, fixes the issue! Thanks, @bastilian!

PR #1297, opened 2021-06-24
feat(FilterConfig): Use TableTools useFilterConfig

bastilian:
This updates the useFilterConfig hook in the TableTools and underlying supporting classes to the latest changes from the `inventory-compliance` package to allow migrating the `InventoryTable` and `ReportsTable` the currently use the one from `inventory-compliance` to use the TableTools version and simplify the migration in #1205.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


akofink (commented):
The code seems reasonable. I noticed a regression with the 'Selected only' switch:

image

src/SmartComponents/SystemsTable/__snapshots__/InventoryTable.test.js.snap
```diff
@@ -404,13 +388,7 @@ exports[`InventoryTable returns with compliantFilter 1`] = `
...
         ]
       }
-      activeFilters={
-        Object {
-          "compliancescore": Array [],
-          "compliant": Array [],
-          "name": "",
-        }
-      }
+      activeFilters={Object {}}
```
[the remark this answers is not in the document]
bastilian: Actually, we don't need the `activeFilters` prop at all as it is now part of the `conditionalFilter`, which are the toolbarProps returned by the `useFilterConfig` hook and spread as `FECInventoryTable` props.

src/Utilities/hooks/useTableTools/useFilterConfig.js
```diff
@@ -2,16 +2,30 @@ import { useState, useEffect, useMemo } from 'react';
...
 import useSelectedFilter from './useSelectedFilter';
 
+const filterValues = (activeFilters) => (
+    Object.values(activeFilters).filter((value) => {
+        if (typeof value === Object) {
+            return Object.values(value).length > 0;
+        }
+
+        if (typeof value === Array) {
+            return value?.length > 0;
```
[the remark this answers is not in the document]
bastilian: Done.

bastilian:
@akofink The issues with the "Selected Only" filter is addressed, the `useMemo` updating the memoized value on a prop change caused the issue. It's now just a `const`, which works, but means that all filter configurations use one builder, which is not ideal, as we have to be careful to not have the build configurations mixed up, which was part of the problem that using `useMemo` caused.

akofink:
I think I found another regression with the host OS filter:

image

bastilian:
> Is there a way we can write a test that covers this?

Yeah, I'll add a test helper that we can add to any table and verify that they render all filters correctly.

bastilian:
@akofink The issue with the OS filter is fixed now. There is now also a `toHaveFiltersFor` test matcher that is used on all tables to valid their filters are rendered correctly. It also interacts with the component and queries elements rather than taking a snapshot.

config/setupTests.js
```diff
@@ -1,5 +1,5 @@
 import 'jest-canvas-mock';
-import { configure, mount, render, shallow } from 'enzyme';
+import { configure, mount, shallow } from 'enzyme';
```
bastilian, line 2: We don't appear to use `render` anywhere in the tests,  I removed it to avoid conflicts with `render` from the `react-testing-library`.

src/SmartComponents/SystemsTable/InventoryTable.test.js
```diff
@@ -11,10 +20,26 @@ jest.mock('react-redux', () => ({
...
     useDispatch: () => (
         jest.fn()
-    ),
-    useStore: jest.fn(() => ({
-        getStore: jest.fn({})
-    }))
+    )
+}));
+
+const InventoryTableMock = (props) => (
```
bastilian, line 26: The `InventoryTable` is imported via chrome, which means it is not available in the test.
We could try to import it directly in the tests, but for now, for the purpose of giving the filter tests a `PrimaryToolbar` to interact and test this works. The `PrimaryToolbar` is the same that `InventoryTable` would use to render the filters.

src/PresentationalComponents/RulesTable/RulesTable.test.js
```diff
@@ -30,4 +40,15 @@ describe('RulesTable', () => {
 
         expect(toJson(wrapper)).toMatchSnapshot();
     });
+
+    it('expect to have filters properly rendered', () => {
+        const filterConfig = buildFilterConfig({
+            showPassFailFilter: false,
+            remediationAvailableFilter: false
+        }).filter((filter) => (filter.label !== 'Severity'));
```
bastilian, line 48: The severity filter has React components in its label, which causes issues with querying it. I tried somehow parsing out only the text, but that didn't work properly either, so I exempt it from testing for now.

src/SmartComponents/SystemsTable/InventoryTable.test.js
akofink: I think this is a fine way to mock it without having to import it.

akofink (commented):
I'm still seeing some issues with the selected only switch. Notice the table below is showing all 1240 rules, not just the 277 selected:

image

And... you know what's coming by now :laughing: Can we cover this with a test?

src/Utilities/hooks/useTableTools/useFilterConfig.js
```diff
@@ -2,18 +2,33 @@ import { useState, useEffect, useMemo } from 'react';
...
+
+        if (typeof value === Array) {
+            return value.length > 0;
+        }
+
+        return !!value;
+    })
+);
+
+const filterConfigBuilder = new FilterConfigBuilder([]);
```
bastilian, line 19: @vkrizan The issue here was that even with setting `filterConfig` (or probably even, because of) as a dependency for the useMemo it was possible that functions were talking to a different instance of `filterConfigBuilder`.

The FilterConfigBuilder has evolved over time and we should really be rewriting away from classes.

vkrizan: Can it be kept within `useFilterConfig` and be memoized for each instance of the hook? `useRef` might do the trick. Two TableToolsTables should not collide.

bastilian: > useRef might do the trick

How would that help?

bastilian: @vkrizan Also, it was memoized within useFilterConfig before, and it has caused more confusion even with a proper dependency to invalidate the memoized builder. 

As I said we have been abusing the builder for things it was not meant for, and rather than trying to adopt code to make it work within, we should rather adopt the builder instead.

vkrizan: `useRef` or `useMemo` (as used before) will keep a single object instance for each hook instance separately. The highlighted line changed it to a module's global. If there are two uses of `useFilterConfig`, they should not share the same config builder instace, but rather have their own instance each.

bastilian: @vkrizan again, it already had been implemented using `useMemo`, and it did not have the desired effect and caused issues.

vkrizan: This one works for me:
```
const filterConfigBuilder = useMemo(() => (new FilterConfigBuilder(filterConfig)), []);
```

Wasn't the issue with not passing the initial config from `filterConfig`?

vkrizan:
Breaks with:

Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=object&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```
div
div
ToolbarItem@https://ci.foo.redhat.com:1337/apps/compliance/js/pfVendor.js:22358:113
div
ToolbarGroupWithRef@https://ci.foo.redhat.com:1337/apps/compliance/js/pfVendor.js:22305:1
div
div
ToolbarContent@https://ci.foo.redhat.com:1337/apps/compliance/js/pfVendor.js:22087:9
div
Toolbar@https://ci.foo.redhat.com:1337/apps/compliance/js/pfVendor.js:21923:9
PrimaryToolbar@https://ci.foo.redhat.com:1337/apps/compliance/js/rhcsVendor.js:8486:83
TableToolsTable@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:4233:20
RulesTable@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2693:16
StateViewPart@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2808:18
StateView@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2816:21
StateViewWithError@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2826:21
ProfileTabContent@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:3242:17
section
c@https://ci.foo.redhat.com:1337/apps/chrome/js/141.f68e2ddb1f49aeeface6.js:2:66141
u@https://ci.foo.redhat.com:1337/apps/chrome/js/141.f68e2ddb1f49aeeface6.js:2:67487
RoutedTabs@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:3072:18
TabbedRules@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:3370:18
StateViewPart@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2808:18
StateView@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2816:21
StateViewWithError@https://ci.foo.redhat.com:1337/apps/compliance/js/src_PresentationalComponents_index_js.js:2826:21
EditPolicyProfilesRules@https://ci.foo.redhat.com:1337/apps/compliance/js/EditPolicy.js:643:16
```

bastilian:
@vkrizan Just pushed a fix for that. The issue was the rendering of the dedicated actions.

bastilian:
@akofink The issue you were seeing with the selected filter should be fixed and there are now basic tests for it.

src/Utilities/hooks/useTableTools/useFilterConfig.js
vkrizan: The issue with the global builder, is that there cannot be different filter configs. For example on TabbedRules, if each tab has a different config, it would not work as one would override the other.

bastilian: @vkrizan This issue is prevented by the useEffect in the useFilterConfig as well as all calls setting the config.

bastilian: > Here is a test:

I'm not sure what this test is about?
To me this looks correct. The test is creating two components one with 4 and the other with 6 filters?

vkrizan: The output of the two components, respectively size of the `filterConfig` items should be 4 and 6. However the result is 4 and 4.

bastilian: @vkrizan If you run this test with the current changes, the result will be 3 and 4, which is correct. The first has the first two elements sliced, which means it will pass on 4 filter items, however, one of these filters is unknown, so minus 1, the result 3. the second passes in all 5 filters, also including the unknown filter, also minus 1, results in 4.

bastilian: Oh, also there is one hidden filter in those somewhere, which also won't be in the resulting filterconfig

akofink: Agree, here is the snapshot I see written:

```diff
--- a/src/Utilities/hooks/useTableTools/__snapshots__/useFilterConfig.test.js.snap
+++ b/src/Utilities/hooks/useTableTools/__snapshots__/useFilterConfig.test.js.snap
@@ -270,3 +270,10 @@ Object {
   },
 }
 `;
+
+exports[`useFilterConfig works concurrently 1`] = `
+Array [
+  "3",
+  "4",
+]
+`
```

Maybe this would be a good test to add regardless? @vkrizan I'm surprised you're seeing a different result :thinking: It wouldn't hurt to let Jenkins give the final word.

bastilian: @vkrizan @akofink the test is now added.

vkrizan: An interesting results are seen when there are different filters used:
```
// import filters, { exampleFilters } from './__fixtures__/filters';
act(() => {
    root = create(
        <>
            <Component counter='1' key='2' filters={ filters }></Component>
            <Component counter='1' key='1' filters={ exampleFilters }></Component>
        </>
    )
});
```

Also the chipBuilder is memoized on the FilterConfigBuilder instance, which would use the first filter config that it was used with.

bastilian: @vkrizan Do you actually see any of these issues manifesting in the UI and breaking functionality?

akofink (commented):
Hey, @bastilian! Sorry for the delay. I was testing this out with David's backend score fix PR and noticed a regression in the report details table, not related to the backend work. I think it's probably something silly because no filters work there at all; the API requests do not contain any filter strings other than the default (policy_id and has_test_results). Oddly, I see two InventoryTable components on the page, and only the first one is getting the correct state for its filterConfig but the other does not. Any ideas?

image

bastilian:
@akofink The issue with filter string should be fixed and the filters should work properly. I can not however reproduce two InventoryTable components.

src/Utilities/hooks/useTableTools/useFilterConfig.js
bastilian: @vkrizan Can you, right now, without any changes, see any issues in the UI?

vkrizan:
The selected filter stopped working. Shows also non-selected rules.

bastilian:
@vkrizan Thank you for testing and reviewing! That should be fixed now and the Selected filter should once again, work as expected. I should have caught that, because a test actually failed because of that.

src/Utilities/hooks/useTableTools/useFilterConfig.js
vkrizan: I see that the the case was fixed with the recent changes.
It works thanks to the shared memory reset on each use, like here
```
    # FilterConfigBuilder
    getChipBuilder = (config) => {
        if (config) {
            this.config = config;
        }
        ...
```

Wouldn't it be more convenient to make the builders stateless, rather than resetting their memory on each use? There is a slight chance of racing with that memory usage.

bastilian: @vkrizan As I said before that whole way we use the filterbuilder is not ideal, because it was not meant for this use, but slowly adopted over time and a good rewrite would be in order to make it more convenient to use, but as long as there is only a slight and not definitive chance of things breaking, it has to do.

src/PresentationalComponents/ReportsTable/ReportsTable.js
```diff
@@ -42,12 +42,14 @@ const ReportsTable = ({ profiles }) => {
...
     const operatingSystems = uniq(profiles.map(({ majorOsVersion }) => (majorOsVersion)).filter((i) => (!!i)));
-    const { conditionalFilter, filtered: filteredProfiles } = useFilterConfig([
+    const { toolbarProps: conditionalFilter, filter } = useFilterConfig({ filters: { filterConfig: [
         ...policyNameFilter,
         ...policyTypes.length > 0 && policyTypeFilter(policyTypes) || [],
         ...operatingSystems.length > 0 && operatingSystemFilter(operatingSystems) || [],
         ...policyComplianceFilter
-    ], profiles);
+    ] } });
+    const filteredProfiles = filter ? filter(profiles) : [];
```
vkrizan, line 51: When would the false condition happen?

vkrizan (approved):
Works well. Thanks.

akofink (approved):

src/PresentationalComponents/ReportsTable/__snapshots__/ReportsTable.test.js.snap
```diff
@@ -15,7 +15,7 @@ exports[`ReportsTable expect to render without error 1`] = `
           Object {
             "filterValues": Object {
               "onChange": [Function],
-              "value": "",
+              "value": undefined,
```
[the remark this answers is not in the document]
bastilian: The config wasn't passed to the defaults init, it is now. These "defaults" are pretty redundant, when I think of it, they just return an empty value of whatever type it is, but have no real effect on the rendering as far as I know.

src/PresentationalComponents/ReportsTable/ReportsTable.js
bastilian: This was probably from initial migration to the new useFilterConfig, it would not happen actually. The condition is removed.

PR #1301, opened 2021-07-01
fix(Policies): RHICOMPL-1818- Add notifications for policy actions

bastilian:
This adds notifications for creating and updating policies:

Screenshot 2021-07-01 at 14 03 33

Screenshot 2021-07-01 at 14 02 57

Screenshot 2021-07-01 at 14 01 16

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
+++ b/src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
@@ -72,6 +72,7 @@ export const FinishedCreatePolicy = ({
             dispatchNotification({
                 variant: 'success',
                 title: `Policy created`,
+                autoDismiss: true,
                 description: <span>
                     From the <strong>SCAP Policies</strong> page, open { name }.
                 </span>
--- a/src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
+++ b/src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
@@ -11,6 +11,7 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import { withApollo } from '@apollo/client/react/hoc';
 import { usePolicy } from 'Mutations';
+import { dispatchNotification } from 'Utilities/Dispatcher';
 
 const EmtpyStateWithErrors = ({ errors }) => (
     (errors && Array.isArray(errors) && errors.length > 0) ? (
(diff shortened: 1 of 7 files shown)


src/SmartComponents/CreatePolicy/FinishedCreatePolicy.js
```diff
@@ -68,10 +69,22 @@ export const FinishedCreatePolicy = ({
         updatePolicy(null, newPolicy, onProgress).then(() => {
             setPercent(100);
             setMessage();
+            dispatchNotification({
```
bastilian, line 72: I tried extracting this similarly to what is in EditPolicy, to also be able and test it nicely, but it broke a few other things that would have needed more work to get right. 

Generally we should look into changing and moving the updatePolicy call here to a more appropriate place and not run as an effect of a component.

akofink (commented):
Editing a policy actually sends a CreateProfile mutation now :/ It looks like there might be multiple issues:

image

bastilian:
@akofink Sorry, I overlooked this. It snuck in when I was dealing with issues in the tests and only appeared with some policies and not all. It should be fixed, the issue was `useCallback`, which didn't have the correct values to pass on.

akofink:
I'm sometimes getting LOTS of notifications when performing a single update action:

image

I'm also seeing a toast notification every time I re-open the modal. In the following screenshot, I deleted a policy from the policies page, saw the toast notification, waited for it to disappear, then used the kebab dropdown to edit another policy, and the old delete notification popped up again. Closing and re-opening the modal makes the notification show every time it's reopened:

image

Same thing if you create a new policy, then edit another policy with the kebab:

image

bastilian:
@akofink Thanks for testing! I could only replicate the issue with extra notifications a handful of times. However, I think what may have happened is that notifications accumulated in the store somehow and got fired with any further notification. It now has `autoDismiss` set, this should reduce the chances of that happening hopefully.

akofink (commented):
It works fairly well for the most part, but I'm still seeing notifications appear again after I open the edit policy modal after editing or creating a policy. The weirdness seems to get worse with error notifications. I can show you live tomorrow and see if you are able to reproduce it as well.

https://user-images.githubusercontent.com/761923/125700741-b1857f8d-0ca3-4a73-aec6-5e226bf5b18e.mp4

akofink (approved):
After meeting to discuss/review the issue, it turned out to be caused by the redux dev tools! Disabling the extension makes this work as expected. Thanks, @bastilian!

PR #1204, opened 2021-05-04
chore(InventoryTable): Update column usage

bastilian:
Then `InventoryTable` component is going to change the way it needs columns to be set[0] and allows applications to remove custom actions from the `entitiesReducer` related to columns. This PR is to prepare the compliance-frontend for this change.

This PR can already be tested with the latest LOCAL_CHROME or with the chrome from `ci-beta`.

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

[0] https://github.com/RedHatInsights/frontend-components/pull/1105

PR #1306, opened 2021-07-14
feat(SystemsTable): Enable tags and tag filtering

bastilian:
This adds the capability to filter systems by tags. This is dependent on https://github.com/RedHatInsights/compliance-backend/pull/927 for the filtering to be actually effective. 

Screenshot 2021-07-28 at 13 35 09

Can be enabled via: https://qa.foo.redhat.com:1337/insights/compliance/systems?tags=enable

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/Columns.js
+++ b/src/SmartComponents/SystemsTable/Columns.js
@@ -136,3 +136,5 @@ export const OperatingSystem  = compileColumnRenderFunc({
     ),
     cell: OperatingSystemCell
 });
+
+export const inventoryColumn = (column) => (column);
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -84,6 +84,22 @@ export const useFetchSystems = ({
     );
 };
 
+const buildApiFilters = (filters = {}) => {
+    const { tagFilters, ...otherFilters } = filters;
+    const tagsApiFilter = tagFilters ? {
+        tags: tagFilters.flatMap((tagFilter) => (
+            tagFilter.values.map((tag) => (
+                `${ encodeURIComponent(tagFilter.key) }/${ encodeURIComponent(tag.tagKey) }=${ encodeURIComponent(tag.value) }`
+            ))
+        ))
+    } : {};
+
+    return {
+        ...otherFilters,
+        ...tagsApiFilter
+    };
+};
+
 export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
     const appendDirection = (attributes, direction) => (
         attributes.map((attribute) => `${attribute}:${direction}`)
@@ -97,9 +113,10 @@ export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
         const sortableColumn = findColumnByKey(orderBy);
         const sortBy = sortableColumn && sortableColumn.sortBy ?
             appendDirection(sortableColumn.sortBy, orderDirection) : undefined;
+        const filterForApi = buildApiFilters(filters);
 
         const fetchedEntities = await fetchEntities(perPage, page, {
-            filters,
+            ...filterForApi,
             sortBy
         });
         const { entities, meta: { totalCount } } = fetchedEntities || {};
@@ -251,3 +268,26 @@ export const useSystemsExport = ({
 
     return exportConfig;
 };
+
+export const useTags = (tagsEnabled) => (
+    tagsEnabled ? {
+        props: {
+            hideFilters: {
+                name: true,
+                tags: false,
+                registeredWith: true,
+                stale: true
+            },
+            showTags: true
+        }
+    } : {
+        props: {
+            hideFilters: {
+                name: true,
+                tags: true,
+                registeredWith: true,
+                stale: true
+            }
+        }
+    }
+);
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -44,6 +44,7 @@ export const ComplianceSystems = () => {
                                     showLink: true,
                                     showOsInfo: true
                                 }),
+                                Columns.inventoryColumn('tags'),
                                 Columns.Policies,
                                 Columns.DetailsLink
                             ]}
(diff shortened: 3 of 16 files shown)


skateman:
@bastilian how can I have the list of possible tags in the filter's dropdown?

bastilian:
@skateman Not 100% sure what you mean. Those are coming from the Inventory API/component.

skateman:
> @skateman Not 100% sure what you mean. Those are coming from the Inventory API/component.

Screenshot from 2021-07-23 14-29-54

bastilian:
@skateman If no host has any tags it will probably be empty, but I'm not sure how tags are or can be added (via insights-client).

skateman:
@bastilian there's one host I created with a tag, but it might not be enough :thinking:

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -251,3 +271,37 @@ export const useSystemsExport = ({
...
 };
+
+export const useTags = (columns) => {
+    const tagsEnabled = useFeature('tags');
+    const tagsColumn = {
+        key: 'tags',
+        title: 'Tags',
+        props: { width: 10, isStatic: true },
+        // eslint-disable-next-line react/display-name
+        renderFunc: (value, systemId) => <TagWithDialog count={ value.length } systemId={ systemId } />
```
bastilian, line 282: This doesn't load very efficiently and we should rather be using the default tags column defined in the Inventory.
To do this we will need to implement "column as a function" in order to merge default and custom columns[0].

[0] https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory/doc/props.md#columns-as-function

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -84,6 +86,23 @@ export const useFetchSystems = ({
     );
 };
 
+const buildApiFilters = (filters = {}) => {
+    const { tagFilters, ...otherFilters } = filters;
+    const tagsApiFilter = tagFilters ? {
+        tags: tagFilters.flatMap((tagFilter) => {
+            const namespace = tagFilter.key;
+            return tagFilter.values.map((tag) => (
+                `${ namespace }/${ tag.tagKey }=${ tag.value }`
```
skateman, line 95: I guess this is the important piece :+1: :medal_military:

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -251,3 +271,37 @@ export const useSystemsExport = ({
 
     return exportConfig;
 };
+
+export const useTags = (columns) => {
+    const tagsEnabled = useFeature('tags');
+    const tagsColumn = {
+        key: 'tags',
+        title: 'Tags',
```
vkrizan, line 279: The column renders a big Spinner in my instance
Screenshot 2021-07-29 at 12-35-05 Systems - Compliance Red Hat Insights

src/SmartComponents/SystemsTable/hooks.js
bastilian: @vkrizan This is passed via variable in a post body, there should be no need to encode.

src/SmartComponents/SystemsTable/hooks.js
bastilian: @vkrizan The `TagWithDialog` component is async loaded, each one individually. I'm currently adding and finishing the changes from https://github.com/RedHatInsights/compliance-frontend/pull/1204, as that will allow us to use default columns, including the default Tags column, which will then be loaded async with the InventoryTable.

src/SmartComponents/SystemsTable/hooks.js
skateman: @bastilian it is true, but you need double escaping in case you have `/` or `=` symbols inside the namespace/tag/value. Otherwise we would not know which are the delimiters and which are just inside symbols.

vkrizan: It is a composite string, made out of three components. Any of the component can contain characters that would create an unexpected results when parsing. Hence the need for encoding. If it were to be sent via separate keys, then the encoding would not be needed.

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -136,3 +136,5 @@ export const OperatingSystem  = compileColumnRenderFunc({
     ),
     cell: OperatingSystemCell
 });
+
+export const inventoryColumn = (column) => (column);
```
bastilian, line 140: It's not much, but it makes things just a little nicer.

bastilian:
@vkrizan @skateman Thanks both for taking a look! The column async loading is now fixed and we use the default column provided. And the variable values are now encoded as well.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -84,6 +84,23 @@ export const useFetchSystems = ({
     );
 };
 
+const buildApiFilters = (filters = {}) => {
+    const { tagFilters, ...otherFilters } = filters;
+    const tagsApiFilter = tagFilters ? {
+        tags: tagFilters.flatMap((tagFilter) => {
+            const namespace = encodeURIComponent(tagFilter.key);
+            return tagFilter.values.map((tag) => (
+                `${ namespace }/${ encodeURIComponent(tag.tagKey) }=${ encodeURIComponent(tag.value) }`
```
skateman, line 93: Nit: it took me some time to find that the namespace is encoded elsewhere :thinking:

bastilian: Put it into the string. :)

skateman:
I've followed the link to enable the tags, but I'm still not able to see them in the dropdown :confused: 
Screenshot from 2021-08-05 15-21-58

Also when clicking on the tags icon in the list, I'm getting an empty page:
Screenshot from 2021-08-05 15-22-41

Am I doing something wrong?

bastilian:
@skateman Are the Tags in the inventory? That is where the list of Tags for the dropdown is pulled from.

skateman (approved):
The Seal of Approval

PR #1307, opened 2021-07-22
fix(SystemsTable): RHICOMPL-734 - Use proper bulk select

bastilian:
Screenshot 2021-07-22 at 14 03 08

**Note:**
This is the best we can do for now and there is still a chance for odd behaviour. I tried a few approaches, but until we are able to make a selection without having to load and/or process additional data for the select callbacks this will be hard to resolve. We need additional data for some callbacks, for example to validate/make the rule selection. 

For “Select All” it will in most cases always need to query more information and stall the selection process until the requests are all done and it can pass on the full system objects to the callback.

For now I believe this is fine to introduce a “Select All” and create a follow up task to start and look into optimising it, step by step.

The first step would be to make the callbacks not depend on more information than the ID and request the data there instead to make it independent from the actual selecting, so that clicking on a checkbox and it being selected is instant, without any delay, at least for most selection operations.

In case of the “Selec All” it will still be required for the bulk select to query for the IDs not yet loaded. So, another followup step should be to remove the need for that. But I’m not sure what the right approach for that would be. A first Idea would be to allow selections to be filters. For example on loading the table first without any additional “table”-filters selected it would pass on a “selection” indicating that all systems matching this account are selected.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -7,6 +7,7 @@ import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
 import { osMinorVersionFilter, GET_MINIMAL_SYSTEMS } from './constants';
 import useExport from 'Utilities/hooks/useTableTools/useExport';
 import { useBulkSelect } from 'Utilities/hooks/useTableTools/useBulkSelect';
+import { dispatchNotification } from 'Utilities/Dispatcher';
 
 const groupByMajorVersion = (versions = [], showFilter = []) => {
     const showVersion = (version) => {
@@ -242,7 +243,14 @@ export const useSystemBulkSelect = ({
         const idFilter = toIdFilter(fetchIds);
         const results = await fetchBatched(fetchSystems, fetchIds.length, {
             ...idFilter && { filter: idFilter }
+        }).catch((error) => {
+            dispatchNotification({
+                variant: 'danger',
+                title: 'Error selecting systems',
+                description: error.message
+            });
         });
+
         return results.flatMap((result) => (
             result.entities
         ));
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -235,6 +235,10 @@ export const useSystemBulkSelect = ({
     });
 
     const fetchFunc = async (fetchIds) => {
+        if (fetchIds.length === 0) {
+            return [];
+        }
+
         const idFilter = toIdFilter(fetchIds);
         const results = await fetchBatched(fetchSystems, fetchIds.length, {
             ...idFilter && { filter: idFilter }
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -4,8 +4,9 @@ import { useDispatch } from 'react-redux';
 import debounce from '@redhat-cloud-services/frontend-components-utilities/debounce';
 import useCollection from 'Utilities/hooks/api/useCollection';
 import { systemsWithRuleObjectsFailed } from 'Utilities/ruleHelpers';
-import { osMinorVersionFilter } from './constants';
+import { osMinorVersionFilter, GET_MINIMAL_SYSTEMS } from './constants';
 import useExport from 'Utilities/hooks/useTableTools/useExport';
+import { useBulkSelect } from 'Utilities/hooks/useTableTools/useBulkSelect';
 
 const groupByMajorVersion = (versions = [], showFilter = []) => {
     const showVersion = (version) => {
@@ -84,6 +85,13 @@ export const useFetchSystems = ({
     );
 };
 
+const fetchBatched = (fetchFunction, total, filter, batchSize = 100) => {
+    const pages = Math.ceil(total / batchSize) || 1;
+    return Promise.all([...new Array(pages)].map((_, pageIdx) => (
+        fetchFunction(batchSize, pageIdx + 1, filter)
+    )));
+};
+
 const buildApiFilters = (filters = {}) => {
     const { tagFilters, ...otherFilters } = filters;
     const tagsApiFilter = tagFilters ? {
@@ -124,63 +132,12 @@ export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
         return {
             results: entities.map((entity) => ({
                 ...entity,
-                selected: (selected || []).map((item) => (item.id)).includes(entity.id)
+                selected: (selected || []).map((id) => (id)).includes(entity.id)
             })),
             orderBy,
             orderDirection,
             total: totalCount
         };
-    };
-};
-
-export const useOnSelect = (onSelectProp, items, preselectedSystems, total) => {
-    const [selectedSystems, setSelected] = useState(preselectedSystems);
-    const selectedSystemIds = selectedSystems.map((system) => (system.id));
-    const isPageSelected = items.filter((item) => (
-        selectedSystemIds.includes(item.id)
-    )).length === items.length;
-
-    const onSelectCallback = (selected) => (
-        onSelectProp && onSelectProp(selected)
-    );
-
-    const onSelect = (_event, select, _index, row) => {
-        const system = items.find(({ id }) => id === row.id);
-        if (!system) {
-            console.error(`System identifed as ${row.id} not found in items for selection!`);
-            return;
-        }
-
-        const selected = select
-            ? [...selectedSystems, system]
-            : selectedSystems.filter((selected) => (selected.id !== row.id));
-
-        setSelected(selected);
-        onSelectCallback(selected);
-    };
-
-    const onBulkSelect = () => {
-        const selectedSystemsWithoutItems = (selectedSystems || []).filter((system) => (
-            !(items.map((item) => (item.id))).includes(system.id)
-        ));
-        const notAllSelected = selectedSystems.length <= total;
-        const selected = notAllSelected ? (
-            !isPageSelected ? [...selectedSystemsWithoutItems, ...items] : selectedSystemsWithoutItems
-        ) : [];
-
-        setSelected(selected);
-        onSelectCallback(selected);
-    };
-
-    useEffect(() => {
-        setSelected(preselectedSystems);
-    }, [preselectedSystems]);
-
-    return {
-        onSelect,
-        onBulkSelect,
-        selectedSystems,
-        isPageSelected
     };
 };
 
@@ -229,32 +186,25 @@ const toIdFilter = (ids) => (
 );
 
 export const useSystemsExport = ({
-    columns, filter, policyId, query, selected, total
+    columns, selected, total, fetchArguments
 }) => {
     const selectionFilter = selected ? toIdFilter(selected) : undefined;
     const fetchSystems = useFetchSystems({
-        query,
+        query: fetchArguments.query,
         variables: {
+            ...fetchArguments.variables,
             filter: selectionFilter ?
-                `${ filter } and (${ selectionFilter })` : filter,
-            ...policyId && { policyId }
+                `${ fetchArguments.variables.filter } and (${ selectionFilter })` :
+                fetchArguments.variables.filter
         }
     });
-
-    const fetchBatched = (total, filter) => {
-        const BATCH_SIZE = 100;
-        const pages = Math.floor(total / BATCH_SIZE) + 1;
-        return Promise.all([...new Array(pages)].map((_, pageIdx) => (
-            fetchSystems(BATCH_SIZE, pageIdx + 1, filter)
-        )));
-    };
 
     const selectedFilter = () => (
         selected?.length > 0 ? toIdFilter(selected) : undefined
     );
 
     const exporter = async () => {
-        const fetchedItems = await fetchBatched(total, selectedFilter());
+        const fetchedItems = await fetchBatched(fetchSystems, total, selectedFilter());
         return fetchedItems.flatMap((result) => (
             result.entities
         ));
@@ -267,6 +217,69 @@ export const useSystemsExport = ({
     });
 
     return exportConfig;
+};
+
+export const useSystemBulkSelect = ({
+    total,
+    onSelect,
+    preselected,
+    fetchArguments,
+    currentPageIds,
+    systemsCache = []
+}) => {
+    // This is meant as a compatibility layer and to be removed
+    const [selectedSystems, setSelectedSystems] = useState([]);
+    const fetchSystems = useFetchSystems({
+        ...fetchArguments,
+        query: GET_MINIMAL_SYSTEMS
+    });
+
+    const fetchFunc = async (fetchIds) => {
+        const idFilter = toIdFilter(fetchIds);
+        const results = await fetchBatched(fetchSystems, fetchIds.length, {
+            ...idFilter && { filter: idFilter }
+        });
+        return results.flatMap((result) => (
+            result.entities
+        ));
+    };
+
+    const cachedOrFetch = async (selectedIds) => {
+        const cachedSystems = systemsCache.filter(({ id }) => (selectedIds.includes(id)));
+        const cachedIds = cachedSystems.map(({ id }) => (id));
+        const fetchIds = selectedIds.filter((id) => (!cachedIds.includes(id)));
+        const fetchedSystems = await fetchFunc(fetchIds);
+
+        return [
+            ...cachedSystems,
+            ...fetchedSystems
+        ];
+    };
+
+    const onSelectCallback = async (selectedIds) => {
+        const systems = await cachedOrFetch(selectedIds);
+        setSelectedSystems(systems);
+        onSelect && onSelect(systems);
+    };
+
+    const itemIdsInTable = async () => {
+        const results = await fetchBatched(fetchSystems, total);
+        return results.flatMap((result) => (
+            result.entities.map(({ id }) => (id))
+        ));
+    };
+
+    const bulkSelect = useBulkSelect({
+        total,
+        onSelect: onSelectCallback,
+        preselected,
+        itemIdsInTable,
+        itemIdsOnPage: () => (currentPageIds)
+    });
+    return {
+        selectedSystems,
+        ...bulkSelect
+    };
 };
 
 export const useTags = (tagsEnabled) => (
--- a/src/Utilities/hooks/useTableTools/__snapshots__/useBulkSelect.test.js.snap
+++ b/src/Utilities/hooks/useTableTools/__snapshots__/useBulkSelect.test.js.snap
@@ -215,7 +215,7 @@ Object {
   "itemId": 6,
   "name": "TEST ITEM #6",
   "rowProps": Object {
-    "selected": false,
+    "selected": true,
   },
   "severity": "low",
 }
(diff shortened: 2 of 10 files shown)


akofink (commented):
@bastilian maybe you'd be willing to walk through this with us? I'm a bit unclear about when API requests happen and why exactly they're required. I tested with 10k hosts, and it did work, though I definitely see what you mean about the lag. Also, the local devel environment is not nearly as quick when handling systems through cyndi as deployed environments, so I may be experiencing an outsized effect.

Regardless, I did notice some weirdness with the counts when using filters (please confirm it's working as expected):

https://user-images.githubusercontent.com/761923/127216843-cb886304-4819-4f93-b429-e290ad01adff.mp4

Also, I was trying to compare using Inventory's UI, and... well, they didn't implement select all! Maybe @katierik knows what's up with that:

image

bastilian:
@akofink Thank you for testing it! I'd be happy to walk through it! We can schedule a meeting or we can stay on after standup. 

> Regardless, I did notice some weirdness with the counts when using filters (please confirm it's working as expected):

Looking at the screen recording, it seems to work as expected. Unless I am missing something. 
It starts out with 9 systems selected, after changing the filters only only 11 systems out of the 10k match and are shown, out of these 11 systems only 1 is already in the 9 selected systems, and the other 8 being outside the filtered systems. 
So when selecting all the 11 systems currently shown the selection count will be 19. 

> Also, I was trying to compare using Inventory's UI, and... well, they didn't implement select all!

I'm not sure why it is not (yet) implemented in the Inventory UI, but there aren't actually many apps that have it already. 
I've only seen another implementation in Vulnerability.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -57,8 +58,8 @@ export const useFetchSystems = ({
 }) => {
     const client = useApolloClient();
 
-    return (perPage, page, requestVariables = {}) => (
-        client.query({
+    return (perPage, page, requestVariables = {}) => {
+        return client.query({
```
vkrizan, line 62: Nit: Unnecessary change to `return`.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -82,15 +81,14 @@ export const useFetchSystems = ({
...
-    );
+        });
+    };
 };
 
-const fetchBatched = (fetchFunction, total, filter) => {
-    const BATCH_SIZE = 100;
-    const pages = Math.floor(total / BATCH_SIZE) + 1;
+const fetchBatched = (fetchFunction, total, filter, batchSize = 100) => {
+    const pages = Math.floor(total / batchSize) + 1;
```
vkrizan, line 89: Didn't you meant to use `Math.ceil`?

src/Utilities/hooks/useTableTools/useBulkSelect.js
```diff
@@ -44,67 +43,34 @@ const mergeArraysUniqly = (arrayA, arrayB) => (
...
-    const filteredItems = filter ? filter(items) : items;
-    const filteredTotal = filteredItems.length;
-    const filtered = filteredTotal !== total;
-    const allFiltertedSelected = selectedIds?.length > 0 ?
-        allItemsIncluded(itemIds(filteredItems), selectedIds) : false;
-
-    const paginatedItems = paginator ? paginator(filteredItems) : filteredItems;
-    const paginatedTotal = paginatedItems.length;
-
+    const [selectedIds, setSelectedItemIds] = useState([]);
```
vkrizan, line 55: Can you please elaborate on the difference between `preselected` and `selectedIds`?

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -251,3 +201,65 @@ export const useSystemsExport = ({
...
+        const idFilter = toIdFilter(fetchIds);
+        const results = await fetchBatched(fetchSystems, fetchIds.length, {
+            ...idFilter && { filter: idFilter }
+        });
+        return results.flatMap((result) => (
+            result.entities
+        ));
+    };
+
+    const cachedOrFetch = async (selectedIds) => {
```
vkrizan, line 231: How is the cache supposed to work? It seems to keep only currently selected systems. Unselecting all clears the cache. I was wondering whether it can cache the pages with the minimal data. Or do you think it would be too much of memory used? It is used anyway on select all.

src/Utilities/hooks/useTableTools/useBulkSelect.js
```diff
@@ -115,51 +81,36 @@ const useBulkSelect = ({
...
         return currentPageSelected ?
             unselectItems(currentPageIds) : selectItems(currentPageIds);
     });
 
-    const selectFiltered = () => onSelectCallback(() => {
-        const currentFilteredIds = itemIds(filteredItems);
-        return allFiltertedSelected ?
-            unselectItems(currentFilteredIds) : selectItems(currentFilteredIds);
+    const selectAll = () => onSelectCallback(async () => {
+        const allItemIds = await itemIdsInTable();
```
vkrizan, line 99: As mentioned before, Select All is pretty heavy if there are lots of systems/pages to get data from. Could we use a spinner while it is being selected? It might be more user-friendly than hanging the page. wdyt?

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -251,3 +201,65 @@ export const useSystemsExport = ({
...
+        setSelectedSystems(systems);
+        onSelect && onSelect(systems);
+    };
+
+    const bulkSelect = useBulkSelect({
+        total,
+        onSelect: onSelectCallback,
+        preselected,
+        itemIdsInTable: async () => {
+            const results = await fetchBatched(fetchSystems, total);
```
vkrizan, line 254: There is some weird (probably race) condition: when doing "Unselect All" it does query the pages.
I'm noticing that `fetchBatched` is called twice in `fetchBatched`. Once here for `itemIdsInTable` and second time on the `onSelectCallback`. It does duplicate queries (twice the size).

src/SmartComponents/SystemsTable/hooks.js
bastilian: `Math.ceil` would in some cases cause a request of an additional page with no results. Rounding down and adding a page is safer.

vkrizan: That's exactly what should ceil prevent. What am I missing?

```
ceil(10/10) = 1
floor(10/10) + 1 = 2
ceil(0/10) = 0
floor(0/10) + 1 = 1
```

bastilian: Maybe "simpler" would have been a better description:

```js
// With Math.ceil
const pages = Math.ceil(0/100)
// we get 0, But we need 1
const pages = Math.ceil(0/100) + 1
// Now we have 1, for that case but for
const pages = Math.ceil(10/100) + 1 
// We get 2, but we want 1

// We could extract the calculation and check when it is 0 and only add it then
const calculatedPages = Math.ceil(0/100)
const pages = calculatedPages === 0 ? calculatedPages + 1 : calculatedPages;

// But with Math.floor we can stick with a simpler definition
const pages = Math.floor(0/100)
// Same thing, 0 But we need 1
const pages = Math.floor(0/100) + 1 
// Now we have 1, for that case and for
const pages = Math.floor(10/100) + 1 
// We get 1 too, and we want 1. nothing further needed
```

src/SmartComponents/SystemsTable/hooks.js
bastilian: The `systemsCache` will always only hold items from what it loaded in for the page. `cachedSystems` could be named `selectedCachedSystems`, which is used to remove any system (ids) to request from the selectedIDs.

This should work fine for now to only request what we need. We could implement a growing cache and keep any system that is ever loaded for a table,  but that could be hard to manage and might lead to inconsistencies.

src/Utilities/hooks/useTableTools/useBulkSelect.js
bastilian: We could add a spinner somewhere to indicate that something is happening, but we should consult with UX how to indicate it best.

src/Utilities/hooks/useTableTools/useBulkSelect.js
bastilian: `preselected` is a selection passed in, while `selectedIds` is the internal state.

src/SmartComponents/SystemsTable/hooks.js
bastilian: > when doing "Unselect All" it does query the pages.

This must have been a bug, it doesn't request anything anymore, and it never should have as it has always just been setting the selection to an empty array. Mind checking again and verifying that you aren't seeing a request there.

There is also now a separate query to request only the ids of systems.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: I see that the 0 creates some issue here. Can you please explain ("we get 0, But we need 1")? If there is nothing to fetch then the promise group can stay empty.

Here are cases where floor would give unexpected results:
```
Math.floor(100/100) + 1 // = 2
Math.floor(200/100) + 1 // = 3
```

If, for some reason, the 0 is problematic, a logical operator can solve it:
```
const pages = Math.ceil(total / batchSize) || 1;
```

Note that empty promise group is auto-resolved:
```
> Promise.all([])
=> Promise { <state>: "fulfilled", <value>: [] }
```

src/Utilities/hooks/useTableTools/useBulkSelect.js
vkrizan: Does that mean that there are two selection states? How does the `preselected` fit into component-lifecycle? Is it used once, or does it re-renders on a update?

src/SmartComponents/SystemsTable/hooks.js
vkrizan: Can the cache contain current page data (in addition to the ones selected)? Do you think it would have bad performance?

src/Utilities/hooks/useTableTools/useBulkSelect.js
bastilian: > Does that mean that there are two selection states? 

Yes, in a way, one in the PolicyForm and one inside the bulk select.

> How does the preselected fit into component-lifecycle?

`preselected` is set in the effect below.

src/SmartComponents/SystemsTable/hooks.js
bastilian: @vkrizan What benefit would that have?

src/SmartComponents/SystemsTable/hooks.js
bastilian: > Math.ceil(total / batchSize) || 1

Is pretty neat. It now uses this.

src/Utilities/hooks/useTableTools/useBulkSelect.js
```diff
@@ -174,14 +125,71 @@ const useBulkSelect = ({
...
+            setPageMemo(-1);
+        }
+    }, [paginatedTotal, setPageMemo]);
+
+    const { selectNone, selectedIds, ...bulkSelect } = useBulkSelect({
+        total: allCount,
+        onSelect,
+        preselected,
+        itemIdsInTable: () => {
+            return filtered ? itemIds(filteredItems) : itemIds(items);
```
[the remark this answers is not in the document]
bastilian: There probably is a way for eslint to find these and require an implicit return, but it might not make sense to enforce that, as there are situations where an implicit return could be an issue, for example with `useEffect`.

vkrizan:
Selection of a single system in Create Policy wizard leads to selection of 2 pages + 1 (101).

bastilian:
@vkrizan Fixed.

src/SmartComponents/SystemsTable/hooks.js
vkrizan: It still queries on Unselect all.

bastilian: @vkrizan Now I saw it! Thank you! This should be fixed now too.

katierik:
Inventory does not have select all implemented -- [here's that jira from UX](https://issues.redhat.com/browse/RHCLOUD-12712) and [here is a jira that talks about the issues more](https://issues.redhat.com/browse/RHCLOUD-10144).

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -251,3 +201,65 @@ export const useSystemsExport = ({
...
+    const [selectedSystems, setSelectedSystems] = useState([]);
+    const fetchSystems = useFetchSystems(fetchArguments);
+
+    const fetchFunc = async (fetchIds) => {
+        if (fetchIds.length === 0) {
+            return [];
+        }
+
+        const idFilter = toIdFilter(fetchIds);
+        const results = await fetchBatched(fetchSystems, fetchIds.length, {
```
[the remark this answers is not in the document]
akofink: I tested this case, and it's true, there's no handler. The select all action just doesn't do anything from the user's perspective. A toast notification could be added in the future.

src/Utilities/hooks/useTableTools/__snapshots__/useBulkSelect.test.js.snap
```diff
@@ -367,37 +575,37 @@ Object {
 }
 `;
 
-exports[`useBulkSelect returns to select none after all selected 2`] = `
+exports[`useBulkSelectWithItems returns to select none after all selected 2`] = `
```
akofink, line 578: This test starts with nothing selected, selects all, then selects none again. For some reason, the snap changed during this step, where select all actually shows `"checked": false` below and `"isDisabled": true` is now set. You didn't change the test, so I'm concerned :thinking: Any idea?

akofink (approved):
Seems to work well. Thanks, @bastilian! I have a test related question, and that's it. :+1:

src/Utilities/hooks/useTableTools/__snapshots__/useBulkSelect.test.js.snap
bastilian: Good catch! Async tests are hard. :D It has to "waitForNextUpdate", which is odd, because I had expected that `act` would actually do that. They now do and the results stay the same now.

src/SmartComponents/SystemsTable/hooks.js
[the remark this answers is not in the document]
bastilian: There is now a `catch`. I could only mock test it, mind trying it and see if it shows a notification in your tests as well?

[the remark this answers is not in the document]
akofink: Hmm... it doesn't seem to work quite right for me. It looks like the uncaught error comes from the promise in [useFetchSystems](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/hooks.js#L63). Below are more logs; I tried to figure it out a bit by placing an empty `catch()` there, but it didn't help :confused: I'm not sure if this could somehow block the notification from the other promise that has the catch added.

```
Uncaught (in promise) Error: Network error: Response not successful: Received status code 500
    ApolloError bundle.esm.js:63
    queryListenerForObserver bundle.esm.js:1247
    broadcastQueries bundle.esm.js:1559
    broadcastQueries bundle.esm.js:1557
    broadcastQueries bundle.esm.js:1555
    networkResult bundle.esm.js:1161
    promise callback*./node_modules/apollo-client/bundle.esm.js/QueryManager</QueryManager.prototype.fetchQuery/</< bundle.esm.js:1152
    step tslib.es6.js:100
    verb tslib.es6.js:81
    __awaiter tslib.es6.js:74
    __awaiter tslib.es6.js:70
    fetchQuery bundle.esm.js:1092
    observeQuery bundle.esm.js:1460
    setUpQuery bundle.esm.js:388
    onSubscribe bundle.esm.js:366
    _this bundle.esm.js:96
    Subscription Observable.js:197
    subscribe Observable.js:279
    result bundle.esm.js:126
    result bundle.esm.js:112
    query bundle.esm.js:1357
    query bundle.esm.js:1353
    query bundle.esm.js:2010
    useFetchSystems hooks.js:63
    fetchBatched hooks.js:92
    fetchBatched hooks.js:91
    _callee6$ hooks.js:278
    Babel 8
    _callee2$ useBulkSelect.js:100
    Babel 16
    selectAll useBulkSelect.js:99
    onClick rhcsVendor.daed1b7d9d7c8f24840b.js:1
    onClick pfVendor.b27a87f6e07bdbac0b69.js:1
    ze chrome-root.f7480e433de1b5ca8b2d.js:2
    Qe chrome-root.f7480e433de1b5ca8b2d.js:2
    kn chrome-root.f7480e433de1b5ca8b2d.js:2
    kn chrome-root.f7480e433de1b5ca8b2d.js:2
    En chrome-root.f7480e433de1b5ca8b2d.js:2
    Tn chrome-root.f7480e433de1b5ca8b2d.js:2
    Le chrome-root.f7480e433de1b5ca8b2d.js:2
    Tn chrome-root.f7480e433de1b5ca8b2d.js:2
    Tn chrome-root.f7480e433de1b5ca8b2d.js:2
    Yt chrome-root.f7480e433de1b5ca8b2d.js:2
    Jt chrome-root.f7480e433de1b5ca8b2d.js:2
    unstable_runWithPriority chrome-root.f7480e433de1b5ca8b2d.js:2
    Go chrome-root.f7480e433de1b5ca8b2d.js:2
    qe chrome-root.f7480e433de1b5ca8b2d.js:2
    $t chrome-root.f7480e433de1b5ca8b2d.js:2
    Cn chrome-root.f7480e433de1b5ca8b2d.js:2
    Rn chrome-root.f7480e433de1b5ca8b2d.js:2
    An chrome-root.f7480e433de1b5ca8b2d.js:2
    An chrome-root.f7480e433de1b5ca8b2d.js:2
    tc chrome-root.f7480e433de1b5ca8b2d.js:2
    _reactRootContainer chrome-root.f7480e433de1b5ca8b2d.js:2
    nc chrome-root.f7480e433de1b5ca8b2d.js:2
    render chrome-root.f7480e433de1b5ca8b2d.js:2
    h 287.ddccf5f074b7cfce941f.js:1
    63749 287.ddccf5f074b7cfce941f.js:1
    __webpack_require__ chrome-root.f7480e433de1b5ca8b2d.js:2
    promise callback*32750 chrome-root.f7480e433de1b5ca8b2d.js:2
    __webpack_require__ chrome-root.f7480e433de1b5ca8b2d.js:2
    <anonymous> chrome-root.f7480e433de1b5ca8b2d.js:2
    <anonymous> chrome-root.f7480e433de1b5ca8b2d.js:2
bundle.esm.js:63
```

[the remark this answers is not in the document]
vkrizan: It has to be wrapped in `try-catch` for an awaited function.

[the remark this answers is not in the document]
akofink: I also thought there might be something weird with the await returning the result instead of the promise, but the promise I linked to above in useFetchSystems didn't have an await :thinking: though, the trace could be pointing to the wrong root cause location.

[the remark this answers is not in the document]
vkrizan: The `fetchSystems` from `useFetchSystems` run by `fetchBatched` for `itemIdsInTable` triggered by `selectAll` from bulk select does not have the rejection or catch handled hence the uncaught error. (God, the callback dependency.)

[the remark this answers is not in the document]
akofink: We can address this as RHICOMPL-2098.

PR #1338, opened 2021-08-19
fix(SystemsTable): RHICOMPL-2101 - Pass tags to export & select requests

bastilian:
This fixes issues with selecting and exporting when filtered by tags as described in RHICOMPL-2098.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/InventoryTable.js
+++ b/src/SmartComponents/SystemsTable/InventoryTable.js
@@ -54,7 +54,11 @@ export const InventoryTable = ({
   const [items, setItems] = useState([]);
   const [total, setTotal] = useState(0);
   const tagsEnabled = useFeature('tags');
-  const { props: tagsProps } = useTags(tagsEnabled);
+  const {
+    props: tagsProps,
+    currentTags,
+    setCurrentTags,
+  } = useTags(tagsEnabled);
 
   const osMinorVersionFilter = useOsMinorVersionFilter(
     showOsMinorVersionFilter
@@ -82,6 +86,7 @@ export const InventoryTable = ({
   const systemFetchArguments = {
     query,
     variables: {
+      tags: currentTags,
       filter: systemsFilter,
       ...(policyId && { policyId }),
     },
@@ -112,6 +117,7 @@ export const InventoryTable = ({
     setTotal(result.meta.totalCount);
     setItems(result.entities);
     setIsLoaded(true);
+    setCurrentTags && setCurrentTags(result.meta.tags);
 
     if (
       emptyStateComponent &&
@@ -135,7 +141,10 @@ export const InventoryTable = ({
     filter: systemsFilter,
     selected: selectedIds,
     total,
-    fetchArguments: systemFetchArguments,
+    fetchArguments: {
+      tags: currentTags,
+      ...systemFetchArguments,
+    },
   });
 
   const mergedColumns = (defaultColumns) =>
--- a/src/SmartComponents/SystemsTable/constants.js
+++ b/src/SmartComponents/SystemsTable/constants.js
@@ -121,8 +121,15 @@ export const GET_MINIMAL_SYSTEMS = gql`
     $perPage: Int
     $page: Int
     $sortBy: [String!]
+    $tags: [String!]
   ) {
-    systems(search: $filter, limit: $perPage, offset: $page, sortBy: $sortBy) {
+    systems(
+      search: $filter
+      limit: $perPage
+      offset: $page
+      sortBy: $sortBy
+      tags: $tags
+    ) {
       totalCount
       edges {
         node {
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -89,6 +89,7 @@ export const useFetchSystems = ({
         const result = {
           entities,
           meta: {
+            ...(requestVariables.tags && { tags: requestVariables.tags }),
             totalCount: data?.systems?.totalCount || 0,
           },
         };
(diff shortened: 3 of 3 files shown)


src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -54,7 +54,11 @@ export const InventoryTable = ({
   const [items, setItems] = useState([]);
   const [total, setTotal] = useState(0);
   const tagsEnabled = useFeature('tags');
-  const { props: tagsProps } = useTags(tagsEnabled);
+  const {
+    props: tagsProps,
+    currentTags,
+    setCurrentTags,
+  } = useTags(tagsEnabled);
```
bastilian, line 61: I don't like this solution.

I'm not sure yet what an ideal solution would look like.

vkrizan: I'd suggest to look into the toolbar components that set the used tags and take that information from them. Having a separate separate (almost isolated) state populated from the response-fed result data seems wrong.

vkrizan (commented):
This fixed exporting but the bluk select is still buggy.

bastilian:
> This fixed exporting but the bluk select is still buggy.

Mind elaborating on what "buggy" means?

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: @vkrizan Yep, but this all is in the InventoryTable, so I'll probably ping someone from platform to work out a better solution.

vkrizan:
> Mind elaborating on what "buggy" means?

After applying a tag filter and Selecting all, it selected all systems, including those that did not have the expecting filter set.

bastilian:
@vkrizan Nice catch! Not sure why I didn't see that before. The tags weren't properly passed to the query, it is now.

akofink:
@bastilian this is RHICOMPL-2098? I don't see export mentioned there.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -112,6 +117,7 @@ export const InventoryTable = ({
     setTotal(result.meta.totalCount);
     setItems(result.entities);
     setIsLoaded(true);
+    setCurrentTags(result.meta.tags);
```
akofink, line 120: Facing some issues on this line:

```
Uncaught (in promise) TypeError: setCurrentTags is not a function
```

bastilian:
@akofink Ah! My bad, I mixed up jira numbers, this is actually https://issues.redhat.com/browse/RHICOMPL-2101, I'll correct the commit and title. The branch name is correct, though. :D

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: Thanks for taking a look! This is fixed now, I forgot to test with the tags turned off. Sorry about that.

akofink: Ah! haha I totally forgot about the feature flag :)

bastilian: Good thing you did. :D I've been hesitant to add tests for feature-flags since they shouldn't be long lived, but I think I'll look into having something in the future to run a few tests with and without a feature flag turned on.

src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -112,6 +117,7 @@ export const InventoryTable = ({
     setTotal(result.meta.totalCount);
     setItems(result.entities);
     setIsLoaded(true);
+    setCurrentTags && setCurrentTags(result.meta.tags);
```
vkrizan, line 120: The data flow of set tags is convoluted. 
```
filter (in toolbar) sets tags 
  -> a query is made to draw the result
    -> used tags are injected into a result of the query
      -> tags state is set for SystemsTable (SystemsTable is redrawn)
       -> Export/Select further uses the state to create queries
```

This creates couple of issues:
* possible race conditions (as export/select has to wait on the result of the draw query)
* error handling, i.e. dependency on the draw query
* second state for tags

Could tags be retrieved from the one state variable that is used to create the main draw query? For example taking it from the filter settings?

src/SmartComponents/SystemsTable/InventoryTable.js
vkrizan: Oh, sorry, just made a another comment on the same: https://github.com/RedHatInsights/compliance-frontend/pull/1338/files#r703415590

Any news on the InventoryTable enhancements?

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: > Could tags be retrieved from the one state variable that is used to create the main draw query? For example taking it from the filter settings?

Not really, we do not really know up until a request for systems is made what filters the user has selected.

akofink (approved):
Sorry for the delay getting back to this one (I figured out how to create tags :laughing:) Works as expected! Thanks, @bastilian!

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: > Any news on the InventoryTable enhancements?

There is not a concrete idea yet, how a better solution would look like or what would need to be improved in the InventoryTable.

PR #1345, opened 2021-08-20
fix(PoliciesOSTabs): RHICOMPL-1834 OS version tabs sorting

xmicha82:
Tabs in policies are now sorted naturally based on OS version

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/Utilities/helpers.js
+++ b/src/Utilities/helpers.js
@@ -1,19 +1,17 @@
 import React from 'react';
+import natsort from 'natsort';
 
 export const uniq = (collection) => [...new Set(collection)];
 
 export const sortingByProp =
   (propName, order = 'asc') =>
   (objA, objB) => {
+    const descending = order != 'asc';
     const propA = (objA && objA[propName]) || '';
     const propB = (objB && objB[propName]) || '';
-    if (propA === propB) {
-      return 0;
-    } else if (order === 'asc') {
-      return propA < propB ? -1 : 1;
-    } else {
-      return propA < propB ? 1 : -1;
-    }
+
+    const sorter = natsort({ desc: descending });
+    return sorter(propA, propB);
   };
 
 // eslint-disable-next-line react/display-name
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,6 +76,7 @@
         "jest": "^26.6.3",
         "jest-canvas-mock": "^2.3.0",
         "mini-css-extract-plugin": "^1.4.1",
+        "natsort": "^2.0.2",
         "npm-run-all": "^4.1.3",
         "react-test-renderer": "^17.0.2",
         "redux-mock-store": "^1.5.4",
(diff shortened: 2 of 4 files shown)


src/Utilities/helpers.js
```diff
@@ -5,8 +5,9 @@ export const uniq = (collection) => [...new Set(collection)];
 export const sortingByProp =
   (propName, order = 'asc') =>
   (objA, objB) => {
-    const propA = (objA && objA[propName]) || '';
-    const propB = (objB && objB[propName]) || '';
+    const tryParseInt = (obj) => parseInt(obj) || obj;
+    const propA = tryParseInt((objA && objA[propName]) || '');
```
bastilian, line 9: This might cause issues with properties that aren't containing any numbers to sort, would it not? 

Maybe for now an option to enable parsing for an integer would be better.
There might also be floats for versions like "7.9" etc.

bastilian: Sorry, my bad. misread the code a little.

vkrizan: It would cause some sort issues in case the string starts with a (parsable) number. For example `123xyz` and `123abcd` would be both resulting in number `123` leading to an incorrect sort.

The simplest way would be to only `parseInt` if it is a number-like string. A better alternative would be to create a natural sorting.

bastilian: +1 to natural sorting. Though maybe before we implement our custom sorter further, maybe we should consider using an already existing solution. I found this: https://github.com/bubkoo/natsort, which looks promising.

We could make use of that also in other places where we do sorting, for example in the tables, where we also sometimes see issues with odd sorting.

src/Utilities/helpers.js
```diff
@@ -1,19 +1,17 @@
...
     const propB = (objB && objB[propName]) || '';
-    if (propA === propB) {
-      return 0;
-    } else if (order === 'asc') {
-      return propA < propB ? -1 : 1;
-    } else {
-      return propA < propB ? 1 : -1;
-    }
+
+    const sorter = natsort({ desc: descending });
```
bastilian, line 13: Just a note, not a required change. The cons above could match the property key here then it could just be 
```js
const desc = order != 'asc';
...
const sorter = natsort({ desc });
```

Since we don't use the `descending` const anywhere else in the function and the statement is not as complex as propA and propB, the statement could also be directly in the argument object, without declaring the `descending` const. But that might just be my preference.

bastilian (approved):
Thank you @xmicha82! Nice work and refactor!

Merge at will. :)

PR #1347, opened 2021-08-23
feat(SystemsTable): RHICOMPL-921 - Add "Last seen" inventory column

bastilian:
This makes use of the recently added timestamps to add the "Last seen" column and removes the details link on the "Systems" page.

Screenshot 2021-09-02 at 11 18 26

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -49,7 +49,7 @@ export const ComplianceSystems = () => {
                   }),
                   Columns.inventoryColumn('tags'),
                   Columns.Policies,
-                  Columns.inventoryColumn('updated'),
+                  Columns.inventoryColumn('updated', { isStatic: true }),
                 ]}
                 query={GET_SYSTEMS}
                 defaultFilter={DEFAULT_FILTER}
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -49,7 +49,7 @@ export const ComplianceSystems = () => {
                   }),
                   Columns.inventoryColumn('tags'),
                   Columns.Policies,
-                  Columns.DetailsLink,
+                  Columns.inventoryColumn('updated'),
                 ]}
                 query={GET_SYSTEMS}
                 defaultFilter={DEFAULT_FILTER}
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -62,6 +62,22 @@ export const useSystemsFilter = (
   return filter;
 };
 
+const renameInventoryAttributes = ({
+  culledTimestamp,
+  staleWarningTimestamp,
+  staleTimestamp,
+  insightsId,
+  lastScanned,
+  ...system
+}) => ({
+  ...system,
+  updated: lastScanned,
+  culled_timestamp: culledTimestamp,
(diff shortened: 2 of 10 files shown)


skateman (approved):
The Seal of Approval
Code & screenshot look god to me :+1:

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -62,6 +62,22 @@ export const useSystemsFilter = (
...
+const renameInventoryAttributes = ({
+  culledTimestamp,
+  staleWarningTimestamp,
+  staleTimestamp,
+  insightsId,
+  lastScanned,
+  ...system
+}) => ({
+  ...system,
+  updated: lastScanned,
```
bastilian, line 74: The Cindy hosts should also have a "updated" field, and we should probably use that instead of the "lastScanned" as it otherwise might differ from what other applications pass as "updated".

akofink: This seems pretty confusing to me. Could we keep `lastScanned` and use that value within the column definition? Then if we ever need the insights last check in (aka last advisor upload), we also have that info available as a separate param.

bastilian: @akofink We can keep the lastScanned. Like it is implemented here at the moment you mean, or?

akofink: There are two 'last updated' type fields. One for inventory (i.e. Advisor uploads/normal insights checkins) and one for Compliance. We may need both in the future on the same page. I'm just requesting that we ensure the naming won't be confusing if that ever happens. I thought `lastScanned` was for Compliance and `updated` was for Advisor. Is that correct?

bastilian: I'm confused. Are you talking about the actual fields in the systems GraphQL type or the property naming here? If the naming here. We cannot change the `updated` to a different name as that is the property that the Inventory table column component for "Last seen" is looking for, which we set to `lastScanned` here, which we get from our data, because we do not expose the update field in out GraphQL API.

My suggestion to use the `updated` field from the hosts-table and not assign our `lastScanned`, we construct from the last report, is made under the assumption that the updated field is changed with any checkin, including compliance. This would ensure that the "Last seen" in Inventory, Advisory, Patch or any other app show exactly the same time.

akofink: Sorry for the confusion! I went back and looked at the mocks for this jira, and I agree with you that the 'Last seen' should probably be the `updated` field from inventory (aka Advisor), not Compliance's `lastScanned` field.

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -49,7 +49,7 @@ export const ComplianceSystems = () => {
                   }),
                   Columns.inventoryColumn('tags'),
                   Columns.Policies,
-                  Columns.DetailsLink,
+                  Columns.inventoryColumn('updated'),
```
vkrizan, line 52: The "Last seen" column is sortable, however the sorting does not work.

Observation: `DetailsLink` column is no longer used.

bastilian: @vkrizan Thanks for catching that! This should not be fixed and we can have custom props now for the default columns.

PR #1351, opened 2021-08-25
feat(Reports): RHICOMPL-2168 - Implement export settings modal

bastilian:
This adds the "Download Report" feature (part 1). After enabling the feature via `?pdfReport=enable` in the URL a button on the Report page:

Screenshot 2021-09-02 at 12 35 49

Which will open a modal with settings for the PDF to export:

Screenshot 2021-09-02 at 12 36 52

Till part 2 is implemented pushing the export button will only query the neccessary data and put it into a workable format for the pdf document added via part 2.

RHICOMPL-2169 will implement part 2 and actually incorporate pdf-generator and implement the document pages.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


src/__factories__/profiles.js
```diff
@@ -0,0 +1,45 @@
...
+  compliant: true,
+  ssgVersion: '0.14.5',
+  supported: true,
+  ...attributes,
+});
+
+const buildProfiles = (count = 5, attributes = {}) =>
+  callAndSort(buildProfile, count, { funcArguments: [attributes] });
+
+export const buildNonCompliantProfiles = (count = 5, attributes = {}) =>
```
bastilian, line 31: This is to try and introduce something that allows us to build mock data for tests, and maybe even for actual component development.

bastilian: For now these are only to satisfy the tests for this PR. And there is probably a better way to construct them. But we can let them evolve a little bit and see what makes sense.

akofink (commented):
The modal looks as expected from the mocks. :+1:

I'm seeing the following issue:
1. Click 'Download PDF' (the modal opens)
2. Click the 'Cancel' button (the modal is dismissed)
3. Click 'Download PDF' again

Expected: The modal opens again

Actual: Error in the console

```
Uncaught DOMException: The object could not be cloned.
```

bastilian:
@akofink Thank you for taking a look and catching! I'll take a look at the issue!

src/SmartComponents/ReportDownload/ReportDownload.js
```diff
@@ -0,0 +1,80 @@
...
+  const exportPDF = policy
+    ? usePDFExport(exportSettings, policy)
+    : () => Promise.resolve([]);
+
+  const actions = [
+    <DownloadButton
+      groupName="Red Hat Insights"
+      key="export"
+      label="Export report"
+      fileName={`Report.pdf`}
```
akofink, line 39: Was this filename specified in the requirements? Maybe it'd make sense to add a timestamp.

bastilian: I don't remember any specified in the requirements (and Jira doesn't load for me atm.), but I did update it to be similarly named as vulnerabilities does: https://github.com/RedHatInsights/vulnerability-ui/blob/e70350d42ef492fcd9db9e30a694c8318f70aa2c/src/Helpers/DownloadReport.js#L41

I'll check the requirements again once I can and update the filename in the PDF implementation PR.

bastilian:
@akofink The issue with canceling should be fixed, but it is not a fix I'd like. I'm not sure why this is causing issues and why `pendo.js` is raising this error. I'll further investigate, but for now it just uses the default browser history back.

src/SmartComponents/ReportDownload/ReportDownload.js
bastilian: Checked the requirements doc and it does not mention any specific filename format.

akofink:
I'm having some trouble understanding the changes within this PR. There are over 6k lines added! The commit that adds the modal makes sense, but I don't understand why all the other changes are necessary for this new modal.

bastilian:
@akofink Aside from the changes that add the button and the modal it adds fetching the data based on the settings given in the modal. For now these will only show up in the console log (prefixed with "DATA"), later this will be passed to the PDF document component which will build the source for the PDF to render. A bulk of the the 6k are 3k of snapshots of that data compiled.

src/SmartComponents/ReportDownload/hooks/__snapshots__/helpers.test.js.snap
```diff
@@ -0,0 +1,46 @@
...
+  "Profile #5 Rule #2",
+  "Profile #5 Rule #3",
+  "Profile #5 Rule #4",
+  "Profile #5 Rule #5",
+  "Profile #5 Rule #6",
+  "Profile #5 Rule #7",
+  "Profile #5 Rule #8",
+  "Profile #5 Rule #9",
+]
+`;
```
akofink, line 36: :thinking: I tried this with an unsupported host, btw. So perhaps that's why my list was empty - I think that's what we want for unsupported hosts. I will retry it with a supported host.

src/SmartComponents/ReportDownload/hooks/__snapshots__/usePDFBuilder.test.js.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`usePDFBuilder buildPages returns a query function 1`] = `
+Array [
+  <div />,
+]
+`;
```
akofink, line 7: Sounds like you're expecting a function, but this is an array.

src/SmartComponents/ReportDownload/hooks/__snapshots__/useQueryExportData.test.js.snap
```diff
@@ -0,0 +1,21 @@
...
+`;
+
+exports[`useQueryExportData returns a query function 1`] = `[Function]`;
+
+exports[`useQueryExportData returns a query function 2`] = `
+Object {
+  "nonCompliantSystems": 20,
+  "topTenFailedRules": 10,
+  "unsupportedSystems": 10,
+}
```
akofink, line 20: Also not a function

akofink (commented):
I'm not completely done reviewing, but I have some comments already (and some inline):

- [ ] The data sent doesn't appear to contain the top failed rules when that box is checked
- [ ] Checking 'Non-compliant systems' while leaving 'Systems with unsupported configuration' unchecked results in my unsupported systems still being sent within `nonCompliantSystems`
  - Remember unsupported systems should not be included in compliant or non-compliant system counts/lists
- [ ] There is no form validation. What is expected when all boxes are unchecked and all text fields empty?
- [X] API error handling seems to work as expected :+1: 

None of these issues are really problematic until we have the PDF generator working and this feature is turned on; however, just based on the data passed, it seems like some things aren't working quite right that we'd need to go back and fix once the remaining pieces of this work are in place.

bastilian:
@akofink I've updated the tests and their titles to appear better in the snapshots and be more readable.

I couldn't reproduce the issue with the failed rules. Do you have any further clues?

> There is no form validation. What is expected when all boxes are unchecked and all text fields empty?

We can have the button disabled in that case, but in https://github.com/RedHatInsights/compliance-frontend/pull/1376.

> Remember unsupported systems should not be included in compliant or non-compliant system counts/lists

I've just looked at the requirements document and it doesn't say wether or not they should be excluded. We do not exclude them in the actual systems table on the report, we just mark them as unsupported.

akofink:
> I've just looked at the requirements document and it doesn't say wether or not they should be excluded. We do not exclude them in the actual systems table on the report, we just mark them as unsupported.

The counts at the top of the report details page shows the number compliant, the number non-compliant, and the number unsupported. The sum of those three is the total sum of hosts.

bastilian:
@akofink Sorry, I totally forgot they were excluded from that. They are now also excluded here in the data from everything else, but the list of unsupported systems.

akofink (approved):
Works quite well! We met to discuss the couple remaining minor issues. Looking forward to merging! Thanks, @bastilian!

PR #1352, opened 2021-08-25
fix: Use FEC Truncate instead of react-truncate

bastilian:
This replaces the outdated react-truncate with the `Truncate` component provided by frontend-components:

Screenshot 2021-08-25 at 11 47 34
Screenshot 2021-08-25 at 11 47 14

⚠️ This requires the package bump and test skipping from https://github.com/RedHatInsights/compliance-frontend/pull/1369

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


bastilian:
@katierik This is replacing the way we truncate text, with a component provided by the platform. However this changes some minor things. Mainly instead of just cutting off the text, it will actually add a "Read more" link. Is this OK, or should we extend the platform component to support the way we are currently cutting text?

katierik:
@bastilian Adding a "read more" is great. 👍

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
```diff
@@ -18,12 +17,15 @@ import {
 import PropTypes from 'prop-types';
 import { UnsupportedSSGVersion } from '../UnsupportedSSGVersion/UnsupportedSSGVersion';
 
+const truncate = (text, length = 150) =>
+  text.length > length ? text.substring(0, length) + '...' : text;
+
```
akofink, line 22: :thinking: so there are cases where we don't want a 'Read more' and instead want this mouseover functionality - I feel like we should extend the Truncate component to allow for that case?

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -98,8 +99,17 @@ DetailsLink.propTypes = {
...
 
-export const Policies = ({ policies }) =>
-  (policies || []).map((p) => p.name).join(', ');
+export const Policies = ({ policies }) => (
+  <Truncate
+    inline
+    text={(policies || []).map((p) => p.name).join(', ')}
+    lenght={100}
+  />
+);
```
akofink, line 108: Nit: we don't need Truncate at all if policies are null/empty.

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -98,8 +99,17 @@ DetailsLink.propTypes = {
   testResultProfiles: propTypes.array,
 };
 
-export const Policies = ({ policies }) =>
-  (policies || []).map((p) => p.name).join(', ');
+export const Policies = ({ policies }) => (
+  <Truncate
+    inline
+    text={(policies || []).map((p) => p.name).join(', ')}
+    lenght={100}
```
akofink, line 106:
```suggestion
    length={100}
```

akofink (commented):
Thanks, @bastilian! Just a few questions inline

akofink:
I'd also nitpick about this being a 'chore' since it's changing behavior from mouseover to 'read more' in at least one place.

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
bastilian: Yes, I think, we can open a PR and add that functionality there. I haven't done that yet, because it also requires the mouse-over and out behaviour to be implemented, for this use-case here.

bastilian:
@akofink I just picked "chore" because that is what I mostly use for package updates, I can change it. :) However, I did not intend to change that behaviour. I must have overlooked that. I'll check again and fix that.

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
akofink: :+1: we can definitely do it after this PR, but it caught my eye that we're using this new Truncate component but also having to define a custom `truncate` function! :laughing:

bastilian:
@akofink I checked that changes, I couldn't find the place you mention where we chance from mouse over to "read me". The only place where anything happens on mouseover, which is already the case in master. The places where we do add a "Read me" are all "on click" and are confirmed above with UX.

akofink (approved):
Works as expected. Thanks!

akofink (changes requested):
Actually, this seems to break the system details page for unsupported systems. The test failure pointed me towards the scenario, and it definitely also fails in the UI. Unfortunately, I'm unsure what the issue is - it's still wrapped inside a minified react error :confused: How do you see the actual trace?

bastilian:
@akofink It was probably missing the latest changes from master. It was still showing the "Update branch" button. I just rebased the commits again on top of the latest master again.

akofink (approved):
Thanks, @bastilian! Not sure why my rebase wasn't correct :thinking: :laughing: Works as expected!

PR #1354, opened 2021-08-26
feat(Policies): RHICOMPL-2116 - Limit systems by minor version too

bastilian:
This updates the queries for the systems in the Create Policy wizard and the Edit Policy modal to also include the specific minor versions of the selected/edited policy in addition just the major OS version.

**How to test:**

1. Open the Create policy wizard 
2. Select a policy and advance to the "Systems" step
3. Verify on the Systems step that only systems specific to the supported OS minor versions are shown.
4. Finish the Wizard
5. Open the "Edit Policy" modal for the policy created
6. Switch to the "Systems" tab
7. Verify that only systems for the specific OS version of that policy are shown.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/EditPolicySystems.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicySystems.js
@@ -69,6 +69,7 @@ PrependComponent.propTypes = {
 };
 
 export const EditPolicySystems = ({
+  policy,
   change,
   osMajorVersion,
   selectedSystems,
@@ -77,7 +78,9 @@ export const EditPolicySystems = ({
     change('systems', newSelectedSystems);
     change('osMinorVersionCounts', countOsMinorVersions(newSelectedSystems));
   };
-
+  const osMinorVersions = policy.supportedOsVersions.map(
+    (version) => version.split('.')[1]
+  );
   return (
     <React.Fragment>
       <TextContent className="pf-u-mb-md">
@@ -107,7 +110,10 @@ export const EditPolicySystems = ({
             showActions={false}
             query={GET_SYSTEMS_WITHOUT_FAILED_RULES}
             defaultFilter={
-              osMajorVersion && `os_major_version = ${osMajorVersion}`
+              osMajorVersion &&
+              `os_major_version = ${osMajorVersion} AND os_minor_version ^ (${osMinorVersions.join(
+                ','
+              )})`
             }
             enableExport={false}
             preselectedSystems={selectedSystems}
@@ -121,6 +127,7 @@ export const EditPolicySystems = ({
 
 EditPolicySystems.propTypes = {
   osMajorVersion: propTypes.string,
+  policy: propTypes.object,
   selectedSystems: propTypes.array,
   change: reduxFormPropTypes.change,
 };
@@ -131,6 +138,7 @@ EditPolicySystems.defaultProps = {
 
 const selector = formValueSelector('policyForm');
 const mapStateToProps = (state) => ({
+  policy: selector(state, 'profile'),
   osMajorVersion: selector(state, 'osMajorVersion'),
   selectedSystems: selector(state, 'systems'),
 });
--- a/src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
+++ b/src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
@@ -45,12 +45,24 @@ PrependComponent.propTypes = {
 };
 
 const EditPolicySystemsTab = ({
-  policy: { id: policyId, osMajorVersion },
+  policy,
   newRuleTabs,
   onSystemSelect,
   selectedSystems,
 }) => {
   const { push, location } = useHistory();
+  const { id: policyId, osMajorVersion, supportedOsVersions } = policy;
+  const osMinorVersions = supportedOsVersions.map(
+    (version) => version.split('.')[1]
+  );
+  const osFilter =
+    osMajorVersion &&
+    `os_major_version = ${osMajorVersion} AND os_minor_version ^ (${osMinorVersions.join(
+      ','
+    )})`;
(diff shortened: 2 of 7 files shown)


src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
```diff
@@ -68,8 +72,11 @@ const EditPolicySystemsTab = ({
         query={GET_SYSTEMS_WITHOUT_FAILED_RULES}
         defaultFilter={
           osMajorVersion &&
-          `os_major_version = ${osMajorVersion} or policy_id = ${policyId}`
+          `os_major_version = ${osMajorVersion} AND os_minor_version ^ (${osMinorVersions.join(
```
vkrizan, line 75: Where did the `policy_id = ${policyId}` condition went? Are we sure that we want to hide systems that are already part of the policy? This does not seem logical to me. @skateman wdyt?

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -77,7 +78,9 @@ export const EditPolicySystems = ({
     change('systems', newSelectedSystems);
     change('osMinorVersionCounts', countOsMinorVersions(newSelectedSystems));
   };
-
+  const osMinorVersions = policy.supportedOsVersions.map(
+    (version) => version.split('.')[1]
```
vkrizan, line 82: @skateman this is not great :grimacing:. Maybe we should reassess what we return on `supportedOsVersions` to avoid custom made minor version parser. Preferably returning already parsed values. Note that API was added via https://github.com/RedHatInsights/compliance-backend/pull/1114/files#diff-0414a45947b697e51d740c5ba32eb6c6445f4c795807250ce68106de5ff7e414R78

src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
vkrizan: If we are going to hide those. Can we add some warning message to the modal? Something like "5 systems are not supported by this policy and will be removed on policy save.". Of course this would be done outside of this PR, if we'd go with it.

bastilian: Look a few lines lower. It now uses the `policyId` prop that will set the filter instead.

vkrizan: `policyId` was used in OR relationship. @bastilian Can you please elaborate on how the `policyId` is being used for querying? It seems to be used as a variable for GraphQL queries, but some do not utilize it. Sorry, it is hard to find out just by reading the code.

Here are the fetch args defined https://github.com/RedHatInsights/compliance-frontend/blob/e6c2f3e090cf2bd0883c324d002ccfabe9a1d88a/src/SmartComponents/SystemsTable/SystemsTable.js#L97-L104
here are they being used for [bulk select](https://github.com/RedHatInsights/compliance-frontend/blob/e6c2f3e090cf2bd0883c324d002ccfabe9a1d88a/src/SmartComponents/SystemsTable/SystemsTable.js#L111-L123) which uses [`GET_MINIMAL_SYSTEMS`](https://github.com/RedHatInsights/compliance-frontend/blob/e6c2f3e090cf2bd0883c324d002ccfabe9a1d88a/src/SmartComponents/SystemsTable/constants.js#L128-L159) that does not utilize it.
And the main query that is used for the edit modal is [`GET_SYSTEMS_WITHOUT_FAILED_RULES`](https://github.com/RedHatInsights/compliance-frontend/blob/e6c2f3e090cf2bd0883c324d002ccfabe9a1d88a/src/SmartComponents/SystemsTable/constants.js#L72-L126).

The `policyId` argument scopes polices under each systems. Hence the issue that I've described is still present.

bastilian: My bad. Thought the policyId was utilised differently. Changed it back to include it in the filter.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
bastilian: @vkrizan It's not great, but it works for now. If we improve the response, we can adjust this later.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -77,7 +78,9 @@ export const EditPolicySystems = ({
...
+  );
```
johnsonm325, line 83: Looks like this is repeated in EditPolicySystemsTab. This could be pulled out into a helper.

src/SmartComponents/CreatePolicy/EditPolicySystems.js
```diff
@@ -107,7 +110,10 @@ export const EditPolicySystems = ({
             showActions={false}
             query={GET_SYSTEMS_WITHOUT_FAILED_RULES}
             defaultFilter={
-              osMajorVersion && `os_major_version = ${osMajorVersion}`
+              osMajorVersion &&
+              `os_major_version = ${osMajorVersion} AND os_minor_version ^ (${osMinorVersions.join(
+                ','
+              )})`
```
johnsonm325, line 116: Looks like this is repeated in EditPolicySystemsTab. This could be pulled out into a helper.

johnsonm325 (approved):
I'm approving this with a suggestion to pull two of these code blocks into a helper function. If you don't want to do it, that's fine.

PR #1375, opened 2021-09-09
chore(lockfile): Add package name into the lockfile

skateman:
`npm install` kept adding this line for me :thinking:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "compliance-frontend",
       "version": "1.1.0",
       "dependencies": {
         "@apollo/client": "^3.3.15",


bastilian:
@skateman that is odd... that shouldn't happen. What version of npm are you using?

bastilian:
I'm using `7.16.0` and it doesn't seem to make that change then running `npm install`.

skateman:
7.21.0

package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "compliance-frontend",
```
akofink, line 8: Isn't this the one that's on line 2?

akofink (approved):
I checked out the frontend-components package-lock.json, and it contains this as well. It doesn't seem to hurt anything to merge :+1:

PR #1378, opened 2021-09-17
Fix federated modules entries.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16201

The inventory compliance was not loaded because the module was not exposed in prod webpack config.
Also, the RootApp in dev env was pointing to an incorrect js file. This is probably why there were two app instances in dev env.


bastilian (approved):
Thank you @Hyperkid123!

PR #1388, opened 2021-09-23
fix(SystemsTable): RHICOMPL-2206 - Prevent EmptyState issues when tag filtering

bastilian:
This prevents issues when tags are enabled and a user filters with them, the "Create Policy" wizard and "Edit Policy" modal would show the empty state instead of the systems table, and block any further selection of systems.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/InventoryTable.js
+++ b/src/SmartComponents/SystemsTable/InventoryTable.js
@@ -122,7 +122,8 @@ export const InventoryTable = ({
     if (
       emptyStateComponent &&
       result.meta.totalCount === 0 &&
-      activeFilterValues.length === 0
+      activeFilterValues.length === 0 &&
+      result?.meta?.tags?.length === 0
     ) {
       setIsEmpty(true);
     }
--- a/src/SmartComponents/SystemsTable/InventoryTable.test.js
+++ b/src/SmartComponents/SystemsTable/InventoryTable.test.js
@@ -1,15 +1,26 @@
+import { render } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import { init } from 'Store';
-import { Table } from '@patternfly/react-table';
-import PrimaryToolbar from '@redhat-cloud-services/frontend-components/PrimaryToolbar';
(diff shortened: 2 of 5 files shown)


src/SmartComponents/SystemsTable/InventoryTable.js
```diff
@@ -122,7 +122,8 @@ export const InventoryTable = ({
     if (
       emptyStateComponent &&
       result.meta.totalCount === 0 &&
-      activeFilterValues.length === 0
+      activeFilterValues.length === 0 &&
+      result?.meta?.tags?.length === 0
```
akofink, line 126: This seems a bit like a hack to me, but we could fix it later if it becomes a problem. When could it be a problem? If we ever need to use tags within the default filter. Like, if we got a requirement to exclude any SAP systems from compliance, we'd need to filter on tags. Anyway, not a problem today :)

akofink (approved):
Nice! Thanks, @bastilian!

src/SmartComponents/SystemsTable/InventoryTable.js
bastilian: @akofink Yeah, it is a bit of a hack. However, it is only meant as a preventative measure, in case we do not get the functionality to only show tags for the current set of systems. So, in the best case this would not be triggered anymore. 

But on a side note, most of the things we do here in the onComplete are all kind of hacky. For example, we only set the tags in a state in order to pass it to the selection and the export. Ideally, both the export and selection would have some sort of table/filter state passed by the FECInventoryTable and be able to get information about the current filters and tags. A similar thing is with the items, which we only do because we want the ids of the current page in the table. 

I am collection nuances like this with the InventoryTable and see if we can ask or propose a good solution to get rid of these.

PR #1399, opened 2021-10-04
Remove local sass-loader dependnecy.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16436


karelhala (approved):

src/PresentationalComponents/ComplianceRemediationButton/__snapshots__/ComplianceRemediationButton.test.js.snap
```diff
@@ -2,29 +2,17 @@
...
-      />
-    </svg>
-     Remediate
-  </button>
+  <div>
+    <h1>
+      Unable to load remedaitions component
+    </h1>
+    <h2>
+      Failed to load
```
bastilian, line 10: ... not a fan of fed-mods in this context. :D

We could mock `RemediationButton`, though, I'm curious why this changes here now?

Hyperkid123: It is simple, the button is no longer a component in FEC but an async component. That is why it fails to load. If you want me to add a "Nice component" mock I can do that.

bastilian: @Hyperkid123 I think we have started a mock somewhere already. I don't mind this for now. But it would still be nice to be able an include the real actual modules/components that would usually be provided via fed mods as local components here. 

We'd like to start more integration tests and not being able to load fed mods in the rendering in tests would require us to start a mock emulating the remediations button and the inventory table.

Hyperkid123: Integration testing tools should be able to load remote chunks if the module is registered properly. Once you start with it, let me know and I can help with the configuration. In JSDOM (jest env) you will never be able to render remote modules I am affraid.

bastilian: Anyways, I'll give this a quick spin later to make sure everything still works as expected and approve it and we can look into the fed mod mocks later.

Hyperkid123:
> @Hyperkid123 I've tried to run the proxy with this, but it throws this:
> 
> ```
> npm run start:proxy                                                                                                                                                                                                ✔  took 54s  ▓▒░
> 
> > compliance-frontend@1.1.0 start:proxy
> > PROXY=true npm run start
> 
> 
> > compliance-frontend@1.1.0 start
> > NODE_ENV=development webpack serve --config config/dev.webpack.config.js
> 
> Configuration provided {
>   rootFolder: '/Users/bastilian/Projects/RedHat/CSfR/compliance-frontend',
>   debug: false,
>   useFileHash: false,
>   deployment: 'apps',
>   appUrl: [ '/insights/compliance' ],
>   env: 'stage-stable',
>   useProxy: true,
>   proxyVerbose: true,
>   useCloud: false,
>   routesPath: '/Users/bastilian/Projects/RedHat/CSfR/compliance-frontend/config/spandx.config.js',
>   routes: {}
> }
> Making proxy from SPANDX routes
> [webpack-cli] Invalid options object. Dev Server has been initialized using an options object that does not match the API schema.
>  - options has an unknown property 'inline'. These properties are valid:
>    object { allowedHosts?, bonjour?, client?, compress?, devMiddleware?, headers?, historyApiFallback?, host?, hot?, http2?, https?, ipc?, liveReload?, magicHtml?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, port?, proxy?, setupExitSignals?, static?, watchFiles?, webSocketServer? }
> ```
> 
> I did do a `rm -rf node_modules` before `npm` installing.

@bastilian  Right, you are passing custom properties to `webpack-dev-server`, but there are breaking changes between v3 and v4. I will update your custom config. I disabled it for my local run.

bastilian:
@Hyperkid123 Thank you!

Hyperkid123:
no :facepalm:

PR #1415, opened 2021-10-19
feat(SystemsTable): RHICOMPL-2379 - Use getTags to retrieve tags

bastilian:
This implements a custom getTags function and passes it to the InventoryTable component to fetch tags from the compliance GraphQL API instead of the inventory API.

https://user-images.githubusercontent.com/7757/138154398-c3b47149-ce12-4f70-b868-5f05a7fe645c.mp4

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -375,6 +375,15 @@ export const useSystemBulkSelect = ({
   };
 };
 
+const searchTagsByKey = (search, tags) =>
+  tags.filter((tagItem) => {
+    if (search || search === '') {
+      return tagItem?.key.indexOf(search) !== -1;
+    } else {
+      return true;
+    }
+  });
+
 const useFetchTag = () => {
   const apiClient = useApolloClient();
 
@@ -389,15 +398,10 @@ const useFetchTag = () => {
           data: {
             systems: { tags },
           },
-        }) => {
-          return tags.filter((tagItem) => {
-            if (search || search === '') {
-              return tagItem.tag.key.indexOf(search) !== -1;
-            } else {
-              return true;
-            }
-          });
-        }
+        }) =>
+          searchTagsByKey(search, tags).map((tag) => ({
+            tag,
+          }))
       );
 
     const start = per_page * page - per_page;


src/SmartComponents/SystemsTable/constants.js
```diff
@@ -158,6 +158,18 @@ export const GET_MINIMAL_SYSTEMS = gql`
   }
 `;
 
+export const GET_SYSTEMS_TAGS = gql`
+  query getSystems($filter: String!, $limit: Int) {
+    systems(search: $filter, limit: $limit) {
+      tags {
```
vkrizan, line 164: I don't think that the pagination works for tags as it is expected within these changes. `tags` represents metadata, containing all available tags that user has. The `search` and `limit` is effective for System objects/records, which are not retrieved in this query.

@akofink is my assumption correct? If so, then we likely need a separate `getTags` query.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -371,8 +375,64 @@ export const useSystemBulkSelect = ({
...
-export const useTags = (tagsEnabled) => {
-  const [currentTags, setCurrentTags] = useState([]);
+const useFetchTag = () => {
+  const apiClient = useApolloClient();
+
+  return async (page, per_page, search, fetchArguments) => {
+    const fetchedTags = await apiClient
+      .query({
+        query: GET_SYSTEMS_TAGS,
+        ...fetchArguments,
```
vkrizan, line 385: Can this do query only when the fetch arguments change? Since the pagination is done by slicing, we can slice data that were already downloaded.

bastilian: Yeah, we could do that, not sure if the fetchArguments is a good trigger for busting that cache though. So for now, I'd refrain from trying to cache that.

akofink (approved):

PR #1418, opened 2021-10-20
fix(Modal): RHICOMPL-2387 - Add modal with compliance class for scoped CSS

bastilian:
Application specific CSS is now scoped on the platform, which means all our components will need to be within an element with the application ID as a class to have styles properly applied. The platform chrome does this already for the main part of our application, but not for modals. This adds a modal with the class predefined to use within compliance and have CSS styles available and similarly adds the compliance class to the `Wizard` (modal)

Before:

Screenshot 2021-10-20 at 17 48 28

After:

Screenshot 2021-10-20 at 17 47 51

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/PresentationalComponents/ComplianceModal/ComplianceModal.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Modal } from '@patternfly/react-core';
+
+// Provides a modal with defaults used in compliance
+const ComplianceModal = ({ children, ...props }) => {
+  return (
+    <Modal className="compliance" {...props}>
```
skateman, line 8: Isn't a way to pass the class to all the modals? Or is this a DRY solution?

bastilian: This is the DRY way, yeah. We could add it separately to each `Modal`, but this is "safer" to me and less risk of forgetting to set the class when adding a new Modal or accidentally removing it.

bastilian: > Is it possible to change defaultProps of a Modal?

? Not really sure how that should work without rebuilding the component...

> ... can it combine className form props and the default one?

Yeah, we should do that. I'll add that.

skateman: What about having a decorator function around the components that adds the className to the params, but doesn't actually create a new component? Simple tradeoff :wink:

src/PresentationalComponents/ComplianceModal/ComplianceModal.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { Modal } from '@patternfly/react-core';
+
+// Provides a modal with defaults used in compliance
+const ComplianceModal = ({ children, className = '', ...props }) => {
+  return (
+    <Modal
+      className={`compliance${className !== '' ? ` ${className}` : ''}`}
```
vkrizan, line 9: Nit: Empty string is considered false. Can be shortened to this and cover a `null` case:
```
`compliance${className ? ` ${className}` : ''}`
```

PR #1427, opened 2021-10-27
fix(EditPolicy): RHICOMPL-741 restrict entering of long floats

xmicha82:
I had to modify how useValidateThreshold hook works, because i could not think of any other solution to how to differentiate between the two error states that can happen

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/PresentationalComponents/ProfileThresholdField/ProfileThresholdField.js
+++ b/src/PresentationalComponents/ProfileThresholdField/ProfileThresholdField.js
@@ -3,21 +3,24 @@ import { Field, reduxForm } from 'redux-form';
 import { FormGroup } from '@patternfly/react-core';
 import { ReduxFormTextInput } from 'PresentationalComponents/ReduxFormWrappers/ReduxFormWrappers';
 import propTypes from 'prop-types';
-import round from 'lodash/round';
 import { thresholdValid } from '../../SmartComponents/CreatePolicy/validate';
-import { PolicyThresholdTooltip } from 'PresentationalComponents';
+import {
+  PolicyThresholdTooltip,
+  ComplianceThresholdHelperText,
+} from 'PresentationalComponents';
 
 export class ProfileThresholdField extends React.Component {
   state = {
     validThreshold: thresholdValid(this.props.previousThreshold),
-    threshold: round(this.props.previousThreshold || 100, 1),
+    threshold: this.props.previousThreshold,
   };
 
-  handleThresholdChange = (threshold) =>
+  handleThresholdChange = (threshold) => {
     this.setState({
       validThreshold: thresholdValid(threshold),
-      threshold: round(threshold, 1),
+      threshold: threshold,
     });
+  };
 
   render() {
     const { threshold, validThreshold } = this.state;
@@ -27,7 +30,9 @@ export class ProfileThresholdField extends React.Component {
         <FormGroup
           fieldId="policy-threshold"
           validated={validThreshold ? 'default' : 'error'}
-          helperTextInvalid="Threshold has to be a number between 0 and 100"
+          helperTextInvalid={
+            <ComplianceThresholdHelperText threshold={threshold} />
+          }
           helperText="A value of 95% or higher is recommended"
           labelIcon={<PolicyThresholdTooltip />}
           label="Compliance threshold (%)"
--- a/src/SmartComponents/CreatePolicy/CreatePolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreatePolicy.js
@@ -147,7 +147,7 @@ export default connect((state) => ({
   osMajorVersion: selector(state, 'osMajorVersion'),
   osMinorVersionCounts: selector(state, 'osMinorVersionCounts'),
   businessObjective: selector(state, 'businessObjective'),
-  complianceThreshold: selector(state, 'complianceThreshold') || '100.0',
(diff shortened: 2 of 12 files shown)


src/PresentationalComponents/ProfileThresholdField/ProfileThresholdField.js
```diff
@@ -9,13 +9,13 @@ import { PolicyThresholdTooltip } from 'PresentationalComponents';
 
 export class ProfileThresholdField extends React.Component {
   state = {
-    validThreshold: thresholdValid(this.props.previousThreshold),
+    validThreshold: thresholdValid(this.props.previousThreshold) === 1,
```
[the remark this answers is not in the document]
xmicha82: since i have to somehow differentiate between the two *incorrect* states (number is not between 0 and 100/number has too many decimal places) and since the current value of the field is not available in the component context, the only way (that i can think of) is to make thresholdValid() return 3 different values (1 for valid, 0 for invalid out of bounds, -1 for invalid too many decimals) and then compare with desired value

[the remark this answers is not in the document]
bastilian: Why not make it so it is actually using two different props in the state?

src/SmartComponents/CreatePolicy/validate.js
```diff
@@ -6,13 +6,27 @@ export const validateBenchmarkPage = (benchmark, osMajorVersion, profile) => {
...
+};
+
+export const thresholdValid = (threshold) => {
+  const parsedThreshold = parseFloat(threshold);
+
+  if (!isShortFloat(parsedThreshold)) return -1;
+
+  if (parsedThreshold < 101 && parsedThreshold >= 0) return 1;
+
+  return 0;
```
bastilian, line 22: This function, if we keep using it, should really only return a/one boolean. 

From the outside it is not obvious what any integer returned from this function means.
It would require going through the whole function to know what each integer means.
Even though this function is relatively small we should still apply the same thinking.

src/SmartComponents/EditPolicy/EditPolicyDetailsTab.js
```diff
@@ -21,6 +22,9 @@ export const useThresholdValidate = () => {
 
 const EditPolicyDetailsTab = ({ policy, setUpdatedPolicy }) => {
   const [validThreshold, validateThreshold] = useThresholdValidate();
+  const [currentThreshold, setCurrentThreshold] = useState(
```
bastilian, line 25: The threshold is already kept in the policy.

Where you use `currentThreshold` you should be able to use `policy.complianceThreshold`.

src/SmartComponents/CreatePolicy/CreatePolicy.js
```diff
@@ -141,7 +141,7 @@ export default connect((state) => ({
   osMajorVersion: selector(state, 'osMajorVersion'),
   osMinorVersionCounts: selector(state, 'osMinorVersionCounts'),
   businessObjective: selector(state, 'businessObjective'),
-  complianceThreshold: selector(state, 'complianceThreshold') || '100.0',
+  complianceThreshold: selector(state, 'complianceThreshold'),
```
bastilian, line 144: Why remove the default?

xmicha82: because it caused an empty compliance threshold to be valid, should this be left in?

src/SmartComponents/EditPolicy/EditPolicyDetailsTab.js
xmicha82: `policy.complianceThreshold` seems to always contain the value, which was passed in when the modal was created and it does not update with changes to the input field (although it should, through the `setUpdatedPolicy` hook?)

bastilian: AH! My bad, sorry! This is a bit confusing here. There is an `updatedPolicy`, which we'll need to pass down to the `EditPolicyDetailsTab`, which will contain the new **updated** threshold, if the threshold is updated through the `setUpdatedPolicy`.

src/SmartComponents/EditPolicy/EditPolicyDetailsTab.js
```diff
@@ -103,6 +112,14 @@ EditPolicyDetailsTab.propTypes = {
       propTypes.number,
     ]),
   }),
+  updatedPolicy: propTypes.shape({
+    description: propTypes.string,
+    businessObjective: propTypes.object,
+    complianceThreshold: propTypes.oneOfType([
+      propTypes.string,
```
bastilian, line 119: A little suspicious, this is. We should probably look into where this is coming from as a string and ensure it will be passed as a number, to avoid any issues like:

Screenshot 2021-11-15 at 13 41 07
d 

:)

bastilian (approved):
Awesome work! Thank you @xmicha82!

bastilian:
@xmicha82 Please, just a reminder to "Squash and Merge" (or squash locally before) so that the commit linter (currently failing on Travis) will pass.

PR #1432, opened 2021-10-28
feat(SystemsTable): RHICOMPL-1808 - Use osVersions meta field for OS filter

bastilian:
This changes the OS filter on systems tables to only show OS version that are present in the set of systems, instead of all possible versions.

Screenshot 2021-11-04 at 10 56 18

**How to test**

1. Visit any page with a SystemsTable
2. Check the OS filter and verify only OS version are shown that have systems in the table

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/constants.js
+++ b/src/SmartComponents/SystemsTable/constants.js
@@ -170,6 +170,18 @@ export const GET_SYSTEMS_TAGS = gql`
   }
 `;
 
+export const GET_SYSTEMS_OSES = gql`
+  query getSystems($filter: String!) {
+    systems(search: $filter) {
+      osVersions {
+        name
+        major
+        minor
+      }
+    }
+  }
+`;
+
 export const policyFilter = (policies, osFilter) => [
   ...systemsPolicyFilterConfiguration(policies),
   ...(osFilter ? systemsOsFilterConfiguration(policies) : []),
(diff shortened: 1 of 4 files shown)


src/SmartComponents/SystemsTable/constants.js
```diff
@@ -170,6 +170,18 @@ export const GET_SYSTEMS_TAGS = gql`
   }
 `;
 
+export const GET_SYSTEMS_OSES = gql`
+  query getSystems($filter: String!) {
+    systems(search: $filter) {
+      osVersions {
+        name
```
akofink, line 177: I don't think we're using `name` anywhere, not that it really hurts to request it.

akofink (approved):

#### Some observations:

When the table is empty, the OS version filter is also empty, which makes sense; however, it still gives a dropdown:
image
Check out the tags filter in this same empty state - it gives a message "No tags available":
image

bastilian:
> When the table is empty, the OS version filter is also empty, which makes sense; however, it still gives a dropdown:

I'll look into that in another PR. We should definitely show something similar as the tags filter does!

PR #1446, opened 2021-11-03
fix(TableTools): RHICOMPL-2432 - Apply filter and sorting to export

bastilian:
This fixes an issue where filters won't be applied to the export.

*How to test?*

1. Visit any page that contains a rules table
2. Export a CSV/JSON and verify all rules are included
3. Apply filters
4. Export CSV/JSON and verify that only the filtered result is exported

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Utilities/hooks/useTableTools/helper.js
+++ b/src/Utilities/hooks/useTableTools/helper.js
@@ -1,2 +1,7 @@
 export const filterSelected = (items, selectedIds = []) =>
   items.filter((item) => selectedIds.includes(item.itemId));
+
+export const filteredAndSortedItems = (items, filter, sorter) => {
+  const filtered = filter ? filter(items) : items;
+  return sorter ? sorter(filtered) : filtered;
+};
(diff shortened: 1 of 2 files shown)


akofink (approved):
Works great! Thanks, @bastilian!

Some other things I observed (hopefully it is as expected/ok):
- For edit policy rules tables and policy creation wizard rules tables, the export is disabled but still visible.
- When disabled, the export icon shows as grayed out and does not have any tooltip or other indication of what it is

bastilian:
> For edit policy rules tables and policy creation wizard rules tables, the export is disabled but still visible.

That is a bug. :) Nice catch! It doesn't pass the proper items, so when filtering for "selected" it will result in no items for the export. I tried fixing it here, but I'll open a Jira for it and solve this in a separate PR, since it might involve a bit more investigation and thinking.

> When disabled, the export icon shows as grayed out and does not have any tooltip or other indication of what it is

The only case that it should be disabled is, i believe only when there are not items in the table and there is nothing to export. In this case it should be somewhat obvious why it is disabled. We could still add a tooltip at any time.

PR #1450, opened 2021-11-03
Fix env vars for inventory frontend

rvsia:
cc @bastilian


bastilian (approved):
Nice catch! Thank you @rvsia!

PR #1457, opened 2021-11-05
feat(devtools): Setup commitlint

bastilian:
This sets up `commitlint`[0] and proposes rules to lint commit messages.

It adds `npm run lint:commit` as a command to validate commits made in a branch compared `master`.

[0] https://commitlint.js.org/

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


skateman:
@bastilian can you elaborate on what rules this linter enforces?

bastilian:
@skateman Of course. The (default) rules are following the conventional commit[0] format, that we already use.

 Since we do have some deviation already there are two customised rules. 
The `body-max-line-length`, because dependabots commits have longer lines in the body.
And `subject-case`, because we tend to use component names as scope/subject and the default expects snake-case.

[0] https://www.conventionalcommits.org/en/v1.0.0/

commitlint.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+  rules: {
+    'body-max-line-length': [2, 'always', 150],
```
vkrizan, line 4: How we came up with 150 chars?

How about:
```
[1, 'always', 80]
```
(warning after 80 chars)

Commits can contain URLs that are not breakable.

bastilian: This is only to satisfy commits by dependabot, we can of course switch to warning after 80.

marleystipich2 (commented):
I've tested this and commitlint is working great for me!
Quick question though @bastilian, I'm not seeing an error or warning message when I use a subject line that is all caps, which I think should go against the rules, like `git commit -m "fix: TEST"`. Is this intentional?
I am seeing the correct warning message when I test a body line length over 80 though.

bastilian:
@marleystipich2 Thank you for testing! Yes, this is intentionally disabled via the `subject-case` setting and allows any case, instead of the default, which is only lower-case. The reason for disabling it is to allow us to use the Jira card ID, which is upper-case and not trigger the rule. There doesn't seem to be any other way to allow these.

marleystipich2 (approved):
Oh good idea! Thanks for the explanation, @bastilian. This works great for me.

skateman:
@vkrizan are you okay with this?

bastilian:
> @vkrizan are you okay with this?

We can always adjust the settings. The current settings are to let through what we have used do far. I took the last 100 commits or so to validate and make sure the pass.

PR #1460, opened 2021-11-05
feat(ComplianceDetails): RHICOMPL-1136 - Implement policy card filtering

bastilian:
https://user-images.githubusercontent.com/7757/145012668-957d020c-577f-47cf-b4d6-6973d9efcb6d.mp4

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -34,7 +34,7 @@ const RulesTable = ({
   const policies = profileRules
     .filter(({ profile }) => !!profile)
     .map(({ profile }) => ({
-      id: profile.policy ? profile.policy.id : profile.id,
+      id: profile.id,
       name: profile.name,
     }));
 
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -22,6 +22,7 @@ const RulesTable = ({
   selectedRules = [],
   hidePassed = false,
   options,
+  activeFilters,
   ...rulesTableProps
 }) => {
   const manageColumnsEnabled = useFeature('manageColumns');
@@ -64,9 +65,13 @@ const RulesTable = ({
           remediationAvailableFilter,
         }),
         ...(hidePassed && {
-          activeFilters: {
-            passed: ['failed'],
-          },
+          activeFilters: (currentActiveFilters) => ({
+            ...currentActiveFilters,
+            passed: currentActiveFilters.passed
+              ? currentActiveFilters.passed
+              : ['failed'],
+            ...activeFilters,
+          }),
         }),
       }}
       options={{
@@ -99,6 +104,7 @@ RulesTable.propTypes = {
   handleSelect: propTypes.func,
   columns: propTypes.array,
   options: propTypes.object,
+  activeFilters: propTypes.object,
 };
 
 export default RulesTable;
--- a/src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
+++ b/src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
@@ -33,7 +33,7 @@ class SystemPolicyCard extends React.Component {
     value.toFixed(fixed) + (withPercent ? '%' : '');
 
   render() {
-    const { policy, style } = this.props;
+    const { policy, style, onClick, isSelected } = this.props;
     const {
       rulesFailed,
       compliant,
@@ -48,7 +48,15 @@ class SystemPolicyCard extends React.Component {
     const truncateDefaults = { expandOnMouseOver: true, hideExpandText: true };
 
     return (
-      <Card style={style}>
+      <Card
+        onClick={(event) => {
+          event.stopPropagation();
+          onClick(policy);
+        }}
+        isSelectable
+        isSelected={isSelected}
+        style={style}
+      >
         <CardBody>
           <TextContent className="margin-bottom-md">
             <Text
@@ -124,6 +132,8 @@ SystemPolicyCard.propTypes = {
     supported: PropTypes.bool,
   }),
   style: PropTypes.object,
+  onClick: PropTypes.func,
+  isSelected: PropTypes.bool,
 };
 
 export default SystemPolicyCard;
--- a/src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
+++ b/src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
@@ -49,6 +49,7 @@ class SystemPolicyCard extends React.Component {
 
     return (
       <Card
+        ouiaId="PolicyCard"
         onClick={(event) => {
           event.stopPropagation();
           onClick(policy);
@@ -60,12 +61,14 @@ class SystemPolicyCard extends React.Component {
         <CardBody>
           <TextContent className="margin-bottom-md">
             <Text
+              ouiaId="PolicyCardName"
               className="margin-bottom-top-none"
               component={TextVariants.h4}
             >
               <Truncate text={name} length={110} {...truncateDefaults} />
             </Text>
             <Text
+              ouiaId="PolicyCardType"
               style={{ color: 'var(--pf-global--Color--200)' }}
               component={TextVariants.small}
             >
@@ -74,7 +77,10 @@ class SystemPolicyCard extends React.Component {
           </TextContent>
           <div className="margin-bottom-md">
             {supported && this.complianceIcon(compliant)}
-            <Text component={TextVariants.small}>
+            <Text
+              ouiaId="PolicyCardFailedRulesScore"
+              component={TextVariants.small}
+            >
               {rulesFailed} rule{rulesFailed === 1 ? '' : 's'} failed{' '}
               <Tooltip
                 position="bottom"
@@ -89,8 +95,8 @@ class SystemPolicyCard extends React.Component {
             </Text>
           </div>
           <Text className="margin-bottom-none" component={TextVariants.small}>
-            <Text>SSG version: {ssgVersion}</Text>
-            <Text>
+            <Text ouiaId="PolicyCardSSGVersion">SSG version: {ssgVersion}</Text>
+            <Text ouiaId="PolicyCardLastScanned">
               Last scanned:{' '}
               {lastScanned !== 'Never' ? (
                 <DateFormat date={Date.parse(lastScanned)} type="relative" />
@@ -103,6 +109,7 @@ class SystemPolicyCard extends React.Component {
         {!supported && (
           <CardFooter style={{ padding: '0' }}>
             <UnsupportedSSGVersion
+              ouiaId="PolicyCardUnsupportedSSG"
               ssgVersion={ssgVersion}
               style={{
                 paddingTop: 'var(--pf-c-alert--PaddingTop)',
--- a/src/PresentationalComponents/SystemPolicyCards/SystemPolicyCards.js
+++ b/src/PresentationalComponents/SystemPolicyCards/SystemPolicyCards.js
@@ -6,12 +6,20 @@ import { Instagram } from 'react-content-loader';
 
 class SystemPolicyCards extends React.Component {
   systemPolicyCards() {
-    const { policies } = this.props;
+    const { policies, selectedPolicies, onCardClick } = this.props;
+
     return policies
       .filter((policy) => policy.rulesFailed + policy.rulesPassed > 0)
       .map((policy, i) => (
         <GridItem sm={12} md={12} lg={6} xl={4} key={i}>
-          <SystemPolicyCard policy={policy} style={{ height: '100%' }} />
+          <SystemPolicyCard
+            policy={policy}
+            style={{ height: '100%' }}
+            onClick={(policy) => onCardClick(policy)}
+            isSelected={selectedPolicies?.find(
+              (policyId) => policyId === policy.id
+            )}
+          />
         </GridItem>
       ));
   }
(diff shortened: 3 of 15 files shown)


src/Utilities/hooks/useTableTools/useFilterConfig.js
```diff
@@ -17,13 +17,27 @@ const filterValues = (activeFilters) =>
...
+  const { filters, setPage, selectedFilter, onDeleteFilter } = options;
   const enableFilters = !!filters;
-  const { filterConfig = [], activeFilters: initialActiveFilters } =
+  const { filterConfig = [], activeFilters: initialActiveFiltersRaw } =
     filters || {};
-  const [activeFilters, setActiveFilters] = useState(
-    filterConfigBuilder.initialDefaultState(initialActiveFilters, filterConfig)
+
+  const [activeFilters, setActiveFilters] = useState({});
+  const initialActiveFilters = perpareInitialActiveFilters(
```
vkrizan, line 38: The name convention of "initial" and "active" is perplexing. "active" are the ones which state is being held here, computed out of "initial" which are being sent via props. Is that correct? If so, then they are no longer initial (in a sense of lifecycle of a component). We might think of refactoring the state management.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
```diff
@@ -141,16 +141,17 @@ class FilterConfigBuilder {
     }
   };
 
-  initialDefaultState = (defaultStates = {}, initConfig) => {
-    let initialState = {};
-    (initConfig || this.config).forEach((filter) => {
+  initialDefaultState = (defaultStates = {}, initConfig) =>
```
vkrizan, line 144: Off-topic: Is there a eslint rule that can warn about lack of parenthesis? Having a function on several lines without parenthesis does not look good with the code visual flow. @skateman thoughts?

vkrizan (commented):
Looks awesome Sebastian! I might give it for a spin.

I left a question for explanation and some other suggestions down below.

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -62,9 +63,13 @@ const RulesTable = ({
...
         ...(hidePassed && {
-          activeFilters: {
-            passed: ['failed'],
-          },
+          activeFilters: (currentActiveFilters) => ({
+            ...currentActiveFilters,
+            passed: currentActiveFilters.passed
+              ? currentActiveFilters.passed
+              : ['failed'],
+            ...activeFilters,
```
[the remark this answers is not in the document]
bastilian: `activeFilters` is what `ComplianceDetails` uses to set the policies, `currentActiveFilters` is what `useFilterConfig` has currently in state. We need to merge the two here to make sure the passed filter is set or not thrown away.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
bastilian: @vkrizan Most of these functions that don't have parenthesis and use implicit return are relatively small 5-10 and as long as they are within that and are indented they are fine to me. They are also most of the time only a loop.

src/Utilities/hooks/useTableTools/useFilterConfig.js
bastilian: @vkrizan We will refactor the whole filter management and building eventually, not only it's state management.

src/Utilities/hooks/useTableTools/useFilterConfig.js
```diff
@@ -82,7 +98,7 @@ const useFilterConfig = (options = {}) => {
     return () => {
       filterConfigBuilder.config = [];
     };
-  }, []);
+  }, [JSON.stringify(initialActiveFilters)]);
```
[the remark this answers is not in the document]
bastilian: > Note, that JSON encode heavy relatively

Benchmarks? :)

We do this in few places where we do, but shouldn't use it. Until we rewrite these places to use non complex values this has to do.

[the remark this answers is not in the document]
vkrizan: Okay. Is there a way to avoid it here for useEffect here?

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
skateman: I agree with @vkrizan large ones should have `()` but I'd be okay with a single-argument one not having them (lambda)

skateman: Let me correct myself, doesn't have to by a single line, but a single (and readable) expression

vkrizan: Implicit returns can be done with parenthesis, hence my suggestion to format it this way:
```
const singleOnelineExpression = (ar) => ar.map(...);
const multilineFn => () => (
  ...
  ...
);
```
The overall "flow" is a bit better.

src/PresentationalComponents/SystemPolicyCards/SystemPolicyCards.js
```diff
@@ -5,24 +5,38 @@ import propTypes from 'prop-types';
...
-        <Grid hasGutter>
-          {this.systemPolicyCards()}
+        <Grid
+          hasGutter
+          onClick={(event) => {
+            event.stopPropagation();
+            onCardClick();
+          }}
+        >
+          {this.systemPolicyCards(onCardClick)}
```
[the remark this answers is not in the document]
bastilian: Updated the `this.systemPolicyCards` function and removed it here.

src/PresentationalComponents/SystemPolicyCards/SystemPolicyCards.js
```diff
@@ -5,24 +5,38 @@ import propTypes from 'prop-types';
...
+    const { loading, onCardClick } = this.props;
 
     return (
       <React.Fragment>
-        <Grid hasGutter>
-          {this.systemPolicyCards()}
+        <Grid
+          hasGutter
+          onClick={(event) => {
+            event.stopPropagation();
```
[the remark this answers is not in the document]
bastilian: Nothing, this was remnants of an earlier revision. It's not even needed.

src/Utilities/hooks/useTableTools/useFilterConfig.js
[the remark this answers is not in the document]
bastilian: @vkrizan Not easily atm.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
bastilian: We will not enable any rule, regardless of anyones opinion. We decided in the past to stick with common rules provided by the `eslint-config` package from the frontend-components. 

Significant whitespace works in Python, should work in JavaScript to.

src/PresentationalComponents/RulesTable/RulesTable.js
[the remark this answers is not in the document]
vkrizan: Thanks. Where is the passed filter being reset? I see that this combines two different types for a very concrete situation. One type (`activeFilters`) is "configuration" and second `currentActiveFilters` is a state that is a superset of what configuration provides.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
```diff
@@ -141,16 +141,17 @@ class FilterConfigBuilder {
...
+    (initConfig || this.config).reduce((acc, filter) => {
       const filterStateName = stringToId(filter.key || filter.label);
-      initialState[filterStateName] =
-        defaultStates[filterStateName] || this.defaultValueForFilter(filter);
-    });
-
-    return initialState;
-  };
+      const defaultState = () => {
```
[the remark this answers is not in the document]
bastilian: Removed it.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
vkrizan: I would respectfully disagree. We, as team, iterate on existing processes to alter them for better experience and maturity. That's one the traits of Agile methodology.

`eslint-config` does not limit us to add new rules if they do not conflict.

src/PresentationalComponents/RulesTable/RulesTable.js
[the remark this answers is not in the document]
bastilian: The passed filter is either set initially, when loading to 'failed'. When a user interacts with the filter in the toolbar, this can change. To make sure that change is not lost when a user set a filter via a policy card, it uses the value in `currentActiveFilters` here.

src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
vkrizan: They do conflict :disappointed: 

```
src/Utilities/hooks/useTableTools/FilterConfigBuilder/FilterConfigBuilder.js
  144:60  error  Delete `·(`    prettier/prettier
  152:10  error  Delete `)⏎··`  prettier/prettier

```

LightOfHeaven1994:
FYI, played with filtering a bit locally, everything worked to me

src/PresentationalComponents/RulesTable/RulesTable.js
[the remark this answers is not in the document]
vkrizan: Okay, that make sense. Is there a possibility to cover all cases, not just "passed" from `activeFilters`. Or is that the only one that needs to have it covered. If not, then any future changes and/or additions to `activeFilters` would have to have this treatment or a new bug would arise.

Since the `activeFilters` are the initial value (I presume set as prop) wouldn't it be safer to assign them directly to the state as it's initial value (either precomputed or not): `useState(preCompute(activeFilters))`? This would ensure that further state updates would no longer be affected by `activeFilters` for the lifecycle of a component. If it needs to be changed, therefore no longer being an initial but rather state, it would need other means to do so (like propagating the setter up). The setter is indeed propagated with the introduction of the callback on activeFilters. What I'm still missing is the why it needs to blend the initial value if the state should already have it in the first place.

Of course that some parts of the filter would likely need some treatment (due to naturality of object blending). What seems improper here is the state management, not necessarily the passed filter functionality.

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
```diff
@@ -48,7 +48,15 @@ class SystemPolicyCard extends React.Component {
     const truncateDefaults = { expandOnMouseOver: true, hideExpandText: true };
 
     return (
-      <Card style={style}>
+      <Card
+        onClick={(event) => {
+          event.stopPropagation();
+          onClick(policy);
+        }}
```
LightOfHeaven1994, line 55: It would be useful to have OUIA IDS for:
Card itself -  ouiaId="PolicyCard"
Policy card name - (Text) ouiaId="PolicyCardName"
Policy card type - (Text) ouiaId="PolicyCardType"
Failed rules and score - (Text) ouiaId="PolicyCardFailedRulesScore"
SSG version - (Text) ouiaId="PolicyCardSSGVersion"
Last scanned - (Text) ouiaId="PolicyCardLastScanned"
Unsupported SSG alert - (UnsupportedSSGVersion -> Alert, seems here we need to send it as a prop) ouiaId="PolicyCardUnsupportedSSG"

src/SmartComponents/SystemDetails/ComplianceDetail.js
```diff
@@ -57,44 +57,92 @@ const QUERY = gql`
...
-        }}
-        profileRules={system?.testResultProfiles.map((profile) => ({
-          system,
-          profile,
-          rules: profile.rules,
-        }))}
+const SystemQuery = ({ data: { system }, loading, hidePassed }) => {
+  const [selectedPolicies, setSelectedPolicies] = useState();
+  const policies = system?.testResultProfiles;
+  const setOrUnsetPolicy = (policy) => {
```
vkrizan, line 63: I'm getting some Policy mismatch:
Screenshot 2021-12-09 at 11-44-41  Systems - Compliance Red Hat Insights

It shows hashes and probably sets wrong ID value, which also leads unlink between Policy filter and cards selection:
Screenshot 2021-12-09 at 11-45-14 Systems - Compliance Red Hat Insights

I'll investigate further, and probably start on fresh data set.

vkrizan: Yes, still the same issue on fresh db with seeded data by `bundle exec rake dev:db:seed`. @skateman is the seeding creating an incorrect data that creates this policy/profile difference?

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
bastilian: 👍 Will add them!

src/SmartComponents/SystemDetails/ComplianceDetail.js
skateman: Yeah, seeded data might create incorrect/unsupported stuff :thinking:

bastilian:
@LightOfHeaven1994  OUIA IDs added. 

@skateman Is the issue Viliam has found relevant or is it due to improper seed data?

skateman:
@bastilian yes

skateman (approved):
The Seal of Approval

PR #1474, opened 2021-11-15
feat(TableTools): RHICOMPL-2544 - Implement column manager

bastilian:
This implements a column manager for tables to use and is enabled for the policies, reports and rules table.

https://user-images.githubusercontent.com/7757/146182240-bbe11b8a-319d-4771-9fcf-0bf4fed46e58.mp4

⚠️ It is currently behind a feature flag. To enable append `?manageColumns=enable` to a URL showing

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/Utilities/hooks/useTableTools/Components/ColumnManager.js
```diff
@@ -0,0 +1,94 @@
...
+import ColumnManagerListItem from './ColumnManagerListItem';
+
+const ColumnManager = ({
+  isOpen,
+  columns,
+  onSave: onSaveArgument,
+  onClose,
+  selectedColumns: selectedColumnsProp = [],
+}) => {
+  const [selectedColumns, setSelectedColumns] = useState(selectedColumnsProp);
```
vkrizan, line 20: Should the selection persist if the modal is canceled?

If not, then a change of `selectedColumnsProp` would not change the internal state of this modal once it is set. Setting a new internal state (in ColumnManager) could then be tied to modal open or there can be a watch on `selectedColumnsProp` changes.

bastilian: @vkrizan No it should not persist when canceling. It now is reset when canceling.

vkrizan: I see. The reset is now made during `onClose`. Would it be safer to do reset it onOpen instead? The prop (upper state) can change during the modal being closed.

bastilian: @vkrizan I'm not sure what you mean with "safer"?

bastilian: There is also no `onOpen`. We'd have to use a `use(Layout)Effect` and depend on the `isOpen`, but it feels more straightforward just to call it `onClose`.

vkrizan: > I'm not sure what you mean with "safer"?

By safer I've meant that the right state is set the right time. Setting it on modal close isn't as precise as on modal open. Therefore `isOpen` watched with `use(Layout)Effect` could be a better option. As mentioned, the state can change when the modal is closed, for example if there would be an inline column selection change/delete. In that case, the modal would not get the update.

bastilian: >  if there would be an inline column selection change/delete

What is that?

vkrizan: An example of column change outside of column manager modal. I meant that if there were a X (cross) on each column that can remove it from the view. Sorting columns can be also another example that can be done directly on the table outside of the modal.

bastilian: @vkrizan At the moment we don't do those things. In theory we could argue that we'd need to watch every argument given to a hook and put it in a state and update it on change of the argument, but it is only useful/needed in certain cases.

Currently, if the columns for a (TableTools) table change the whole component should re-render and call the hook(s) with the new columns.

vkrizan: > Currently, if the columns for a (TableTools) table change the whole component should re-render and call the hook(s) with the new columns.

The hook would be called on re-render, however the previous state would persist. State persists for the lifecycle of a component. New state story is created for new components, therefore the old one would have to be unmounted and destroyed first (which is not what's happening on re-render).

My concern is scoped to ColumnManager and its state management. At the moment the state change on modal cancel is insufficient and does not account for future changes. The suggestion to move it on modal open seems more logical than the status quo (for the reasons described). Do you foresee any issues with it?

bastilian: @bastilian I don't see any issues now, if there are issues with it in the future we can address them then.

src/Utilities/hooks/useTableTools/Components/__snapshots__/TableToolsTable.test.js.snap
```diff
@@ -43,6 +43,13 @@ exports[`TableToolsTable expect to have filters properly rendered 2`] = `
 exports[`TableToolsTable expect to render without error 1`] = `
 <Fragment>
   <PrimaryToolbar
+    actionsConfig={
+      Object {
+        "actions": Array [
+          undefined,
```
vkrizan, line 49: Is this correct that `undefined` is returned?

src/Utilities/hooks/useTableTools/Components/ColumnManagerListItem.js
```diff
@@ -0,0 +1,48 @@
...
+          checked={isSelected}
+          onChange={onChange}
+          otherControls
+          isDisabled={isDisabled}
+        />
+      </DataListControl>
+      <DataListItemCells
+        dataListCells={[
+          <DataListCell key={`${key}-cell`}>
+            <label>{column.title}</label>
```
vkrizan, line 32: Please tie the label to the checkbox. A click on the label should toggle the checkbox (accessibility).

vkrizan (commented):
Works well so far.

Another behavioral issue that I've found is when having only one column selected then the kebab (or other icon) is not right aligned. I'd expect the only column to fit and the kebab menu to be right aligned:
Screenshot 2021-12-21 at 15-23-07 SCAP policies - Compliance Red Hat Insights

The Systems view (`/insights/compliance/systems`) and Report's systems (`/insights/compliance/reports/<uuid>`) lack the column manager. Would we add it there as a as a flow up?

src/Utilities/hooks/useTableTools/Components/__snapshots__/TableToolsTable.test.js.snap
bastilian: Yes, this is intended and correct.

src/Utilities/hooks/useTableTools/Components/ColumnManager.js
```diff
@@ -0,0 +1,96 @@
...
+      isOpen={isOpen}
+      variant="small"
+      onClose={onClose}
+      description={description}
+      actions={[
+        <Button key="save" variant="primary" onClick={onSave}>
+          Save
+        </Button>,
+        <Button key="cancel" variant="secondary" onClick={onClose}>
+          Cancel
```
vkrizan, line 61: Since the change of manager to be stateless makes the changes live, the Cancel button is no longer needed. The Save button can be changed to `Close`.

However, I believe that UX behavior should be to change the columns on Save only. We'd need to verify that, so it is consistent across the platform. It is possible to do with stateless manager, by exposing events up and keeping track of the second state on the upper level.

bastilian:
@vkrizan The kebab menu not right aligned is not really an issue with this feature, but a general styling issue. We can address that separately.

The `SystemsTable` will need to be worked on more and separately as it might be a bit more difficult to integrate with the `InventoryTable`.

src/Utilities/hooks/useTableTools/Components/ColumnManager.js
bastilian: Removed that commit.

skateman (approved):
The Seal of Approval

PR #1475, opened 2021-11-15
Update filters to support selectable groups

karelhala:
### OS conditional filter group

OS filter supports selectable groups. Currently this is broken because the conditional filter went trough some changes, this PR fixes such issue, bye sending the change event to parent component everytime and by mapping the selected values correctly.

* This feature will be fully supported once https://github.com/RedHatInsights/frontend-components/pull/1310 is merged. However since the conditional filter is part of inventory it's not required to wait for 1310 to be merged. We can freely merge this one and once FEC PR is merged and inventory updated it will start working as it should.

### Without frontend PR

Screenshot from 2021-11-15 10-59-27

### With updated PR

Screenshot from 2021-11-15 10-53-45

### DIFF
--- a/src/Utilities/hooks/useTableTools/FilterConfigBuilder/ChipBuilder.js
+++ b/src/Utilities/hooks/useTableTools/FilterConfigBuilder/ChipBuilder.js
@@ -44,7 +44,17 @@ class ChipBuilder {
 
         return Object.keys(currentValue[groupValue]).map((itemValue) =>
           currentValue[groupValue][itemValue]
-            ? group?.items.find(findWithString(itemValue))
+            ? [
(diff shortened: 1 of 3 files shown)


karelhala:
@bastilian btw there has been recently added new type of filter in group filter called `TreeView` it was designed and made specifically for this use case (collapsible items, conditional checkboxes, multiple nesting, etc.)

bastilian:
@karelhala Yeah, i know. For now the group filter was wanted for this. Maybe if we see that the list of values to filter becomes too big we could move to the TreeView.

bastilian (approved):
Thank you @karelhala! Tested it with fed mods locally.

PR #1491, opened 2021-12-01
change(Systems Table): RHICOMPL-2164 move OS into separate column

xmicha82:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/Columns.js
+++ b/src/SmartComponents/SystemsTable/Columns.js
@@ -1,4 +1,6 @@
+import React from 'react';
 import { nowrap } from '@patternfly/react-table';
+import { Tooltip } from '@patternfly/react-core';
 import { complianceScoreString } from 'PresentationalComponents';
 import { profilesRulesFailed } from 'Utilities/ruleHelpers';
 import { renderComponent } from 'Utilities/helpers';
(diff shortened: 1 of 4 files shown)


src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -136,6 +138,27 @@ export const OperatingSystem = compileColumnRenderFunc({
...
+export const OSHeader = () => {
+  return (
+    <Tooltip content={<div>Operating System</div>}>
+      <span>OS</span>
+    </Tooltip>
+  );
+};
+
+export const OS = compileColumnRenderFunc({
+  title: <OSHeader />,
```
bastilian, line 150: This might be doable by using the info transform that is mentioned here in the second example: https://www.patternfly.org/v4/components/table/react-legacy/#table-columns

xmicha82: That does not really do what is needed here. The tooltip transform creates the *questionmark-in-a-circle* tooltip, not the *hover-over* tooltip. Also, I was not able to find any transform, that would create the *hover-over* tooltip

bastilian: @xmicha82 There should be a way of defining what is displayed as a tooltip in the table header that does not require manually adding a `Tooltip` component. Maybe the 'tooltip' prop (which you probably need to add to the `props` prop) would work? https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/components/Table/HeaderCell.tsx#L22

bastilian: One thing small thing this will need is a `original` property set, to make sure the export functions properly:

Screenshot 2021-12-07 at 13 16 52

sthirugn:
@xmicha82 this is open for a while now, can you please look into this when you are back? Thanks!

bastilian (approved):
Awesome! Everything is now in order and working! Nice job @xmicha82!

PR #1492, opened 2021-12-02
fix(RulesTable): RHICOMPL-2479 - Adapt rule selection to include profile

bastilian:
https://github.com/RedHatInsights/compliance-frontend/commit/f1a64fb40df7a6aa7c3c34e6ecec052ed21c6373#diff-f3088209a94cef4be43a8666799ccd2cb4b52fed78c2d58f4e078d310bdba6a7 Introduced the profile as a prefix for the identifier to address selection issues in reports with multiple policies, but it introduced issues with preselecting items in the table, which would render the rules tables in the wizard and the edit policy empty.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/TabbedRules/ProfileTabContent.js
+++ b/src/PresentationalComponents/TabbedRules/ProfileTabContent.js
@@ -131,11 +131,19 @@ const ProfileTabContent = ({
             remediationsEnabled={false}
             columns={columns}
             profileRules={[{ profile, rules: rules || [] }]}
-            selectedRefIds={selectedRuleRefIds}
+            selectedRules={selectedRuleRefIds.map(
+              (refId) => `${profile.id}|${refId}`
+            )}
             handleSelect={
               handleSelect &&
               ((selectedRuleRefIds) =>
-                handleSelect(profile, newOsMinorVersion, selectedRuleRefIds))
+                handleSelect(
+                  profile,
+                  newOsMinorVersion,
+                  selectedRuleRefIds.map(
+                    (profileRuleRefId) => profileRuleRefId.split('|')[1]
+                  )
+                ))
             }
             {...rulesTableProps}
           />
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -18,12 +18,14 @@ const RulesTable = ({
   remediationAvailableFilter = false,
   selectedFilter = false,
   handleSelect,
-  selectedRefIds = [],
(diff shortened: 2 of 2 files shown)


LightOfHeaven1994 (approved):
Tested myself - Create Policy and Edit policy Rules view looks good! Tried multiversion with couple Rule tabs, everything works. Ack

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
```diff
@@ -131,11 +131,19 @@ const ProfileTabContent = ({
             remediationsEnabled={false}
             columns={columns}
             profileRules={[{ profile, rules: rules || [] }]}
-            selectedRefIds={selectedRuleRefIds}
+            selectedRules={selectedRuleRefIds.map(
+              (refId) => `${profile.id}|${refId}`
```
vkrizan, line 135: Can it produce an object or an array instead of arbitrary concating?

bastilian: Not yet.

vkrizan: Can you please elaborate? Is the string required somewhere?

bastilian: @vkrizan It would be tricky to make everything work with an object and compare the items in the table with the objects in the selection.

vkrizan: I see. UUIDs would be a better candidate for general selection. RefIds would then have to be gathered or kept for the API calls made to update or create tailoring. I believe that APIs can take UUIDs as well. @skateman This sounds like a tech debt.

vkrizan (commented):
The fix looks good.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
bastilian: @vkrizan In this case even with the uuid of the rule, I think it would cause the same issues we've seen with just using the refId.

vkrizan: True. It is indeed many to many relationship.

PR #1498, opened 2021-12-08
feat: RHICOMPL-2400 Update GQL endpoint for supported profiles

marleystipich2:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

Converting this to draft because still in progress.

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -107,6 +107,7 @@ export const CreateSCAPPolicy = ({ change, selectedOs, selectedProfile }) => {
   const setOsMajorVerson = (osMajorVersion) => {
     if (selectedOs !== osMajorVersion.node.osMajorVersion) {
       change('systems', []);
+      change('profile', undefined);
     }
 
     change('osMajorVersion', osMajorVersion.node.osMajorVersion);
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -22,20 +22,23 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import propTypes from 'prop-types';
 
-const BENCHMARKS_AND_PROFILES = gql`
-  query benchmarksAndProfiles {
-    latestBenchmarks {
-      id
-      title
-      refId
-      version
-      osMajorVersion
-      profiles {
-        id
-        name
-        refId
-        description
-        complianceThreshold
+const SUPPORTED_PROFILES_BY_OS_MAJOR = gql`
+  query supportedProfilesByOSMajor {
+    osMajorVersions {
+      edges {
+        node {
+          osMajorVersion
+          supportedProfiles {
+            id
+            name
+            refId
+            description
+            complianceThreshold
+            benchmark {
+              id
+            }
+          }
+        }
       }
     }
     profiles(search: "external = false and canonical = false") {
@@ -44,7 +47,7 @@ const BENCHMARKS_AND_PROFILES = gql`
           id
           refId
           benchmark {
-            refId
+            osMajorVersion
           }
         }
       }
@@ -62,18 +65,16 @@ const PolicyTooltip = () => (
   </Tooltip>
 );
 
-export const CreateSCAPPolicy = ({
-  change,
-  selectedBenchmarkId,
-  selectedProfile,
-}) => {
-  const { data, error, loading } = useQuery(BENCHMARKS_AND_PROFILES, {
+export const CreateSCAPPolicy = ({ change, selectedOs, selectedProfile }) => {
+  const { data, error, loading } = useQuery(SUPPORTED_PROFILES_BY_OS_MAJOR, {
     fetchPolicy: 'no-cache',
   });
 
-  const inUseProfileRefIds = (profiles, benchmark) =>
+  const inUseProfileRefIds = (profiles, osMajorVersion) =>
     profiles
-      .filter((profile) => benchmark.refId === profile.node.benchmark.refId)
+      .filter(
+        (profile) => osMajorVersion == profile.node.benchmark.osMajorVersion
+      )
       .map((profile) => profile.node.refId);
 
   if (error) {
@@ -84,30 +85,31 @@ export const CreateSCAPPolicy = ({
     return <Spinner />;
   }
 
-  const benchmarks = data.latestBenchmarks;
-  let selectedBenchmark;
+  const osMajorVersions = data.osMajorVersions?.edges;
+  let selectedOsMajorVersion;
   let validProfiles;
-  if (selectedBenchmarkId) {
-    selectedBenchmark = benchmarks.find(
-      (benchmark) => benchmark.id === selectedBenchmarkId
+  if (selectedOs) {
+    selectedOsMajorVersion = osMajorVersions.find(
+      (os) => os.node.osMajorVersion === selectedOs
     );
     const userProfileRefIds = inUseProfileRefIds(
       data.profiles.edges,
-      selectedBenchmark
+      selectedOs
     );
-    validProfiles = selectedBenchmark.profiles.map((profile) => ({
-      ...profile,
-      disabled: userProfileRefIds.includes(profile.refId),
-    }));
+    validProfiles = selectedOsMajorVersion?.node.supportedProfiles.map(
+      (profile) => ({
+        ...profile,
+        disabled: userProfileRefIds.includes(profile.refId),
+      })
+    );
   }
 
-  const setBenchmark = ({ id, osMajorVersion }) => {
-    if (selectedBenchmark?.osMajorVersion !== osMajorVersion) {
+  const setOsMajorVerson = (osMajorVersion) => {
+    if (selectedOs !== osMajorVersion.node.osMajorVersion) {
       change('systems', []);
     }
 
-    change('benchmark', id);
-    change('osMajorVersion', osMajorVersion);
+    change('osMajorVersion', osMajorVersion.node.osMajorVersion);
   };
 
   return (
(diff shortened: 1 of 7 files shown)


src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -22,8 +22,8 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import propTypes from 'prop-types';
 
-const BENCHMARKS_AND_PROFILES = gql`
-  query benchmarksAndProfiles {
+const BENCHMARKS_AND_PROFILES_AND_OS_MAJOR_VERSIONS = gql`
+  query benchmarksAndProfilesAndOsMajorVersions {
     latestBenchmarks {
```
skateman, line 27: You won't need this, all the data can be loaded via the query you're adding.

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -38,6 +38,20 @@ const BENCHMARKS_AND_PROFILES = gql`
...
+    osMajorVersions {
+      edges {
+        node {
+          osMajorVersion
+          supportedProfiles {
+            id
+            name
+            refId
+            description
+            complianceThreshold
```
skateman, line 50: You can also load the benchmark version here as I have it in the issue in the sample query.

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -67,9 +81,12 @@ export const CreateSCAPPolicy = ({
   selectedBenchmarkId,
   selectedProfile,
 }) => {
-  const { data, error, loading } = useQuery(BENCHMARKS_AND_PROFILES, {
-    fetchPolicy: 'no-cache',
-  });
+  const { data, error, loading } = useQuery(
+    BENCHMARKS_AND_PROFILES_AND_OS_MAJOR_VERSIONS,
```
skateman, line 85: `SUPPORTED_PROFILES_BY_OS_MAJOR`

skateman (changes requested):

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
marleystipich2: Okay, yeah I did have it in there but I removed it because I wasn't using it but I'll add it back in there in case we ned it for something

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
marleystipich2: Sorry, which part isn't needed? Do you mean I can load all of the benchmark data in the osMajorVersions query?

skateman: Yes, you only need the benchmark version I believe and that's possible different for each profile. Part of the data coming from this query is obsolete (latest benchmarks) and the rest is a partial duplicate (latest profiles).

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
skateman: You need it, because the version for each profile might be different...

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
marleystipich2: Okay, yeah I see what you mean now. The only part I'm confused about is that latestBenchmark is still being used like this:
```
const benchmarks = data.latestBenchmarks;
...
 if (selectedBenchmarkId) {
    selectedBenchmark = benchmarks.find(
      (benchmark) => benchmark.id === selectedBenchmarkId
    );
```
Is there a way I can find the `selectedBenchmark` without latestBenchmark?

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -95,7 +112,15 @@ export const CreateSCAPPolicy = ({
       data.profiles.edges,
       selectedBenchmark
     );
-    validProfiles = selectedBenchmark.profiles.map((profile) => ({
+    const supportedProfilesForBenchmark = (osMajorVersions, benchmark) =>
+      osMajorVersions.filter(
```
bastilian, line 116: You can use `.find` here and avoid the `[0]`.

vkrizan: ... and please add safe navigation if the item isn't found or treat it other way. :slightly_smiling_face:

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
skateman: Yeah, the `selectedBenchmark` should come from under the (selected) profile, as it can be different for each.

marleystipich2: Oh okay that makes more sense. I'll work on that now!

marleystipich2:
@bastilian Can you take a look at this, when you get a chance? I made some updates that were needed but now, when I select an OS version and then a profile, the "Next" button stays disabled. So it seems like it thinks the form isn't fully filled out but I'm not sure why?

skateman:
@marleystipich2 you need to make [this](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/CreatePolicy.js#L55) condition pass in order to enable the next button, it's defined [here](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/validate.js#L1-L7)

marleystipich2:
@skateman Oh that helps a lot. Thanks!

skateman:
@marleystipich2 can you please address the issues in travis?

marleystipich2:
@skateman Okay, I fixed the failing tests. Let me know if everything looks good to you!

src/__fixtures__/benchmarks_rules.js
```diff
@@ -726,3 +726,120 @@ export const profileRefIdsQuery = {
...
+                        __typename: 'Profile'
+                    },
+                ],
+                __typename: 'OsMajorVersion',
+            },
+            __typename: 'OsMajorVersionEdge'
+        }
+    ]
+};
+
```
skateman, line 845: Not sure if the file is generated or not, but a newline is missing :thinking:

skateman:
The code looks good, but there's a backend issue we need to address before merging :disappointed: profiles with different `ref_ids` but the same name showed up in the UI :disappointed: maybe we need to rely on the profiles in the supported SSG yaml file as you originally implemented the backend part :shrug: 

@bastilian can you please take a look at this? The code is okay, just want to hold this off until the backend issue is resolved.

src/__fixtures__/benchmarks_rules.js
bastilian: ... I'd handy if we could generated these. ;)

skateman: In this case @marleystipich2 this needs an additional newline :wink:

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -170,6 +173,7 @@ const selector = formValueSelector('policyForm');
 export default compose(
   connect((state) => ({
     selectedBenchmarkId: selector(state, 'benchmark'),
```
bastilian, line 175: The `selectedBenchmarkId` could probably be removed. It seems to only occur here.

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -121,23 +124,21 @@ export const CreateSCAPPolicy = ({
...
-                  />
-                );
-              })}
+        <FormGroup label="Operating system" isRequired fieldId="osMajorVersion">
+          {osMajorVersions &&
+            osMajorVersions.map((os) => {
+              const osMajorVersion = os.node.osMajorVersion;
+              return (
+                <Tile
+                  key={os.node.osMajorVersion}
```
bastilian, line 133: This could also be `osMajorVersion`.

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
```diff
@@ -84,30 +85,32 @@ export const CreateSCAPPolicy = ({
     return <Spinner />;
   }
 
-  const benchmarks = data.latestBenchmarks;
-  let selectedBenchmark;
+  const osMajorVersions = data.osMajorVersions?.edges;
```
bastilian, line 88: Nice! Maybe we can (in the future) phase out the term "benchmark" completely (from the UI at least).

src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
marleystipich2: I tried removing this originally but `BenchmarkId` is still being used on the Review tab and it was getting it's value from this so when I removed this, the Review tab wasn't working.. Maybe that could be part of phasing out the term "benchmark"? Getting the Review tab to work correctly without `BenchmarkId`?

skateman: You can substitute the data coming from the new GQL request, the `benchmarkId` is no longer valid here anyway. Again, referring to the example query in the jira ticket :wink:

marleystipich2: So use benchmark `version` instead? And update the Review page to take the benchmark version instead of id?

skateman: I don't know for what this data is being used for, but you can query the ID of the benchmark in GQL under each profile if it's necessary.

bastilian:
> @bastilian can you please take a look at this? The code is okay, just want to hold this off until the backend issue is resolved.

@skateman From my side this is PR is approved, is the issue in the backend resolved?

skateman:
@bastilian backend is fine, but we need a UX adjustment for this before merging. Katie started working on it after yesterday's meeting.

bastilian:
Closing this as the switch to this endpoint has been incorporated into https://github.com/RedHatInsights/compliance-frontend/pull/1518

PR #1635, opened 2022-04-25
fix(Filters): Fix search filters not working

leSamo:
This bug is present in production.

Fixes [RHICOMPL-2897](https://issues.redhat.com/browse/RHICOMPL-2897).

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@patternfly/react-icons": "^4.53.16",
     "@patternfly/react-table": "^4.71.16",
     "@patternfly/react-tokens": "^4.54.16",
-    "@redhat-cloud-services/frontend-components": "^3.8.9",
+    "@redhat-cloud-services/frontend-components": "^3.6.2",
     "@redhat-cloud-services/frontend-components-notifications": "^3.2.5",
     "@redhat-cloud-services/frontend-components-pdf-generator": "^2.6.11",
     "@redhat-cloud-services/frontend-components-remediations": "^3.2.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "@patternfly/react-icons": "^4.53.16",
     "@patternfly/react-table": "^4.71.16",
     "@patternfly/react-tokens": "^4.54.16",
-    "@redhat-cloud-services/frontend-components": "^3.8.9",
+    "@redhat-cloud-services/frontend-components": "^3.6.2",
```
skateman, line 16: Is this really necessary? If yes, was this bug escalated to that repo?

skateman (changes requested):
Commit message should be something like:
```
fix(filters): RHICOMPL-2897 event handling in search inputs
```

package.json
leSamo: I don't know what the root cause is and it may take longer to investigate, so this is just a hotfix so we don't have the bug in production for too long.

skateman: I'm a bit concerned that these multiple repos somehow move together and downgrading just one might break other things :disappointed:

skateman: I pinged @Hyperkid123 about this problem, he or someone from his team will get back to us.

skateman: Okay, I was told that it's safe to revert and they will take care of this problem.

skateman:
@mkholjuraev you reported a duplicate of this issue, can you please take a look if this PR fixes it? :pray:

Hyperkid123:
@skateman @leSamo latest version of front-end components fixes the issue for some time now. Can you please upgrade instead of a downgrade?

PR #1643, opened 2022-04-28
fix(SystemsTable): RHICOMPL-2705 filter generation in OS filter

xmicha82:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -108,11 +108,11 @@ export const systemsOsFilterConfiguration = (policies) => [
 const toSystemsOsMinorFilterConfigurationItem =
   (osVersions) => (majorVersion) => ({
     label: `RHEL ${majorVersion}`,
-    value: majorVersion,
+    value: `${majorVersion}.99`, // .99 is here to distinguish state with only major version selected
     groupSelectable: true,
-    items: osVersions[majorVersion].map((minorVersion) => ({
+    items: osVersions[majorVersion]?.map((minorVersion) => ({
       label: `RHEL ${majorVersion}.${minorVersion}`,
-      value: minorVersion,
+      value: `${majorVersion}.${minorVersion}`,
     })),
   });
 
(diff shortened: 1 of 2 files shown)


src/constants.js
```diff
@@ -108,11 +108,11 @@ export const systemsOsFilterConfiguration = (policies) => [
 const toSystemsOsMinorFilterConfigurationItem =
   (osVersions) => (majorVersion) => ({
     label: `RHEL ${majorVersion}`,
-    value: majorVersion,
+    value: `${majorVersion}.99`, // .99 is here to distinguish state with only major version selected
```
vkrizan, line 111: This is a red flag for me. :smiley: 

Is there a better approach without using `.99` :snowflake:? Maybe having a tree view or tree-view-like behavior, when a major version is selected it would select all of its minor versions (child leafs).

skateman: :red_circle: :fire: :fire_engine: :rotating_light:

bastilian: @xmicha82 What is the difficulty with identivying the major version with a version like `7`, with the minor versions being something like `7.4`?

src/constants.test.js
```diff
@@ -0,0 +1,33 @@
...
+  [1, 2, 3, 4, 5, 6, 7, 8, 9],
+  [0, 1, 2, 3, 4, 5],
+];
+
+describe('filterString', () => {
+  const [{ filterString }] = systemsOsMinorFilterConfiguration(osMajorVersions);
+  console.log(typeof filterString);
+
+  it('should generate filter with all selected versions', () => {
+    const filter = filterString({ '6.99': { '6.9': true } }); // prettier-ignore
```
bastilian, line 20: `// prettier-ignore` why? :D

xmicha82: because prettier autoformats `'6.99'` in key to `6.99`, which breaks the test

bastilian: Why does it break the test?

xmicha82: ok, i would have sworn the tests weren't passing without strings in keys when i ran the tests last week... but they do pass
removing the `// prettier-ignore` then :salute:

src/constants.js
xmicha82: this is a really hacky way to deal with this issue, i admit, but it's the only way i could think of because i am unable to find the reason why currently selected major version appears as a minor version as well

bastilian: Won't the major version always be a single digit?

bastilian:
@xmicha82 What is that status of this?

src/constants.js
MichalSajdik: I rechecked this with xmicha82 and major version is like never single digit. In case of majir version 6, a value of 6.6 is returned, in case of major version of  7, a value of 7.7 is returned, we were unable to find exact place in the code where this happens

bastilian: I'd say `.0`, or even better `._`, is better than `.99`. Or anything else. Even if there will likely never be a "99" minor version, there will for sure not be a `._` minor version.

But I think, this is just working around the issue you mention @MichalSajdik you weren't able to find.

@xmicha82 let's look at this together next week.

xmicha82:
@bastilian last week we did some pair programming with Michal Sajdik to try to fix this and were not alble to find where this problem originates. we're gonna have a look at it again when Michal is at the office

bastilian:
@xmicha82 If you want we can also jump on a call any time and look into it and see what we can come up with.

jiridostal:
@bastilian anything that needs to be done here?

bastilian:
Closing this as compliance should use the Inventory OS helpers to build it's filters.

PR #1663, opened 2022-05-24
fix(ReportDetails): RHICOMPL-2743 conditionally hide unsupported legend

xmicha82:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/ReportDetails/Components/hooks/useLegendData.js
+++ b/src/SmartComponents/ReportDetails/Components/hooks/useLegendData.js
@@ -31,30 +31,34 @@ const useLegendData = (donutValues, profile) => {
       )} non-compliant`,
       symbol: { fill: blue200.value },
     },
-    {
-      name: `${donutValues[2].y} ${pluralize(
-        donutValues[2].y,
-        'system'
-      )} not supported`,
-      symbol: { fill: chart_color_gold_300.value },
-      popover: {
-        title: 'Unsupported SSG versions',
-        content: (
-          <>
-            <Text variant="p">
-              These systems are running unsupported versions of the SCAP
-              Security Guide (SSG) for the version of RHEL installed on them.
-              Assessment of rules failed/passed on these systems is a best-guess
-              effort and may not be accurate.
-            </Text>
-            <Text variant="p">
-              The policy&apos;s compliance score excludes these systems.
-            </Text>
-          </>
-        ),
-        footer: <SupportedSSGVersionsLink />,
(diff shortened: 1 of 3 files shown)


src/SmartComponents/ReportDetails/Components/hooks/useLegendData.js
```diff
@@ -16,6 +16,8 @@ const useLegendData = (donutValues, profile) => {
   const notReportingHostCount =
     totalHostCount - unsupportedHostCount - testResultHostCount;
 
+  console.log(profile);
```
MichalSajdik, line 19: I believe we do not wanna have console.log in master

src/SmartComponents/ReportDetails/Components/hooks/useLegendData.js
```diff
@@ -31,30 +33,34 @@ const useLegendData = (donutValues, profile) => {
...
-            </Text>
-            <Text variant="p">
-              The policy&apos;s compliance score excludes these systems.
-            </Text>
-          </>
-        ),
-        footer: <SupportedSSGVersionsLink />,
-      },
-    },
+    ...(unsupportedHostCount > 0
```
MichalSajdik, line 36: It seems like ternary operator is unnecessary.
Maybe this condition can be rewritten with && as is shown here: https://stackoverflow.com/a/40560953
I would imagine something like (unsupportedHostCount > 0) && {<code that you want>}

xmicha82: i used the ternary for consistence, it is used to conditionally show the *not reporting* systems as well. i think it's a better idea to have it the same in both places

bastilian: @MichalSajdik I believe here it might also be needed to have it `(unsupportedHostCount > 0) && {code that you want} || []`, because it would try to spread `false` and fail because it's not an "iteratable". In which case we should prefer the common ternary operator, like it is now, ... for now. Personally, I find the `.. && .. || ..` still harder to comprehend.

bastilian (approved):
Nice work @xmicha82! Thank you!

PR #1699, opened 2022-06-29
fix(clowder): RHCLOUD-19846 Add config files into FE container

gburges:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [x] File Management
- [ ] Memory Management


Dockerfile
```diff
@@ -12,7 +12,7 @@ ENV WORKDIR /compliance/
 RUN mkdir -p $WORKDIR
 WORKDIR $WORKDIR
 
-COPY package*.json $WORKDIR
+COPY /config/ $WORKDIR
```
bastilian, line 15: `COPY . $WORKDIR` Should work as well and if the frontend should be built it probably will need it.

gburges: updated, ty

bastilian (approved):
LGTM! Thank you for taking care of this @gburges!

Dockerfile
```diff
@@ -12,7 +12,7 @@ ENV WORKDIR /compliance/
 RUN mkdir -p $WORKDIR
 WORKDIR $WORKDIR
 
-COPY package*.json $WORKDIR
+COPY . $WORKDIR
```
vkrizan, line 15: Doesn't this add the whole project to the container? Isn't this Dockerfile for local use (considering `start:proxy` in the `CMD`, and the workdir is mounted in?

bastilian: Yes, and that is on purpose. In local development the volume definition in the `docker-compose.yml` will override this.

gburges: it does, the current problem is that the config dir is not copied into the frontend container. This change make sure we're not missing any files ever

PR #1708, opened 2022-07-07
feat(ReportPDF): RHICOMPL-3004 - Use topFailedRules GQL query

bastilian:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


bastilian:
@skateman We can not do it that way, because the `GET_SYSTEMS` query could be called multiple times in order to get all hosts for the export and the rules will only be needed once.

src/SmartComponents/ReportDownload/constants.js
```diff
@@ -98,3 +76,24 @@ export const GET_PROFILE = gql`
...
 `;
+
+export const GET_RULES = gql`
+  query getProfiles($filter: String!, $policyId: ID!) {
+    profiles(search: $filter) {
+      totalCount
+      edges {
+        node {
+          topFailedRules(policyId: $policyId) {
+            id
```
skateman, line 87: Do we really need the ID, because it is not really useful as it comes out from an aggregated function :thinking: IMO it should be dropped if it can be.

bastilian: i think we can remove that.

skateman (approved):
The Seal of Approval
Tested with 9k hosts, works well.

PR #1714, opened 2022-07-21
fix(ReportsTable):fixed the n/a in the table

Fewwy:
It was requested to show the "number%" instead of N/A even in cases when there is a 0 compliance score.
This function was utilized both by table and Donut Chart.

image
image


LightOfHeaven1994:
according to the screenshot it looks fine :)

bastilian:
@Fewwy Thank you! This looks good to me. It would be good to add a test for the cases where the score is 0.

The best place would probably be the `Cells` tests of the SystemsTable[0]. Just on test with a fitting testSystem should be enough. 

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/Cells.test.js#L61

bastilian (approved):
LGTM! Thank you @Fewwy!

PR #1715, opened 2022-07-22
fix(ReportPDF): Added a query parameter for the rulesFailed

Fewwy:
Added a rulesFailed to the GraphQL request. That was the last part missing, everything else works fine.
Now PDF report contains the rules failed number.
(Don't look at the N/A - this patch is not merged yet)
image


skateman (approved):
The Seal of Approval

PR #1725, opened 2022-08-03
feat(ReportsTable): new compliance filters

Fewwy:
Two new Compliance filters for the Reports table

image

image

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -170,10 +170,18 @@ export const COMPLIANT_SYSTEMS_FILTER_CONFIGURATION = [
   {
     type: conditionalFilterType.checkbox,
     label: 'Compliance',
-    filterString: (value) => `compliant = ${value}`,
+    filterString: (value) => `${value}`,
     items: [
-      { label: 'Compliant', value: 'true' },
-      { label: 'Non-compliant', value: 'false' },
+      {
+        label: 'Compliant',
+        value: 'compliant = true AND supported_ssg = true',
+      },
+      {
+        label: 'Non-compliant',
+        value: 'compliant = false AND supported_ssg = true',
+      },
+      { label: 'Not supported', value: 'supported_ssg = false' },
+      { label: 'Never reported', value: 'reported = false' },
     ],
   },
   {
--- a/src/SmartComponents/SystemsTable/__snapshots__/SystemsTable.test.js.snap
+++ b/src/SmartComponents/SystemsTable/__snapshots__/SystemsTable.test.js.snap
@@ -858,11 +858,19 @@ exports[`SystemsTable returns with compliantFilter 1`] = `
                 "items": Array [
                   Object {
                     "label": "Compliant",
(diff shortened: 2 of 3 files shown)


Victoremepunto:
How did the commitlint task passed for this commit message ? :thinking: - there's no mention to the JIRA task

skateman:
> How did the commitlint task passed for this commit message ? thinking - there's no mention to the JIRA task

we don't require a JIRA ticket in the commit message, but it would be nice to have it here :see_no_evil:

xmicha82:
hey, is this correct behaviour?
image

LightOfHeaven1994:
LGTM, thanks @Fewwy !

LightOfHeaven1994 (approved):

LightOfHeaven1994 (changes requested):
Screenshot from 2022-08-04 11-31-51
I was able to reproduce bug that Martin attached

Fewwy:
@LightOfHeaven1994 Ok I'm looking in to the why the bug is happening

skateman:
It might be a backend issue, I'm investigating.

Fewwy:
@skateman Any updates?

skateman:
I'm deep in debugging a memory leak and our SLO, so this has a lower prio atm.

src/constants.js
```diff
@@ -170,10 +170,12 @@ export const COMPLIANT_SYSTEMS_FILTER_CONFIGURATION = [
   {
     type: conditionalFilterType.checkbox,
     label: 'Compliance',
-    filterString: (value) => `compliant = ${value}`,
+    filterString: (value) => `${value}`,
     items: [
-      { label: 'Compliant', value: 'true' },
-      { label: 'Non-compliant', value: 'false' },
+      { label: 'Compliant', value: 'compliant = true' },
+      { label: 'Non-compliant', value: 'compliant = false' },
```
skateman, line 176: I think what @LightOfHeaven1994 is describing can be fixed by these changes:
```suggestion
      { label: 'Compliant', value: 'compliant = true AND supported_ssg = true' },
      { label: 'Non-compliant', value: 'compliant = false AND supported_ssg = true' },
```

skateman (commented):
Not sure if UX would approve it, but I think we should restrict these filters a little bit. Ideally by allowing only one to be selected from these. Is that something doable? We're very limited here on the backend :disappointed:

src/constants.js
LightOfHeaven1994: @Fewwy can you try David's suggestion? also PR needs to be rebased as we use new tags gql query

skateman: @LightOfHeaven1994 he's on PTO, I rebased and adjusted the PR, can you please take a look?

LightOfHeaven1994: @skateman Works great now! thanks for the fix <3

LightOfHeaven1994 (approved):

PR #1738, opened 2022-08-22
fix(EditPolicy): RHICOMPL-1783 - Correct behaviour of alert for new rules

bastilian:
This corrects the behaviour of the alert about new rules at the bottom of the "Systems" tab.
it now correctly identifies if there is a new set of rules to configure if a systems with a minor version is selected that had not yet been covered by a rule set.

*How to test:*

1. Find a profile with multiple supported RHEL minor versions
2. Select systems only of one minor version and create the policy
3. Edit the policy.
4. The modal on the Systems tab should not show the alert when first loaded.
5. Select systems of a different minor version
6. The Alert should appear and allow to switch to the "Rules" tab via the link
7. (When removing all systems of a specific minor version, the alert should disappear again)
8. After saving and reopening there should again not be any alert on the Systems tab

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/hooks/index.js
+++ b/src/SmartComponents/EditPolicy/hooks/index.js
@@ -1,0 +1,3 @@
+import { useState } from 'react';
+
+export const useNewRulesAlertState = (...args) => useState(...args);
--- a/src/PresentationalComponents/TabbedRules/TabbedRules.js
+++ b/src/PresentationalComponents/TabbedRules/TabbedRules.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import propTypes from 'prop-types';
 import { Tab, Badge } from '@patternfly/react-core';
 import { RoutedTabs } from 'PresentationalComponents';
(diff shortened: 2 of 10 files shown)


src/SmartComponents/EditPolicy/hooks/index.js
```diff
@@ -0,0 +1,3 @@
+import { useState } from 'react';
+
+export const useNewRulesAlertState = (...args) => useState(...args);
```
bastilian, line 3: This is really just to make the form easier testable, since the entire Inventory table is mocked and it would be trickier to simulate the selecting of systems.

adonispuente (commented):
The PR seems to be working as intended, but I have a question about a particular scenario.
 I wasnt sure what the desired outcome is:

- When editing a policy and adding a system of a different RHEL version : If I remove all the rules in the rules tab > go back to systems tab to deselect & reselect the RHEL system > in the rules tab, should rules be present again? At the moment the tab will is empty. The only way to get the rules to appear again is to close the modal or save and re-edit the modal.

johnsonm325:
> The PR seems to be working as intended, but I have a question about a particular scenario. I wasnt sure what the desired outcome is:
> 
> * When editing a policy and adding a system of a different RHEL version : If I remove all the rules in the rules tab > go back to systems tab to deselect & reselect the RHEL system > in the rules tab, should rules be present again? At the moment the tab will is empty. The only way to get the rules to appear again is to close the modal or save and re-edit the modal.

This is a really good point, but not sure it needs to be fixed in this PR as I see it was present before. But, I think it should be fixed because if you add a new minor version system, it populates in the Rules tab with the new set of rules.

@bastilian Let us know if you think this needs some discussion. I would think that if adding a new minor version updates the rules, then removing systems of an existing minor version should also update the rules. So if I have all 8.6 version selected, go to edit the policy and then remove all of my 8.6 systems, the Rules tab should show no rules under the 8.6 tab.

bastilian:
@adonispuente @johnsonm325 There are two situations here. The first is when a policy has already been saved with a set of rules, they will and should stay even if there is no system assigned anymore. In case a system is added via the "edit policy" and then removed, but in between there was no saving, the rules should (if I remember right) disappear again.

PR #1744, opened 2022-08-25
perf(GQL): RHICOMPL-3248 do not query rules in system queries

skateman:
I am not 100% sure about this, but I haven't seen any issues after introducing it. @bastilian can you please verify and let me know if I'm wrong. This probably affects @xmicha82's work as it tampers with the queries, but in case this is off, his work will partially reduce the performance impact at least at some places where this query is being used.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/PolicyDetails/__snapshots__/PolicySystemsTab.test.js.snap
+++ b/src/SmartComponents/PolicyDetails/__snapshots__/PolicySystemsTab.test.js.snap
@@ -450,76 +450,6 @@ exports[`PolicySystemsTab expect to render with no systems table 1`] = `
                                         },
                                         "selectionSet": undefined,
                                       },
-                                      Object {
-                                        "alias": undefined,
-                                        "arguments": Array [],
-                                        "directives": Array [],
-                                        "kind": "Field",
-                                        "name": Object {
-                                          "kind": "Name",
-                                          "value": "rules",
-                                        },
-                                        "selectionSet": Object {
-                                          "kind": "SelectionSet",
-                                          "selections": Array [
-                                            Object {
-                                              "alias": undefined,
-                                              "arguments": Array [],
-                                              "directives": Array [],
-                                              "kind": "Field",
-                                              "name": Object {
-                                                "kind": "Name",
-                                                "value": "refId",
-                                              },
-                                              "selectionSet": undefined,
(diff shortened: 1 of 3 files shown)


xmicha82 (commented):
this actually causes incorrect data to appear in the Reports Systems Table
correct:
ReportsSystemTable_correct
after the change:
ReportsSystemTable_query_rmd

skateman:
@xmicha82 as far as I understand, the `rules` object here is being partitioned to an array of failing and passing rules which then is somehow being used to count the failed rules and display it on the frontend. However, we have the `rulesFailed` attribute on `testResultProfiles` so we can bypass this. However, the code is very hard to follow :disappointed:

bastilian:
This is not possible. We need to pass the ruleIds to remediations, if I'm not wrong. There is an issue open on Jira to change the behaviour and query the rules when the user clicks the button only, but for now we can not remove the rules from the query.

skateman:
@bastilian yes, I understand that in certain pages where you have rules for a single system, you might need this information. But you're querying this for pages where you show systems and the number of failed rules only. If you're asking for all the rules available under 50 systems just to count the number of failed rules. Can you please help me clarify if there's any page where we actually require the rules (not just their count) on a page where we load more than a single system?

bastilian:
@skateman When you are on a Reports details page and select a few systems and click on the Remediations button the ruleIDs are required to be passed to the wizard.

vkrizan:
> @skateman When you are on a Reports details page and select a few systems and click on the Remediations button the ruleIDs are required to be passed to the wizard.

Can those rule ids be queried asynchronously when the button / remediation wizard is opened. That would alleviate a lot of pressure from backend.

bastilian:
@vkrizan Yes, but we can not just remove them from the query, without changing that first. (See https://issues.redhat.com/browse/RHICOMPL-3071)

skateman:
Already drafted up a proposal to fix this, with an optimized backend in that ticket :zap:

PR #1718, opened 2022-07-28
refactor(SystemsDetails): WIP

bastilian:
## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1748, opened 2022-09-01
Optimizations

mkholjuraev:
Resolves: [RHICOMPL-3201](https://issues.redhat.com/browse/RHICOMPL-3201)

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/PoliciesTable/Columns.js
+++ b/src/PresentationalComponents/PoliciesTable/Columns.js
@@ -6,10 +6,10 @@ import { LinkWithPermission as Link } from 'PresentationalComponents';
 import { GreySmallText, SystemsCountWarning } from 'PresentationalComponents';
 import { renderComponent } from 'Utilities/helpers';
 
-const PolicyNameCell = ({ id, policy }) => (
+const PolicyNameCell = ({ id, policy, policyType }) => (
   <TextContent>
     <Link to={'/scappolicies/' + id}>{policy.name}</Link>
-    <GreySmallText>{policy.name}</GreySmallText>
+    <GreySmallText>{policyType}</GreySmallText>
   </TextContent>
 );
 
@@ -31,7 +31,7 @@ export const Name = {
 
 const PolicyType = {
   title: 'Policy Type',
-  renderExport: (policy) => policy.name,
+  renderExport: (policy) => policy.policyType,
 };
 
 const osString = (policy) => `RHEL ${policy.osMajorVersion}`;
--- a/src/PresentationalComponents/PoliciesTable/Columns.js
+++ b/src/PresentationalComponents/PoliciesTable/Columns.js
@@ -6,10 +6,10 @@ import { LinkWithPermission as Link } from 'PresentationalComponents';
 import { GreySmallText, SystemsCountWarning } from 'PresentationalComponents';
 import { renderComponent } from 'Utilities/helpers';
 
-const PolicyNameCell = ({ id, policy, policyType }) => (
+const PolicyNameCell = ({ id, policy }) => (
   <TextContent>
     <Link to={'/scappolicies/' + id}>{policy.name}</Link>
-    <GreySmallText>{policyType}</GreySmallText>
+    <GreySmallText>{policy.name}</GreySmallText>
   </TextContent>
 );
 
@@ -31,7 +31,7 @@ export const Name = {
 
 const PolicyType = {
   title: 'Policy Type',
-  renderExport: (policy) => policy.policyType,
+  renderExport: (policy) => policy.name,
 };
 
 const osString = (policy) => `RHEL ${policy.osMajorVersion}`;
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -31,7 +31,7 @@ const QUERY = gql`
           osMajorVersion
           policy {
             id
-            name            
+            name
           }
           businessObjective {
             id
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -30,6 +30,7 @@ const QUERY = gql`
           complianceThreshold
           totalHostCount
           osMajorVersion
+          policyType
           policy {
             id
             name
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -24,6 +24,7 @@ const QUERY = gql`
       edges {
         node {
           id
+          name
           description
           refId
           complianceThreshold
--- a/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
+++ b/src/SmartComponents/CompliancePolicies/CompliancePolicies.js
@@ -24,19 +24,14 @@ const QUERY = gql`
       edges {
         node {
           id
-          name
           description
           refId
           complianceThreshold
           totalHostCount
           osMajorVersion
-          policyType
           policy {
             id
-            name
-          }
-          hosts {
-            id
+            name            
           }
           businessObjective {
             id
--- a/src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
@@ -38,15 +38,9 @@ const PROFILES_QUERY = gql`
           refId
           osMinorVersion
           osMajorVersion
-          policy {
-            id
-          }
-          policyType
           benchmark {
             id
-            refId
             latestSupportedOsMinorVersions
-            osMajorVersion
           }
           rules {
             id
--- a/src/SmartComponents/CreatePolicy/ReviewCreatedPolicy.js
+++ b/src/SmartComponents/CreatePolicy/ReviewCreatedPolicy.js
@@ -10,45 +10,17 @@ import {
 } from '@patternfly/react-core';
 import { formValueSelector } from 'redux-form';
 import { connect } from 'react-redux';
-import gql from 'graphql-tag';
 import propTypes from 'prop-types';
-import Spinner from '@redhat-cloud-services/frontend-components/Spinner';
-import { useQuery } from '@apollo/client';
-
-const REVIEW = gql`
-  query review($benchmarkId: String!) {
-    benchmark(id: $benchmarkId) {
-      id
-      osMajorVersion
-    }
-  }
-`;
 
 const ReviewCreatedPolicy = ({
-  benchmarkId,
+
   name,
   businessObjective,
   complianceThreshold,
   parentProfileName,
   osMinorVersionCounts,
-}) => {
-  const { data, error, loading } = useQuery(REVIEW, {
-    variables: { benchmarkId },
-  });
-
-  if (error) {
-    return error;
-  }
-
-  if (loading) {
-    return <Spinner />;
-  }
-
-  const {
-    benchmark: { osMajorVersion },
-  } = data;
-
-  return (
+  osMajorVersion
+}) => (
     <TextContent>
       <Text component={TextVariants.h1}>Review</Text>
       <Text>Review your SCAP policy before finishing.</Text>
@@ -99,7 +71,6 @@ const ReviewCreatedPolicy = ({
       </TextList>
     </TextContent>
   );
-};
 
 ReviewCreatedPolicy.propTypes = {
   benchmarkId: propTypes.string,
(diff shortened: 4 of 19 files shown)


src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js
```diff
@@ -38,15 +38,9 @@ const PROFILES_QUERY = gql`
           refId
           osMinorVersion
           osMajorVersion
-          policy {
```
mkholjuraev, line 41: I did not see any place where this field is used.

src/SmartComponents/CreatePolicy/ReviewCreatedPolicy.js
```diff
@@ -10,45 +10,17 @@ import {
 } from '@patternfly/react-core';
 import { formValueSelector } from 'redux-form';
 import { connect } from 'react-redux';
-import gql from 'graphql-tag';
 import propTypes from 'prop-types';
-import Spinner from '@redhat-cloud-services/frontend-components/Spinner';
-import { useQuery } from '@apollo/client';
-
-const REVIEW = gql`
```
mkholjuraev, line 18: I believe this query can be removed thanks to the osMajorVerions prop passed from the wizard steps.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -21,43 +21,6 @@ import {
 import { sortingByProp } from 'Utilities/helpers';
 import * as Columns from '@/PresentationalComponents/RulesTable/Columns';
 
-const PROFILES_QUERY = gql`
```
mkholjuraev, line 24: This might be completely wrong, just an idea. Can we use the data passed (`policy.policy.profiles`)  from parent EditPolicy instead of fetching it again?  Also, I did not see much usage of these data fields.

src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -86,16 +85,6 @@ export const QUERY = gql`
         id
         osMinorVersion
       }
-      rules {
```
mkholjuraev, line 89: Here also, I did not find the usage of `rules`

src/SmartComponents/ReportDetails/ReportDetailsWithNotReportedSystems.js
```diff
@@ -75,6 +70,7 @@ export const QUERY = gql`
   }
 `;
 
+//what is the difference between ReportDetails and ReportDetailsWithNotreportedSystems?
```
mkholjuraev, line 73: Do not pay attention to this :). I understand now.

src/SmartComponents/ReportDownload/constants.js
```diff
@@ -49,6 +49,7 @@ export const GET_SYSTEMS = gql`
   }
 `;
 
+//why inner policy and benchmark query? NOTE: benchmark verions is usually used with os info.
```
mkholjuraev, line 52: This is more of a question.

mkholjuraev:
@skateman can you also please have a look and provide your feedback?

src/SmartComponents/ReportDetails/ReportDetails.js
```diff
@@ -151,7 +146,7 @@ export const ReportDetails = ({ route }) => {
           <Grid hasGutter>
             <GridItem sm={9} md={9} lg={9} xl={9}>
               <PageHeaderTitle title={pageTitle} />
-              <SubPageTitle>{profile.policyType}</SubPageTitle>
+              <SubPageTitle>{policyName}</SubPageTitle>
```
skateman, line 149: I'm not 100% sure about it, `profile.name˙ should be equal to the `policyType` afaik.

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -52,6 +54,9 @@ const PROFILES_QUERY = gql`
             remediationAvailable
             identifier
           }
+=======
+          osMajorVersion  
+>>>>>>> 7be077f (wip)
```
skateman: :red_circle: :fire:

mkholjuraev: fixed in the next commit :). I will remove this commit as well.

mkholjuraev:
@bastilian can you have a quick look whenever you are free?

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -29,10 +29,18 @@ const QUERY = gql`
 
 export const SystemDetails = ({ route }) => {
   const { inventoryId } = useParams();
-  const { data, error, loading } = useQuery(QUERY, {
-    variables: { inventoryId },
-  });
-  const systemName = data?.system?.name;
+  const { state: locationState } = useLocation();
```
bastilian, line 32: I know that in some occasions we do use the location state, but it doesn't feel proper.

mkholjuraev: In my opinion, according to [docs](https://reactrouter.com/en/v6.3.0/getting-started/concepts#locations) location state is designed for such cases. I suggest going in this way, and getting QA feedback,

bastilian: What is the case here? Also, where and how is the state set? What if the state is not set properly and how are system details shown in that case? I'm not a big fan of this, even if I used/introduced it in the past. :)

bastilian: Big fan of shuffling data from one view/route to the other via the state, because if one goes directly to such a page it might cause issues.

src/SmartComponents/ReportDetails/ReportDetails.js
mkholjuraev: During our  discussion in the office, you and @bastilian thought that they are the same. We tried this in the UI as well and did not see any problems.

src/SmartComponents/CreatePolicy/__snapshots__/ReviewCreatedPolicy.test.js.snap
```diff
@@ -176,7 +176,6 @@ exports[`ReviewCreatedPolicy expect to render without error 1`] = `
                           }
                         >
                           RHEL 
-                          7
```
skateman, line 179: This seems a bit alarming, `RHEL.7`? :thinking:

bastilian: This is likely missing an update in the tests, to pass in a major version as a prop, since it changed where that comes from.

src/SmartComponents/PolicyDetails/PolicyDetails.js
bastilian: We will need those for now, till the changes for requesting rules async are done.

bastilian: Sorry, you are right they aren't used anymore.

src/SmartComponents/SystemsTable/Cells.js
```diff
@@ -21,12 +21,15 @@ import {
   NEVER,
 } from 'Utilities/ruleHelpers';
 
-const SystemLink = ({ id, children }) => (
-  <Link to={{ pathname: `/systems/${id}` }}>{children}</Link>
+const SystemLink = ({ id, name, children }) => (
+  <Link to={{ pathname: `/systems/${id}`, state: { systemName: name } }}>
+    {children} + {name}
```
bastilian, line 26: This results in very strange output. :D 

Screenshot 2022-09-20 at 10 46 15

mkholjuraev: Ups, I will fix it

src/SmartComponents/CreatePolicy/__snapshots__/ReviewCreatedPolicy.test.js.snap
mkholjuraev: Fixed in tests, also osMajorVersion is changed to to 'isRequired' in the proptypes.

src/SmartComponents/SystemDetails/SystemDetails.js
mkholjuraev: I believe that we are just sending the system id to the routed component through the location state. If there is not such a state/not properly set, then the name of the system is fetched from the API. I would say this is safe.

bastilian: It is still not obvious where the system name in the location state is set or how or when. Another way we could do this is to allow the Compliance Details to call back with the system name when it loaded the details. There is a sketch of this here: https://github.com/RedHatInsights/compliance-frontend/pull/1718/files#diff-d002195aa23b5c7ab970c83c31119ce7aad7e98d40149fd02adb7ed9a0dea329R17

We could also restructure the components to have the full query here at the top and pass down the details, but we'd need to wrap/change the exported federated module to query it there as well.

mkholjuraev: Okay, I will revert the commit for this specific bit of code for the location state. I see that you have already some idea that is more promising.

mkholjuraev: Done!

src/SmartComponents/ReportDetails/ReportDetails.js
bastilian: As far as I know, the `policyType` is the name of the profile the policy was created with, and the `name` is what the user entered when creating/editing the policy.

bastilian: So, they are not equal. Many Policies in the test data might make it appear that there is no difference, but that is only, because many have not edited the name.

src/SmartComponents/CompliancePolicies/CompliancePolicies.js
```diff
@@ -24,20 +24,15 @@ const QUERY = gql`
       edges {
         node {
           id
-          name
```
bastilian, line 27: I don't think these can be removed, `name` and `policyType`.

bastilian:
@mkholjuraev The only thing I found during testing this is that the Policies table has broken sorting and filtering for the name, but reverting the removal of the `policyType` and `name` should restore that.

Other than that all works great!

mkholjuraev:
@bastilian thank you for the review. I will revert that comment in that case.

bastilian:
@mkholjuraev OH! I just noticed. What we could do is to use the `policy.policy.name`[0] instead of the `policyType` on the Compliance Policies page.

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CompliancePolicies/CompliancePolicies.js#L36

mkholjuraev:
@bastilian I returned the 'name' into the query as I believe it would not affect the performance much. Using `policy.policy.name` seemed kind of ugly :)

bastilian:
@mkholjuraev Ugly or not, we still need to get the "policy type" from somewhere. ;) Either we add the `policyType` back into the query, or we use `policy.policy.name`.

src/PresentationalComponents/PoliciesTable/Columns.js
```diff
@@ -6,10 +6,10 @@ import { LinkWithPermission as Link } from 'PresentationalComponents';
 import { GreySmallText, SystemsCountWarning } from 'PresentationalComponents';
 import { renderComponent } from 'Utilities/helpers';
 
-const PolicyNameCell = ({ id, policy, policyType }) => (
+const PolicyNameCell = ({ id, policy }) => (
   <TextContent>
     <Link to={'/scappolicies/' + id}>{policy.name}</Link>
-    <GreySmallText>{policyType}</GreySmallText>
+    <GreySmallText>{policy.name}</GreySmallText>
```
bastilian, line 12: This needs to be the `policyType` again or `policy.policy.name`

src/PresentationalComponents/PoliciesTable/Columns.js
```diff
@@ -31,7 +31,7 @@ export const Name = {
 
 const PolicyType = {
   title: 'Policy Type',
-  renderExport: (policy) => policy.policyType,
+  renderExport: (policy) => policy.name,
```
bastilian, line 34: Same as on line 12.

mkholjuraev:
Returned `policyType` (in the last [commit](https://github.com/RedHatInsights/compliance-frontend/pull/1748/commits/b2fb91a9386b1c277aa63892dd872b1a22819008)). Hopefully, everything is okay now :)

bastilian (approved):
LGTM! Thank you @mkholjuraev!

bastilian:
@mkholjuraev When merging, mind adding the Jira also to the commit? :)

bastilian:
@mkholjuraev Yeah, You can change the commit message when doing a "Squash & Merge".

PR #1754, opened 2022-09-02
perf(GQL): drop the complianceThreshold from CreateSCAPPolicy

skateman:
As we discussed with @bastilian @mkholjuraev and @Fewwy that this is not necessary here anymore.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
+++ b/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js
@@ -33,7 +33,6 @@ const SUPPORTED_PROFILES = gql`
             name
             refId
             description
-            complianceThreshold
             supportedOsVersions
             benchmark {
               id
--- a/src/SmartComponents/CreatePolicy/EditPolicyDetails.js
+++ b/src/SmartComponents/CreatePolicy/EditPolicyDetails.js
@@ -83,7 +83,7 @@ export const EditPolicyDetails = ({ change, policy, refId }) => {
             defaultValue={policy.businessObjective}
           />
         </FormGroup>
-        <ProfileThresholdField previousThreshold={policy.complianceThreshold} />
+        <ProfileThresholdField previousThreshold={100} />
       </Form>
     </React.Fragment>
   );


mkholjuraev:
@skateman there are other occurrences of `complianceThreshold` in the codebase (EditPolicy, PolicyDetails, ReportDetails ...). Should we also drop them? Ups, Github has a small crash.

skateman:
> @skateman there are other occurrences of `complianceThreshold` in the codebase (EditPolicy, PolicyDetails, ReportDetails ...). Should we also drop them? Ups, Github has a small crash. I hit 'Close the pr' button suddenly sorry :)

I am not sure about those, but this one is 100% not needed anymore and drops an unnecessary join with policies.

mkholjuraev:
It looks like with this PR, there will be a change in the UI. I am not sure if this is expected. When I open 'Create a new policy' wizard and get to the second step. I see **Threshold has to be a number between 0 and 100
Threshold values can have a maximum of one decimal place** by default. On the actual UI, this is not [the case.](https://console.stage.redhat.com/beta/insights/compliance/scappolicies/new)

skateman:
> It looks like with this PR, there will be a change in the UI. I am not sure if this is expected. When I open 'Create a new policy' wizard and get to the second step. I see **Threshold has to be a number between 0 and 100 Threshold values can have a maximum of one decimal place** by default. On the actual UI, this is not [the case.](https://console.stage.redhat.com/beta/insights/compliance/scappolicies/new)

@bastilian can you give us a hand here? How the threshold here can be used? :thinking:

bastilian:
@skateman Policies could have a predefined/default compliance threshold, we should triple check on that. Because maybe this isn't important and we could replace this with a static default of 100%. The reason why it fails only on the second step is, because in the first step we put the whole profile into the redux store[0] as the defaults for (redux-)form. 


[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js#L129

skateman:
> @skateman Policies could have a predefined/default compliance threshold, we should triple check on that. Because maybe this isn't important and we could replace this with a static default of 100%. The reason why it fails only on the second step is, because in the first step we put the whole profile into the redux store[0] as the defaults for (redux-)form.
> 
> [0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js#L129

The profiles don't have this field, so the predefined value makes no sense here. What should I change and where to make this work?

bastilian:
@skateman We will need to add a static default here then: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/EditPolicyDetails.js#L86 just using "100" should do the trick.

bastilian (approved):
LGTM! Thank you @skateman!

PR #1772, opened 2022-09-30
fix(editPolicyInline): fixing a bug with inline edits

Fewwy:
Found a bug that was breaking the save function. 
PLUS
A small improvement
I added a logic that removes unsaved changes if the user closes inline edit by pressing on the "pencil" icon-button.


bastilian (approved):
Saving works now! Thank you for fixing this @Fewwy!

PR #1788, opened 2022-10-19
feat(RHICOMPL-3212): Additional Rules table page

adonispuente:
This implements https://issues.redhat.com/browse/RHICOMPL-3212 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -23,9 +23,12 @@ const RulesTable = ({
   hidePassed = false,
   options,
   activeFilters,
+  removeColumnDropdown,
   ...rulesTableProps
 }) => {
-  const manageColumnsEnabled = useFeature('manageColumns');
+  const manageColumnsEnabled = !removeColumnDropdown
```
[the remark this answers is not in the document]
adonispuente: oh nice, ill send that up now

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -0,0 +1,98 @@
...
+              {
+                profile: {
+                  refId: data.profile.refId,
+                  name: data.profile.name,
+                },
+                rules: data.profile.rules,
+              },
+            ]}
+            manageColumns={true}
+            options={{ pagination: false }}
```
bastilian, line 81: ... i'm sorry, but i only realised this now. We can do `options={{ pagination: false, manageColumns: false }}` and avoid adding a custom prop to the rules table entirely.

xmicha82 (changes requested):
could you please tidy up the structure so the components are in the correct place?
please move `ViewAffectedLink.js` to `PresentationalComponents/ViewAffectedLink/ViewAffectedLink.js` and `PolicyRules` to `PresentationalComponents/PolicyRules`

xmicha82:
also, in the mocks it says the rules view should open in a new browser window. [This](https://stackoverflow.com/questions/47574490/open-a-component-in-new-window-on-a-click-in-react) can be helpful i believe

adonispuente:
> 
Good catch, I didnt see that! I would prefer to just add the target="_blank" attribute to the Link component

adonispuente:
> another thing: i would prefer to have some kind of loading placeholder while the rules load. This shows up while the rules are loading (which is not insignificant time, since there are a lot of rules) Screenshot from 2022-10-20 15-17-11 and there is no sign that anything is loading.

Good point, I dont see a mock, but I could replace the words with a spinner icon as it loads, and have the table just not show up until theres data to use

src/SmartComponents/PolicyRules/PolicyRules.js
adonispuente: When implementing I see that you need to have  ` manageColumns: true,` inside the main rulesTable to still display the 3 dots dropdown. Then when adding `options={{ pagination: false, manageColumns: false }}`, it wont 'override' it

src/PresentationalComponents/PolicyRules/PolicyRules.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
```
bastilian, line 1: This qualifies in my opinion as a "SmartComponent".

src/SmartComponents/PolicyRules/PolicyRules.js
bastilian: You'll need to remove what you added to the RulesTable for manageColumns here: https://github.com/RedHatInsights/compliance-frontend/pull/1788/files#diff-4216fa7cd062592b6adaf89dd868e009c72ef7d3359e5b626dd4e05cf1b6a8fdR89

The options are passed down as options for the TableTools and enable the manage columns.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
```diff
@@ -118,7 +125,16 @@ const ProfileTabContent = ({
...
             <ProfileSystemCount count={systemCount} />
           </Text>
-          <SSGVersionText {...{ profile, newOsMinorVersion }} />
+          <Flex>
+            <FlexItem>
+              <SSGVersionText {...{ profile, newOsMinorVersion }} />
+            </FlexItem>
+            {hash !== '#rules' && (
+              <FlexItem align={{ default: 'alignRight' }}>
+                <ViewAffectedLink message={VIEW_POLICY_RULES} policy={policy} />
```
bastilian, line 134: The text that `VIEW_POLICY_RULES` returns won't probably be used anywhere else but here. I'd just put the string straight in here, or even directly into the ViewAffectedLink, if we are only using the component with the same text.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
```diff
@@ -118,7 +125,16 @@ const ProfileTabContent = ({
             </span>
             <ProfileSystemCount count={systemCount} />
           </Text>
-          <SSGVersionText {...{ profile, newOsMinorVersion }} />
+          <Flex>
+            <FlexItem>
+              <SSGVersionText {...{ profile, newOsMinorVersion }} />
+            </FlexItem>
+            {hash !== '#rules' && (
```
bastilian, line 132: I'm not sure I understand why this is necessary.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
bastilian: Also, in the final version of this feature the link should only show up in the Wizard, maybe we can look for a way when we add the link there to solve this differently and not have this in ProfileTabContent directly, but passed in from higher up.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
adonispuente: Theres multiple rule table instances, I didnt want to add the 'view affected link' component to the policy details rule page

bastilian: hm. And why use a URL hash? and not a prop or something?

bastilian: We have to be careful with this, because compliance uses the URL anchor on some pages to identify the current tab.

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
bastilian: We should definitely look into providing that link differently into here. The passing down of the `policy` object looks also a bit iffy. I think for 3212 this link is not needed yet, so we could leave it out and as long as the page is viewable somehow it should good to merge.

adonispuente: hmm alright, ill look into adding it in, in another PR. Ill remove it from this one

src/constants.js
```diff
@@ -216,6 +216,7 @@ export const COMPLIANCE_TABLE_DEFAULTS = {
       });
     },
   },
+  manageColumns: true,
```
adonispuente, line 219: I kept this in since it just makes it cleaner

src/PresentationalComponents/TabbedRules/ProfileTabContent.js
bastilian: Yeah, I think we should think about what feature would we need to add to TabbedRules/RulesTable to be able and use it for adding a link here. The final version of this feature should also, i believe, offer a way of resetting the rule selection in the rules table next to the link, so it can't be something like a "ViewAffectedLink" feature, but more a "put any content here" feature.

adonispuente:
When merging I will update the commit message

src/PresentationalComponents/CheckboxGroup/__snapshots__/CheckboxGroup.test.js.snap
```diff
@@ -8,6 +8,7 @@ exports[`CheckboxFieldArray expect to render with default checked fields 1`] = `
     id="12345"
     isChecked={false}
     isDisabled={false}
+    isRequired={false}
```
bastilian, line 11: There is something fishy here. When running `rm -rf node_modules && npm install && npm run test -- -u` it'll revert these changes.

The issue is not yet visible on Travis CI, because the commit message linting is blocking the actual tests to run.

bastilian:
@adonispuente I'm afraid, you'll have to fix up the messages before we can merge this, because the actual tests are not run.

src/Routes.js
```diff
@@ -108,6 +108,16 @@ const policiesRoutes = [
     ),
     modal: true,
   },
+  {
+    path: '/policies/:policy_id/default_ruleset',
```
bastilian, line 112: Minor detail, but we can change this also in the link PR. The first part should be `/scappolicies/` to align with the other policy routes.

bastilian: We'll also need to find a way to render this page "chromeless", or check with UX if that is important.

bastilian (approved):
Nice job! LGTM! Thank you @adonispuente!

bastilian:
@xmicha82 Mind taking another look as well and give your approval.

xmicha82 (approved):
LGTM, thx @adonispuente

PR #1799, opened 2022-10-27
fix(RemediationButton): RHICOMPL-3303 - Correctly provide data

bastilian:
This refactors the RemediationButton to a function component and has the logic extracted into discrete helpers to compose the result to fix the issues with the remediations button mentioned in the Jira.

The button should now pass data to remediate correctly and only be active if something is selected, if the rules and system SSG version is supported and if the rules have a remediation available. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/CreatePolicy/__snapshots__/CreateSCAPPolicy.test.js.snap
```diff
@@ -105,34 +105,46 @@ exports[`CreateSCAPPolicy should render benchmarks and no policies until one is
                       "compliant": true,
```
bastilian, line 105: This can be safely ignored. The factories for this were extended and therefore the snapshots needed updating. Same for ReportDownload

adonispuente (commented):
When clicking any system, this error always pops up in the console.

Screen Shot 2022-10-27 at 10 38 03 AM

When pressing the remediate button, im greeted with these errors in the console and im unable to go through the remediation process.

Screen Shot 2022-10-27 at 10 39 27 AM

bastilian:
@adonispuente I'm not sure why, but compliance in development currently doesn't work and serves stage code instead of local code. Please use just `npm run start:proxy` and it should work fine.

adonispuente (approved):
This PR does allow for remediation of systems, the PR works as intended.

I'd like to note though that the bulk selector needs some love. Currently, with 254 systems it cannot actually select all systems because the graphql query times out. I see some closely-ish related tickets , but nothing about this in particular. Is this a problem we're aware of? 

Also,  im not very familiar with compliance. Does this information have anything to do with showing a user what systems are remediable? I thought maybe that this would mean the 357 systems would be remediable as they are compliant, but during testing that seems not the be the case. 
I know you listed the criteria for the button to be enabled but while I was testing I couldnt see how a user would know that. It could just be somewhere else that im not aware of though

Screen Shot 2022-10-27 at 3 11 35 PM

skateman:
Okay, query timing out is a red flag. Are you sending the query with 254 systems, or in batches of 100?

bastilian:
@skateman The additions to the query that we need for remediations to work properly again are here: https://github.com/RedHatInsights/compliance-frontend/pull/1799/files#diff-a7a8b7bbdb714abc2f3f1f113536b1283221ab92ee8acb13a759428862b6ca82

bastilian:
@skateman @adonispuente The timing out is a separate issue, this is purely to restore functionality. https://issues.redhat.com/browse/RHICOMPL-3261 is to optimise it and finally request the required data at a later point and remove the rules entirely from this query.

PR #1815, opened 2022-11-10
fix(ComplianceScore): RHICOMPL-3370 - wrap score in div to display in…

jiridostal:
Tooltip content is expected to be a single node. Wrapped the score in div and it's displayed correctly.


bastilian (approved):
Nice fix! Thank you @jiridostal! 

Only eslint seems to be unhappy. :)

PR #1835, opened 2022-11-29
feat(useFeature): Add feature flag for debugging flags

bastilian:
This is a small change to the `useFeature` hook to allow enabling and disabling it's log output via a feature flag.

By default, as it is undefined it will always be disabled, but can be enabled by appending a parameter to any url on compliance like `/insights/compliance/reports?debugFeatures=enable`.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Utilities/hooks/useFeature.js
+++ b/src/Utilities/hooks/useFeature.js
@@ -1,15 +1,30 @@
+import { useEffect, useMemo } from 'react';
 import { features } from '@/constants';
 import { useLocation, useHistory } from 'react-router-dom';
 export const LOCAL_STORE_FEATURE_PREFIX = 'insights:compliance';
 
+const getFlagValue = (feature) => {
+  const defaultValue = features[feature];
+  const storedValue = !!localStorage.getItem(
+    `${LOCAL_STORE_FEATURE_PREFIX}:${feature}`
+  );
+
+  return storedValue || defaultValue;
+};
+
 const setFeatureFlag = (featureValue, feature) => {
-  const value = featureValue === 'enable';
+  const value = featureValue === 'enable' || featureValue === 'true';
+  const debugFeatures = getFlagValue('debugFeatures');
 
   if (!value) {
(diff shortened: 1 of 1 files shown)


src/Utilities/hooks/useFeature.js
```diff
@@ -1,15 +1,30 @@
...
+  const storedValue = !!localStorage.getItem(
+    `${LOCAL_STORE_FEATURE_PREFIX}:${feature}`
+  );
+
+  return storedValue || defaultValue;
+};
+
 const setFeatureFlag = (featureValue, feature) => {
-  const value = featureValue === 'enable';
+  const value = featureValue === 'enable' || featureValue === 'true';
```
mkholjuraev, line 16: It feels to me that using only 'enabled' would be easier to grasp that a feature is being dealt with.

mkholjuraev (commented):
LGTM. Just one thing that I think is worth mentioning: enabling debugging features working, but param **?debugFeatures=enable** is not persisted in the URL. Some other url push is clearing this out. I think this needs to be fixed to have consistency.


https://user-images.githubusercontent.com/59481011/207309857-c6f41193-8bb9-4e9c-ba03-865b800a8981.mp4

src/Utilities/hooks/useFeature.js
bastilian: Added this, because I go instinctively to 'true' :D

bastilian:
@mkholjuraev Thanks for taking a look. The redirect after it set the feature flag in the local storage is intentional.

mkholjuraev (approved):
I understand.

For the future: migrating to unleash might help to manage the usage of feature flags more easily.

PR #1852, opened 2022-12-20
feat(cypress): add cypress to the compliance

Fewwy:
Adding cypress testing library to the project. So anyone could write cypress tests when they need to
## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/Reports/Reports.cy.js
```diff
@@ -0,0 +1,91 @@
...
+
+  it('expect to render emptystate', () => {
+    Cypress.config('interceptions', {
+      data: {
+        profiles: { edges: [] },
+      },
+      error: false,
+      loading: false,
+    });
+    cy.get('table').should('have.length', 1);
```
mkholjuraev, line 89: I think we can improve this test by checking if there is an empty state component as well inside the table. Now the test only verifies if there is a table.

package.json
```diff
@@ -8,6 +8,8 @@
   "dependencies": {
     "@apollo/client": "3.4.16",
     "@babel/runtime": "^7.19.4",
+    "@cypress/webpack-dev-server": "^3.1.2",
+    "@patternfly/patternfly": "^4.222.4",
```
mkholjuraev, line 12: Should not this be installed as a dev-dependency?

Fewwy: Good catch, thanks will do that

gkarat (commented):
Can we also integrate cypress to code coverage within this PR already? :-)

package.json
```diff
@@ -82,6 +82,21 @@
...
   },
+  "nyc": {
+    "report-dir": "cypress-coverage",
+    "include": [
+      "src/Mutations",
+      "src/PresentationalComponents",
+      "src/Modules",
+      "src/SmartComponents",
+      "src/__factories__",
+      "src/Utilities"
```
gkarat, line 93: Is there a reason to omit other files under `src`? Can we use just:
```suggestion
      "src/**/*"
```

package.json
```diff
@@ -82,6 +82,21 @@
...
+    ],
+    "exclude": [
+      "src/**/*.spec.ct.js",
```
gkarat, line 96: This will exclude all cypress test files from coverage 🤔 we should remove this entry
```suggestion
```

gkarat (changes requested):

PR #1853, opened 2023-01-05
Fix several empty state layout bugs and inconsistencies

gkarat:
I found and fixed a couple of places in compliance where error/empty states had bad or inconsistent layout.

## Before the change

1. On all pages, broken error state layout, and on the reports page, the header is not rendered: image
3. On the reports page, the empty state is centered with Bullseye which is not consistent with all other platform empty states (centered on top, but no vertically), the header has `h1` which is not correct (from the accessibility perspective), and also there is unnecessary extra `<br />` between the header and content: image

## After the change

1. image
2. image

### DIFF
--- a/src/PresentationalComponents/ComplianceEmptyState/ComplianceEmptyState.js
+++ b/src/PresentationalComponents/ComplianceEmptyState/ComplianceEmptyState.js
@@ -5,7 +5,6 @@ import {
   Title,
   TextContent,
   Button,
-  Bullseye,
   EmptyState,
   EmptyStateBody,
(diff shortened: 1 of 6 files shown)


Fewwy:
The PR looks good to me but you have errors in the ci.int :(

Fewwy (approved):

LightOfHeaven1994:
I found little locator change so IQE test fails, I'm releasing fix right now, after it's done, I will retest and merge this PR

gkarat:
Thanks Egor!

PR #1854, opened 2023-01-06
fix: Fix not-equal sign in error states

gkarat:
Fixes https://issues.redhat.com/browse/RHICOMPL-3415.

This makes `ErrorCard` component render the Something went wrong message if `POST /graphql` return the error code different from 404. Also, incorrect error message parameter passing is fixed.

## How to test

Run this PR together with Inventory locally. In dev tools, intentionally block the `POST /graphql` request. Go to the Inventory Details page for some system and open Compliance tab. You should see the "Something went wrong" message instead of the only not-equal sign reported in the bug ticket.

## Before the change

image

## After the change

image


gkarat:
Need to also have an ack from the compliance team on whether this implemented right for all edge cases

bastilian (approved):
LGTM! Thank you @gkarat!

PR #1862, opened 2023-01-12
fix: Unsupported systems not in system count

johnsonm325:
This PR replaces the bar chart on the reports table with the donut chart. It also fixes the counts which were previously incorrect.

The "total number of systems" considered in the main reports view do not take into account neither the reported or unsupported systems in the X of Y systems.

Though, the bar chart displayed on each of the report entries does take them into account (displaying the unsupported systems in yellow, and the never-reported systems in gray)

Steps to Reproduce:

have one unsupported system reporting
have 2 more additional systems associated to the reported policy Check reports page

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/ReportDetails/Components/hooks/useDonutChart.js
+++ b/src/SmartComponents/ReportDetails/Components/hooks/useDonutChart.js
@@ -1,4 +1,11 @@
-import { ChartThemeVariant } from '@patternfly/react-charts';
+import React from 'react';
+import propTypes from 'prop-types';
+import {
+  ChartThemeVariant,
+  ChartTooltip,
+  ChartLabel,
+} from '@patternfly/react-charts';
+import { Icon } from '@patternfly/react-core';
 import black300 from '@patternfly/react-tokens/dist/esm/global_palette_black_300';
 import blue200 from '@patternfly/react-tokens/dist/esm/chart_color_blue_200';
 import blue300 from '@patternfly/react-tokens/dist/esm/chart_color_blue_300';
@@ -6,6 +13,38 @@ import chart_color_black_200 from '@patternfly/react-tokens/dist/esm/chart_color
 import chart_color_gold_300 from '@patternfly/react-tokens/dist/esm/chart_color_gold_300';
 import { fixedPercentage } from 'Utilities/TextHelper';
 import useLegendData from './useLegendData';
+import { SquareFullIcon } from '@patternfly/react-icons';
+
+const DonutLabel = ({
+  x,
+  y,
+  datum,
+  chartColorScale,
+  flyoutValues,
+  ...rest
+}) => {
+  const iconLeftEdge = x - flyoutValues[datum._x - 1] / 2 + 5;
+  const percentOfDonut = (datum.endAngle - datum.startAngle) / 360;
+  rest.text = `${rest.text}: ${(percentOfDonut * 100).toFixed(2)}%`;
+  return (
+    <g>
+      <foreignObject height="100%" width="100%" x={iconLeftEdge} y={y - 11}>
+        <Icon>
+          <SquareFullIcon color={chartColorScale[datum._x - 1]} />
+        </Icon>
+      </foreignObject>
+      <ChartLabel x={x + 10} y={y} {...rest} />
+    </g>
+  );
+};
+
+DonutLabel.propTypes = {
+  x: propTypes.number,
+  y: propTypes.number,
+  datum: propTypes.object,
+  chartColorScale: propTypes.array,
+  flyoutValues: propTypes.array,
+};
 
 const useDonutChart = (profile) => {
   const {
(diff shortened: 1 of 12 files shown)


LightOfHeaven1994:
@johnsonm325 Could we also include never reported systems to counter? I checked the JIRA and I think there is a typo about it. We wanted to have chart bar looking the same as Donut chart in Report details page (/reports/{policy_id}). So X of Y where Y includes compliant + non-compliant + not supported + never reported.

**Optional:** Also there is another thing about percentage counting but IDK how much effort need to fix it (screenshot below). So here I have 33% pass rate but that's wrong as I have there 5 systems (1 compliant, 2 non-compliant, 1 unsupported and 1 never reported). So the percentage has to be 20% as only 1 system from 5 is compliant.

Do you think you can handle percentage in this JIRA or do I need to create a separated JIRA?
image

johnsonm325:
@LightOfHeaven1994 I spoke with Mohit and realized there's a whole bunch of things I missed here. I need to revisit this Jira card and determine how I want to break up the work. I may do it all on one PR, or I may break it up across PRs.

LightOfHeaven1994:
@johnsonm325 sounds good, thanks :)

Fewwy:
@johnsonm325 updated the pr from master and took a look at the numbers here. I think they are not correct :(
image

LightOfHeaven1994:
Charts look awesome. There are 2 minor things - 
1) if you hover part of a chart it doesn't show `State: value` and instead of this I see [ object Object]. Same happens on both - Reports page and Report details page

What I see currently:
image

Mocks:
image

**Another thing I found on ReportDetails page:**
2) We don't want to show unsupported/never reported labels if there are 0 systems for them. It works in prod now so here we have regression
image

johnsonm325:
@LightOfHeaven1994 All of this should be fixed now. @bastilian maybe you can take a look?

LightOfHeaven1994:
Looks perfect now! 
Small thing is that we are missing percentage in tooltip but it's not a blocker for this PR I think.
It would be great if someone from FE team will take a look on PR too
image

johnsonm325:
Please refer to this comment: https://issues.redhat.com/browse/RHICOMPL-3187?focusedCommentId=21611862&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21611862

src/PresentationalComponents/ReportsTable/ReportsTable.scss
```diff
@@ -0,0 +1,5 @@
+.reports-table {
```
mkholjuraev, line 1: In case there is only one CSS rule, can we have it in the React component directly? Because, loading css file blocks rendering, lowering performance.  Or we can have one general CSS file for all CSS rules of Compliance.

Fewwy:
@johnsonm325 Overall this PR LGTM, but I checked it on the account with only 1 report available, can we look at more reports together today?

src/SmartComponents/ReportDetails/Components/hooks/useDonutChart.js
```diff
@@ -1,11 +1,50 @@
...
+  x,
+  y,
+  datum,
+  chartColorScale,
+  flyoutValues,
+  ...rest
+}) => {
+  const iconLeftEdge = x - flyoutValues[datum._x - 1] / 2 + 5;
+  const percentOfDonut = (datum.endAngle - datum.startAngle) / 360;
+  rest.text = `${rest.text}: ${(percentOfDonut * 100).toFixed(2)}%`;
```
mkholjuraev, line 28: Do we always have the `text` property of the `rest` object?

Fewwy (approved):

src/SmartComponents/ReportDetails/Components/hooks/useDonutChart.js
```diff
@@ -39,6 +79,20 @@ const useDonutChart = (profile) => {
   return {
     chartProps: {
       data: donutValues,
+      labelComponent: (
+        <ChartTooltip
+          flyoutWidth={({ index }) => {
```
mkholjuraev, line 84: How about : `flyoutWidth={({ index }) => flyoutValues[index]}`

src/SmartComponents/ReportDetails/Components/hooks/useDonutChart.js
johnsonm325: Yes, I believe so.

LightOfHeaven1994 (approved):
LGTM, I checked it locally once again and it works perfectly. failing tests are happening because some selenium issues

PR #1891, opened 2023-02-09
feat: RHICOMPL-2124 - Show group categories for rules

bastilian:
This introduces tree table support for the underlying utility hooks for tables and makes use of it in the rules tables in compliance. 

How to test:

1) Append `ruleGroups=enable` to any page/url in compliance
Test a rules table on:

2) a Policy details page
3) a Systems details page
3.1) a Systems details page with multiple policies (my stage accound has such a system)
4) a Report details page
5) when editing a policy
6) in the "Create Policy" wizard.

Please do verify also with the feature flag disabled (`ruleGroups=disable`).

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


bastilian:
@LightOfHeaven1994 Do you know why the Jenkins builds keep failing here?

LightOfHeaven1994:
@bastilian yep, sorry for that. IQE rbac plugin released a broken version so it breaks for everyone. It was fixed recently, I will fix it today.

bastilian:
@LightOfHeaven1994 Thank you!

mkholjuraev (commented):
Codewise, everything seems good. Honestly, I am not able to suggest an improvement as the code looks neat and robust as always. Thus, I am focusing more on how it works and the real feel.

I feel a bit of latency on the rules tab. It takes some time to open a row and do an action on the tab. I observed a few possible regressions:

1. Scap policies and rules table do not have the empty state in this PR.
2. On the system  (demo_failing_system_pci-dss_9) detail page, I got the following error:
**Cannot read properties of undefined (reading 'id')**
3. I believe the flag **ruleGroups=enable** in the URL is not kept while I browse from page to page. This forces me to again append the flag if I want to see rule groups in system detail coming from report detail.

bastilian:
@mkholjuraev Thank you for reviewing! 

1) This issue is now fixed and the empty state should properly show again.
2) I was unable to reproduce this one. Even when recreating the same policy. 
3) The feature flag is stored in localStorage and only reset if you explicitly disable it via the same param. This has been working fine for me as well, and I could not reproduce the issue.

Regarding the latency: The responses can get very large. Right now the backend returns the whole tree regardless of the amount of rules assigned to the policy, which means it will also return branches where there would be no rules to show within. There is a Jira open to address this and have the backend only return branches that will have rules attached, which should give back some performance.

mkholjuraev (approved):
LGTM! @bastilian thank you. I have thoroughly played with the UI and could not find outstanding issues except for the bulk selection.

PR #1900, opened 2023-03-01
feat(RulesTable): RHICOMPL-3418 - Add editing of a value to rules' details component in RulesTable

bastilian:
This adds value editing for rules.

How to test:

1) Create a policy via the wizard
2) On the rules step look for a rule that has values for example:
    * Set SSH Daemon LogLevel to VERBOSE
    * Modify the System Login Banner (Should show value, but not allow editing)
3) Save the policy and navigate to it details page to the "Rules" tab
4) Search for the same rules you have set a value for and verify it shows the value(s) set
5) Test resetting the value to it's default via the reload icon
6) Edit the value again and verify it is save correctly
7) Edit another value and verify that all values are saved and persist and none are wiped
6) Open the "Edit Policy" modal via the "Edit Rules" button
8) Verify again rules are shown correctly and it is possible to edit them
9) Save and verify the policy details also reflects the changes

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Mutations/hooks/usePolicy.js
+++ b/src/Mutations/hooks/usePolicy.js
@@ -9,10 +9,13 @@ const usePolicy = () => {
   const associateSystems = useAssociateSystems();
   const associateRules = useAssociateRules();
 
-  return async (policy, updatedPolicy, onProgress) => {
-    const selectedRuleRefIds = updatedPolicy?.selectedRuleRefIds || [];
-
-    const expectedUpdates = 3 + selectedRuleRefIds.length;
+  return async (
+    policy,
+    { values = {}, selectedRuleRefIds, ...updatedPolicy },
+    onProgress
+  ) => {
+    const expectedUpdates =
+      3 + Object.keys(values).length + (selectedRuleRefIds || []).length;
     let progress = 0;
     const dispatchProgress = () => {
       if (onProgress) {
@@ -28,7 +31,7 @@ const usePolicy = () => {
 
     const mutatedPolicy = await policyMutation(
       policy?.id,
-      updatedPolicy,
+      { ...updatedPolicy, selectedRuleRefIds },
       businessObjectiveId
     );
     dispatchProgress();
@@ -42,8 +45,27 @@ const usePolicy = () => {
     } = await associateSystems(policy, updatedPolicy.hosts);
     dispatchProgress();
 
-    for (const profileSelectedRuleRefIds of selectedRuleRefIds) {
+    for (const profileSelectedRuleRefIds of selectedRuleRefIds || []) {
       await associateRules(profileSelectedRuleRefIds, profiles);
+      dispatchProgress();
+    }
+
+    for (const [profileId, profileValues] of Object.entries(values)) {
+      const realProfile = profiles.find(
+        ({ id, parentProfileId }) =>
+          id === profileId || parentProfileId === profileId
+      );
+      const convertedValues = Object.fromEntries(
+        Object.entries(profileValues).map(([valueId, valueValue]) => {
+          const refId = realProfile?.benchmark.valueDefinitions.find(
+            ({ id }) => id === valueId
+          )?.refId;
+
+          return [refId || valueId, valueValue];
+        })
+      );
+
+      await policyMutation(realProfile.id, { values: convertedValues });
       dispatchProgress();
     }
 
--- a/src/Mutations/hooks/usePolicyMutation.js
+++ b/src/Mutations/hooks/usePolicyMutation.js
@@ -6,12 +6,17 @@ const usePolicyMutation = () => {
   const [createProfile] = useMutation(CREATE_PROFILE);
 
   return async (id, updatedPolicy, businessObjectiveId) => {
-    const { name, description, complianceThreshold } = updatedPolicy;
+    const { name, description, complianceThreshold, values } = updatedPolicy;
     const details = {
-      name,
-      description,
-      businessObjectiveId,
-      complianceThreshold: parseFloat(complianceThreshold),
+      ...(name && { name }),
+      ...(description && { description }),
+      ...((businessObjectiveId || businessObjectiveId === null) && {
+        businessObjectiveId,
+      }),
+      ...(complianceThreshold
+        ? { complianceThreshold: parseFloat(complianceThreshold) }
+        : {}),
+      ...(values && { values }),
     };
 
     let mutatedPolicy;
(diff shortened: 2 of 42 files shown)


src/SmartComponents/PolicyDetails/PolicyDetails.js
```diff
@@ -105,7 +115,7 @@ export const PolicyDetails = ({ route }) => {
       (profile) => !!profile.osMinorVersion
     );
   }
-
+  console.log('POL', policy);
```
mkholjuraev, line 118: To be removed :)

src/PresentationalComponents/TabbedRules/helpers.js
mkholjuraev, line 1: To be removed ?

mkholjuraev (commented):
The code looks robust, @bastilian can you add testing instructions? I do not see it in the ticket as well

marleystipich2:
I did some testing for this because I needed screenshots for the rule group/value editing blog and I ran into an issue when I was editing a value and then went to try to reset the value to default. The value appeared to save correctly when I edited it but when I opened the value back up to reset it to the default, the checkmark on the right of the value edit field was replaced with the loading circle and it doesn't ever stop loading.

I know there are issues saving the value correctly right now so maybe this is why but just wanted to let you know in case you didn't notice this!

bastilian:
@marleystipich2 @LightOfHeaven1994 The rule editing should now all work as expected.

marleystipich2 (commented):
I tested and everything is looking great! I noticed 2 small issues though when I was testing: 
- If there are 2 values that are editable on a rule and you update them both, the reset to default button only resets one of the values back to it's default (This problem only occurs on the create policy modal, not in the other places where you can edit values)
- When you type in a new value and then click the 'X', the new value stays there instead of switching back to the current value. It doesn't seem like the value is actually saved though but it does stay on the page. This occurred in all the areas where you can edit a value.

src/Mutations/hooks/usePolicyMutation.js
```diff
@@ -6,12 +6,17 @@ const usePolicyMutation = () => {
...
-    const { name, description, complianceThreshold } = updatedPolicy;
+    const { name, description, complianceThreshold, values } = updatedPolicy;
     const details = {
-      name,
-      description,
-      businessObjectiveId,
-      complianceThreshold: parseFloat(complianceThreshold),
+      ...(name && { name }),
+      ...(description && { description }),
+      ...((businessObjectiveId || businessObjectiveId === null) && {
```
adonispuente, line 13: why implement if buisinessObjectiveId is null?

src/Mutations/hooks/usePolicy.js
```diff
@@ -42,11 +45,30 @@ const usePolicy = () => {
     } = await associateSystems(policy, updatedPolicy.hosts);
     dispatchProgress();
 
-    for (const profileSelectedRuleRefIds of selectedRuleRefIds) {
+    for (const profileSelectedRuleRefIds of selectedRuleRefIds || []) {
       await associateRules(profileSelectedRuleRefIds, profiles);
       dispatchProgress();
     }
 
+    for (const [profileId, profileValues] of Object.entries(values)) {
```
adonispuente, line 53: while your adding failsafes we can ad `|| []` here too

src/PresentationalComponents/RulesTable/RuleValueEdit.js
```diff
@@ -0,0 +1,101 @@
...
+              style={{ 'margin-left': '5px' }}
+            >
+              <PencilAltIcon />
+            </Button>
+          )}
+          {nonDefaultValues && (
+            <Button
+              onClick={resetValues}
+              variant="plain"
+              style={{ 'margin-left': '5px' }}
```
adonispuente, line 56: we can replace these style tags with `className=pf-u-ml-sm`

src/PresentationalComponents/TabbedRules/helpers.js
adonispuente, line 1: This file is created but not used/empty

src/Mutations/hooks/usePolicyMutation.js
bastilian: `null` is used in order to unset the business objective via the API, so if it is a `null` it should be passed. It is separate because `null` is also "falsy".

src/Mutations/hooks/usePolicy.js
bastilian: This is already taken care of on line 14.

src/PresentationalComponents/RulesTable/RuleValueEdit.js
bastilian: Applied.

src/PresentationalComponents/TabbedRules/helpers.js
bastilian: Removed.

bastilian:
@marleystipich2 The issues when there are two values to edit should now be resolved, and closing the edit will revert the value now back to the previous value.

@LightOfHeaven1994 Values that are edited to '0' should now be saveable.

@adonispuente The things you pointed out should be resolved too. 

Thank you all for reviewing and testing!

bastilian:
@LightOfHeaven1994 Also, the saving bug you found in the Edit Policy should work now as intended.

LightOfHeaven1994:
> @LightOfHeaven1994 Also, the saving bug you found in the Edit Policy should work now as intended.

Great, I'm going to take a look quickly once again and will let you know

LightOfHeaven1994:
@bastilian I took a look on all issues were mentioned here or  by myself. All of them are resolved! 

Last 1 thing I found is happening when you open Edit Rules modal. Now it saves values but it doesn't use values from the initial state. Example:

You have **OSPP policy with assgined RHEL7.8 host.** 
1) You open Policy details, searching for rule **CCE-27455-5** that have value X. You change this default value X to a new custom Y.
2) Then you think that you forget to change the value for another rule **CCE-80222-3** - and for example you open Edit rules modal.
3) When modal is opened you could see that for rule you changed on policy details you have your custom value Y (that's correct).
4) Now you are searching for second rule **CCE-80222-3** and setting value for it.

Now you expect when you save changes you will have 2 values set:
**CCE-27455-5** has value Y as it was
**CCE-80222-3** also has a new value as you edited it on Edit modal

But you see that only there was only 1 value set in the GQL request and it's for a new rule **CCE-80222-3**
Value for **CCE-27455-5** was reset to default X.

All steps captured on attached video
[Screencast from 2023-04-27 11-26-01.webm](https://user-images.githubusercontent.com/33912805/234822839-c08af942-8b82-4b76-831f-4a355f3333e9.webm)

bastilian:
@LightOfHeaven1994 This should be addressed now as well and existing values are passed properly.

LightOfHeaven1994:
LGTM from my side. Great work! Ready to move it to stage to test it once again there

adonispuente (approved):
LGTM, nice work!

PR #1948, opened 2023-06-15
fix(RHICOMPL-3728): Disable selections in Edit Systems for policy

johnsonm325:
Description of problem
Selecting systems in the "Edit systems" modal for an OpenSCAP policy is broken. Clicking on the checkboxes next to each system doesn't reliably result in that checkbox being ticked.

How reproducible
Always.

Steps to Reproduce:

- Create an OpenSCAP policy.
- Navigate to the "systems" tab for the policy.
- Click the "edit systems" button on that page.
- Start clicking checkboxes next to systems.

Actual results
The selected checkboxes might be toggled, or they might not. They might toggle immediately, or some seconds later, and they might toggle zero times, or once, or twice.

Expected results
The selected checkbox toggles immediately, and exactly once.

The issue is that we have to update the "Rules" tab in the "Edit Systems" modal with each system selection. The bug happens when clicking a second system while the api call is still being made to update the "Rules" tab.

The fix is to disable system selections while the api call is being made.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/EditPolicy/EditPolicySystemsTab.js
```diff
@@ -69,6 +70,7 @@ const EditPolicySystemsTab = ({ policy, onSystemSelect, selectedSystems }) => {
         remediationsEnabled={false}
         preselectedSystems={selectedSystems}
         onSelect={onSystemSelect}
+        setDisabledSelection={setDisabledSelection}
```
bastilian, line 73: This is a bit odd. Why would the action needed to be passed in? Could it not be directly import it only in the `SystemsTable`?

johnsonm325: So, I pass it down to SystemsTable because we only want to use this `setDisabledSelection` action from the `EditPolicySystemsTab`, so by passing it down, we can check whether `setDisabledSelection === undefined` to determine if we will use the action. Hopefully that makes sense.

bastilian: @johnsonm325 I think, we want this behaviour on every SystemsTable that allows selecting systems, like in the "Create Policy", "Report Details" and the "Edit Policy" systems. If we want to enable it only on specific places we should rather add a prop to activate it and import the action within the SystemsTable

johnsonm325:
@mkholjuraev I think that this behavior may be expected. In your example, the list of systems on SCAP policy (the policy details, not the modal you have opened) should only have 1 system. So when you open up the Edit Systems modal, that system from the table should be pre-selected somewhere in the systems table in the modal.

Hopefully that makes sense.

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -23,6 +23,7 @@ import {
 } from './hooks';
 import { constructQuery } from '../../Utilities/helpers';
 import { COMPLIANCE_REPORT_TABLE_ADDITIONAL_FILTER } from '../../constants';
+import { setDisabledSelection } from '../../store/Actions/SystemActions';
```
bastilian, line 26: This can be directly in the `hooks.js`. There seems to be no need to pass it down like this.

AsToNlele (approved):
LGTM, good job! :smile:

PR #1965, opened 2023-07-14
feat(ESSNTL-3718): Add Inventory group filter and column

Fewwy:
Implements https://issues.redhat.com/browse/ESSNTL-3718.

Column: "Group", non-sortable, hidden if the unleash feature flag is set to false.
Filter: multiple-choice, renders chips, hidden if the unleash feature flag is set to false, mustn't update URL.

- [x] /systems: support column
- [x] /systems: support filter
- [x] /reports/%id: support column
- [x] /reports/%id: support filter
- [x] /scappolicies/new: support column
- [x] /scappolicies/new: support filter

## How to test

In the views that are updated with the new filter, try to filter out hosts by groups. You can create groups/add new hosts beforehand. Check that the hosts have correct group name in the Group column.

## Screenshots

/systems
image

/scappolicies/new
image

/reports/%id
image

### DIFF
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -44,6 +44,9 @@ export const ComplianceSystems = () => {
                   Columns.customName({
                     showLink: true,
                   }),
+                  Columns.inventoryColumn('groups', {
+                    requiresDefault: true,
+                  }),
                   Columns.inventoryColumn('tags'),
                   Columns.OS,
                   Columns.Policies,
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -66,6 +66,7 @@ export const ComplianceSystems = () => {
                 enableEditPolicy={false}
                 remediationsEnabled={false}
                 policies={policies}
+                showGroupsFilter
               />
             )}
           </StateViewPart>
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -82,8 +82,21 @@ export const useFetchSystems = ({
 }) => {
   const client = useApolloClient();
 
-  return (perPage, page, requestVariables = {}) =>
-    client
+  return (perPage, page, requestVariables = {}) => {
+    let filter = variables.filter;
+
+    // additional filtering can be sent within requestVariables
+    if (
+      requestVariables.filter !== undefined &&
+      typeof requestVariables.filter === 'string'
+    ) {
+      filter = [
+        ...filter.split(' and '),
+        ...requestVariables.filter.split(' and '),
+      ].join(' and ');
+    }
+
+    return client
       .query({
         query,
         fetchResults: true,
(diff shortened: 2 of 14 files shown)


src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -157,3 +158,15 @@ export const inventoryColumn = (column, props) => ({
   key: column,
   ...props,
 });
+
+export const Groups = compileColumnRenderFunc({
+  title: 'Group',
+  key: 'groups',
+  exportKey: 'groups',
+  props: {
+    width: 20,
```
gkarat, line 167: Do we really need to set this width? Looks better without :)

image

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -44,6 +44,7 @@ export const ComplianceSystems = () => {
                   Columns.customName({
                     showLink: true,
                   }),
+                  Columns.Groups,
```
gkarat, line 47: I can see that with the feature flag disabled, the column will persist.

image

I am also afraid this will behave the same way for all other places you added this column (as I see from your PR). We have to hide the column if the feature flag is set to false.

gkarat (changes requested):
The filter works fine, and I see the parameter added to API requests. However, the response is not affected by this parameter, and the column is always N/A. Let's make sure we at least hide everything groups-related with the disabled feature flag. And then, either within this PR or another, make sure we propagate the group information for each host to appropriate column.

gkarat:
Not yet ready for review

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -82,8 +83,21 @@ export const useFetchSystems = ({
 }) => {
   const client = useApolloClient();
 
-  return (perPage, page, requestVariables = {}) =>
-    client
+  return (perPage, page, requestVariables = {}) => {
+    let filter = cloneDeep(variables.filter);
```
bastilian, line 87: The filter string should already be fully compiled at this point, or?

bastilian: ... Should have marked the whole section. :D I meant that this here shouldn't be necessary. Assembling the filter should happen before the fetch function is called.

Fewwy (commented):
LGTM

Fewwy (commented):
LGTM

src/SmartComponents/SystemsTable/hooks.js
gkarat: @bastilian, I am not really sure how to do that since the inventory filter values come separately from the compliance filters. So far useGetEntities -> buildApiFilters is the only place where I can catch the inventory filters and then merge it with the compliance ones together (within `filter`).

I do see the proper way is to refactor the current `useSystemsFilter` but it will take me way more time... Perhaps we can put it as TODO item in the code? How do you feel about this solution?

bastilian: "useGetEntities -> buildApiFilters" should be the only place where it is needed to combine the filters. So, I'm confused why we need it here too.

gkarat: How you will get the inventory filters before passing it to useSystemsFilter? Selectors?

gkarat (approved):

PR #1985, opened 2023-08-30
fix(ESSNTL-3718): Wrap group names in brackets

gkarat:
Fixes https://issues.redhat.com/browse/ESSNTL-3718.

Fixes improper group_name filter construction in gq filter.

## How to test

Have a group which name has spaces (like "Test group"). Add some compliance systems to it, and make sure you are able to filter by this group on /compliance/systems.


AbdullahWins (commented):
Seems alright but why would u need to do that!

LightOfHeaven1994 (approved):
tested locally, works! Thanks

LightOfHeaven1994:
> Seems alright but why would u need to do that!

Filter didn't work with the group names that have space in it, so quotes are needed

PR #1991, opened 2023-09-19
chore: Refactor PresentationalComponent tests to testing-library

bastilian:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


package.json
```diff
@@ -128,7 +128,6 @@
     "jest-environment-jsdom": "^29.6.2",
     "natsort": "^2.0.3",
     "npm-run-all": "^4.1.5",
-    "react-test-renderer": "^17.0.2",
```
gkarat, line 131: What's the reason for removing it? I think we can keep support of snapshot tests

src/PresentationalComponents/ComplianceEmptyState/ComplianceEmptyState.test.js
```diff
@@ -1,21 +1,34 @@
...
+
 describe('ComplianceEmptyState', () => {
   it('expect to render without error if no policies exist', () => {
     useQuery.mockImplementation(() => ({
       data: { profiles: { totalCount: 0 } },
       error: undefined,
       loading: undefined,
     }));
-    const wrapper = shallow(<ComplianceEmptyState client={{}} />);
+    const { container } = render(<ComplianceEmptyState client={{}} />);
```
gkarat, line 29: I think in this test where you don't have any further nesting, you can just import `screen` from @testing-library/react and access the queries through `screen.queryByText('No policies')`

gkarat: IIRC, this is more recommended way of accessing queries than spreading the `container` object

src/Utilities/hooks/useTableTools/testHelpers.js
```diff
@@ -1,14 +1,18 @@
...
+import { render, act } from '@testing-library/react';
 import { queryByText } from '@testing-library/dom';
 import userEvent from '@testing-library/user-event';
 
 const toggleFilterDropDown = (container) => {
   const filterToggle = container.querySelector('button.pf-c-dropdown__toggle');
-  userEvent.click(filterToggle);
+  act(() => {
+    userEvent.click(filterToggle);
+  });
```
gkarat, line 9: What's the reason for this change? Just want to understand more the specifics of `act`

package.json
bastilian: Only the `ProgressBar` test was using it, and we should try to stick to one framework.

src/Utilities/hooks/useTableTools/testHelpers.js
bastilian: This was pointed out in a warning when running any test that invokes these test helpers, like the ReportsTable tests.

src/PresentationalComponents/ComplianceEmptyState/ComplianceEmptyState.test.js
bastilian: I'll go over those tests again and replace the queries to query via `screen`

package.json
gkarat: Fair enough

src/PresentationalComponents/ComplianceScore/ComplianceScore.test.js
```diff
@@ -10,8 +13,9 @@ describe('auxiliary functions to reducer', () => {
       compliant: false,
     };
 
-    const dangerIcon = mount(<ComplianceScore {...system} />);
-    expect(toJson(dangerIcon)).toMatchSnapshot();
+    const { container } = render(<ComplianceScore {...system} />);
+
+    expect(queryByText(container, '10%')).not.toBeNull();
```
gkarat, line 18: nit-pick:
```suggestion
    expect(screen.getByText('10%'));
```
should be almost an equal change since `get` queries will error if there are no elements or more than one element found.

Also works for many other places where you import `queryByText` from @testing-library/dom (will not comment on those).

src/PresentationalComponents/ReduxFormWrappers/ReduxFormWrappers.test.js
```diff
@@ -7,24 +9,26 @@ describe('ReduxFormTextInput', () => {
         onChange: jest.fn(),
         value: 'Value',
       },
+      'aria-label': 'text-input',
```
gkarat, line 12: nit-pick: This would be ideally added to the implementation instead with some meaningful label description. I am also not sure you utilise this in the test at all (querying by the `input` tag instead).

src/PresentationalComponents/ReportsTable/ReportsTable.test.js
```diff
@@ -11,31 +13,34 @@ import { filterHelpers } from 'Utilities/hooks/useTableTools/testHelpers.js';
...
-
-    expect(toJson(wrapper)).toMatchSnapshot();
+    const { container } = render(<ReportsTable profiles={profiles} />);
+
+    expect(
+      queryByText(
+        container.querySelector('tbody>tr button'),
+        'C2S for Red Hat Enterprise Linux 7'
+      )
+    ).not.toBeNull();
```
gkarat, line 43: nit-pick:
```suggestion
    getByRole('button', { name: 'C2S for Red Hat Enterprise Linux 7' });
```

src/PresentationalComponents/StateView/StateView.test.js
```diff
@@ -16,11 +19,13 @@ describe('StatView', () => {
       </StateView>
     );
 
-    expect(toJson(wrapper)).toMatchSnapshot();
+    expect(queryByText(container, 'LOADING')).not.toBeNull();
+    expect(queryByText(container, 'DATA')).toBeNull();
```
gkarat, line 23: nit-pick:
```suggestion
    expect(queryByText('DATA')).not.toBeInTheDocument();
```

but this would also require importing '@testing-library/jest-dom'

src/PresentationalComponents/TabSwitcher/TabSwitcher.test.js
```diff
@@ -19,107 +20,107 @@ jest.mock(
   () => () => ({})
 );
 
-describe('TabSwitcher', () => {
+describe.skip('TabSwitcher', () => {
```
gkarat, line 23: This is most probably not intended

src/Utilities/hooks/useTableTools/testHelpers.js
gkarat, line 9: nit-pick:
```suggestion
    await userEvent.click(filterToggle);
```

would also require marking the function as async

gkarat (commented):
LGTM!

- There are several things that can be improved because RTL provides more neat approach to some queries (see my comments), we can consider fixing them in the future. 
- I would also recommend turning on the RTL eslint plugin, though it will mark many places with errors. This would provide you with some linter auto-correction. For example, it wraps the test function immediately in async if you use any userEvent.* functions. Or replaces
```
expect(...).not.toBeVisible()
```
to
```
expect(...).not.toBeInTheDocument()
```
- Combining the linter with logTestingPlayground suggested queries will align the code a lot to the RTL recommendations.

- I didn't look at the deleted tests assuming you know what you are doing hehe
- Please take a look at the .skip occurrences: we most probably want to avoid all of them:
Screenshot 2024-01-22 at 12 44 05

src/Utilities/hooks/useTableTools/testHelpers.js
bastilian: These helpers need to be completely rewritten. I tried, but I think it'll be done in a following PR.

src/PresentationalComponents/TabSwitcher/TabSwitcher.test.js
bastilian: This one is un-skipped.

src/PresentationalComponents/StateView/StateView.test.js
bastilian: Done.

src/PresentationalComponents/ReportsTable/ReportsTable.test.js
bastilian: Done.

src/PresentationalComponents/ComplianceScore/ComplianceScore.test.js
bastilian: Done.

src/PresentationalComponents/ReduxFormWrappers/ReduxFormWrappers.test.js
bastilian: This is a bit of a special case. These are just wrapper components that should only add props required for use with redux-form. Adding an aria-label with a default would not be needed for that.

package.json
```diff
@@ -106,8 +106,8 @@
     "@redhat-cloud-services/frontend-components-config": "^5.1.1",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^12.1.5",
-    "@testing-library/react-hooks": "^7.0.2",
-    "@testing-library/user-event": "^13.5.0",
+    "@testing-library/react-hooks": "^8.0.1",
+    "@testing-library/user-event": "^12.8.3",
```
bastilian, line 110: These are set to the most recent and compatible versions for now, just to avoid any incompatibilities, but they will be upgraded to the latest with react 18.

src/SmartComponents/SystemsTable/Columns.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable testing-library/render-result-naming-convention */
```
bastilian, line 1: This linter is triggered here due to the function names including "render".

bastilian:
@gkarat Thank you for reviewing! The comments should be all addressed now. 

The majority of the Presentational components test are migrated with this PR and another PR is started to continue the migration here: https://github.com/RedHatInsights/compliance-frontend/pull/2026

src/Utilities/TestWrapper.js
AsToNlele, line 1: Thanks for adding this!

AsToNlele (approved):
Well done @bastilian! :smile: LGTM

PR #2011, opened 2023-11-10
feat(RHINENG-1840): Support sort by group column (#679)

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-1840.

Recognise group column sort parameter (when available). This updates the systems table on the following views:
- /systems
- /reports/%id
- /scappolicies/new

## How to test

The Inventory change is not yet merged to ensure that the tenant apps support sorting first and avoid breaking change.

1. Pull this Inventory PR locally https://github.com/RedHatInsights/insights-inventory-frontend/pull/2070.
2. Run Compliance with this PR `npm run start -- --port=8004`
3. Run Inventory with `LOCAL_APPS=compliance:8004~http npm run start:proxy`
4. Navigate to the updated views (in stage-stable environment) and make sure you can sort by system groups.

## Screenshots

image

image

image

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -115,8 +115,12 @@ export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
   const appendDirection = (attributes, direction) =>
     attributes.map((attribute) => `${attribute}:${direction}`);
 
-  const findColumnByKey = (key) =>
-    (columns || []).find((column) => column.key === key);
+  const findColumnByKey = (sortKey) =>
+    (columns || []).find(
+      (column) =>
+        column.key === sortKey || // group column has a sort key different to its main key
+        (column.key === 'groups' && sortKey === 'group_name')
+    );
 
   return async (
     _ids,
--- a/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
+++ b/src/SmartComponents/ComplianceSystems/ComplianceSystems.js
@@ -46,6 +46,7 @@ export const ComplianceSystems = () => {
                   }),
                   Columns.inventoryColumn('groups', {
                     requiresDefault: true,
+                    sortBy: ['groups'],
                   }),
                   Columns.inventoryColumn('tags'),
                   Columns.OS,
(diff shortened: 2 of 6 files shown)


src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -115,8 +115,12 @@ export const useGetEntities = (fetchEntities, { selected, columns } = {}) => {
   const appendDirection = (attributes, direction) =>
     attributes.map((attribute) => `${attribute}:${direction}`);
 
-  const findColumnByKey = (key) =>
-    (columns || []).find((column) => column.key === key);
+  const findColumnByKey = (sortKey) =>
+    (columns || []).find(
+      (column) =>
+        column.key === sortKey || // group column has a sort key different to its main key
+        (column.key === 'groups' && sortKey === 'group_name')
```
gkarat, line 122: Here is a tricky part, @bastilian. 

We use the "groups" key to merge the column parameters with the inventoryColumn default parameters (see https://github.com/RedHatInsights/compliance-frontend/pull/2011/files#diff-2364f2b37d7c47fd001d20bb85cb3efa52d6b4b3357c50639488b9a8d22f12eeR172 for instance). 

But when it comes to sorting, Inventory sends the value of the `sortKey` parameter: https://github.com/RedHatInsights/insights-inventory-frontend/pull/2070/files#diff-eab6c7546d4cc71833c6530472a903e2c20922dbc85174697715832d2ed6286aR55 which is "group_name" in our case, not "groups".  By "sends the value" I mean what tenant apps receive in getEntities function from orderBy parameter. This is why findColumnByKey won't recognise the group column when sorted by it.

This solution works, but it introduces additional condition which is not really good. But not sure, how to reimplement it properly.

bastilian: I'm not sure I fully understand the issue...

bastilian: Why can it not just be `sortBy: ['group_name'],` above in the columns?

gkarat: @bastilian, compliance GraphQL requires the sorting parameter to be 'groups', not 'group_name'

PR #2016, opened 2023-11-30
chore(fec): FEC bump

adonispuente:
Compliance has the latest FEC code, this just removes code in Compliance that it is already importing from FEC.


bastilian:
I'm closing this to clean up the compliance PRs a bit and this is obsolete now.

PR #2035, opened 2024-02-07
fix(SystemsTable): RHINENG-1923 fix os filtering

xmicha82:
Fixes https://issues.redhat.com/browse/RHINENG-1923.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -3,6 +3,7 @@ import packageJson from './../package.json';
 import { conditionalFilterType } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
 import { dispatchNotification } from 'Utilities/Dispatcher';
 import sortBy from 'lodash/sortBy';
+import { cloneDeep } from 'lodash';
 
 export const APP_ID = 'compliance';
 export const DEFAULT_TITLE = 'Compliance | Red Hat Insights';
@@ -116,7 +117,7 @@ const toSystemsOsMinorFilterConfigurationItem =
       .reverse()
       .map((minorVersion) => ({
         label: `RHEL ${majorVersion}.${minorVersion}`,
-        value: minorVersion,
+        value: `${majorVersion}.${minorVersion}`,
       })),
   });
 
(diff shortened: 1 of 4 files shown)


src/constants.js
```diff
@@ -136,19 +136,72 @@ const emptyFilterDropDownItem = {
...
+        .reduce((acc, o) => {
+          acc[`${o.minor}`] = true;
+          return acc;
+        }, {});
+
+      return acc;
+    } else {
+      acc[majorVersion] = Object.entries(minorVersions).reduce(
+        (acc, [minorVersion, toInclude]) => {
+          acc[minorVersion.split('.')[1]] = toInclude;
```
gkarat, line 176: Just a thought: we could explore https://github.com/npm/node-semver. It already provides with a lot of helpful functions like `minor`. This will change the code quite a lot, but it can make us more secure when working with versions.

@bastilian, @xmicha82, maybe you'd have some opinion around that? Or we better not to invest time in it?

gkarat (changes requested):
This seems to be not working for the systems page: when I click on the RHEL 7 major version, it 1) doesn't select all the checkbox for the 7.x minor versions, 2) includes hosts with the version RHEL 8.x in the results.

https://github.com/RedHatInsights/compliance-frontend/assets/31385370/846885de-a0b1-4fe5-9b98-8ca49309fa9a

src/constants.js
```diff
@@ -116,7 +116,7 @@ const toSystemsOsMinorFilterConfigurationItem =
       .reverse()
       .map((minorVersion) => ({
         label: `RHEL ${majorVersion}.${minorVersion}`,
-        value: minorVersion,
+        value: `${minorVersion}.${majorVersion}`,
```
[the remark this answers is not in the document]
xmicha82: that is indeed right, must've messed that one up somehow

src/constants.js
bastilian: @gkarat I think for now we don't need to do anything complex. These OS version come from the SCAP Policies (not really the systems) imported into Compliance, which will probably always be for RHEL and should work here.

gkarat (commented):
When filtering by OS, there is the same issue: I select RHEL 8 checkbox, but the results include systems of version 7 too (Compliance > Systems page)

This can give some hint: I can see that the GQL request doesn't contain any information about the OS. It is just blank `()` value

image

PR #2045, opened 2024-03-20
chore(RHINENG-8630): install axios version ^1.6.8

mkholjuraev:

Installs Axios version 1.6.8 as a direct dependency. Before this, Axios was not installed, and probably the app was using Axios instance from a shared context or some peer dependency.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/config/helpers.js
+++ b/config/helpers.js
@@ -27,7 +27,14 @@ const includeFedMods = (plugins, fedMods) => ({
         exposes: fedMods,
         shared: [
           {
-            'react-router-dom': { singleton: true, requiredVersion: '*' },
+            'react-router-dom': {
+              singleton: true,
+              requiredVersion: '*',
+            },
+            axios: {
+              singleton: true,
+              requiredVersion: '^1.6.8',
+            },
           },
         ],
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@redhat-cloud-services/frontend-components-utilities": "^3.7.4",
         "@types/react": "^17.0.50",
         "apollo-boost": "^0.4.9",
+        "axios": "^1.6.8",
         "classnames": "^2.3.2",
         "graphql": "^15.8.0",
         "graphql-tag": "^2.12.6",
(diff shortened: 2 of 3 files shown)


gkarat (commented):
@mkholjuraev, this doesn't lock the transitive axios dependency: other packages can still use versions different to what we have listed in this PR. You should use "overrides" instead that forces all dependencies to fulfil the version criteria: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

mkholjuraev:
> @mkholjuraev, this doesn't lock the transitive axios dependency: other packages can still use versions different to what we have listed in this PR. You should use "overrides" instead that forces all dependencies to fulfil the version criteria: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

@gkarat thank you for the review! I wanted to avoid customizations of dependencies. I believe that Axios is included in the shared scope, so the first version loaded into that scope would be used by the Compliance as well. Do we still want to override the package version?

gkarat:
@mkholjuraev, ah, I thought we are only dealing with dependencies used in a build time. In that case we have to look at the problem from two angles:

- to ensure compliance is built using axios which is not affected by the CVE by providing `overrides`: this way, all npm dependencies, if they have their own axios dependency, will adjust their axios versions to use the one we set in `overrides`. This might lead to required versions conflict (if for example dependency A locks axios to be some exact version, like "1.2.3", not "^1.2.3"), but given it's a simple patch/minor version change, there will be most probably no issues.
- for the shared federated modules scope: I am not really sure we really mark `axios` as shared. If this was the case, we would have listed it here https://github.com/RedHatInsights/compliance-frontend/blob/master/config/helpers.js#L28-L32. This means, compliance always uses its own axios dependency and it ignores axios packages coming from other fed modules if present.

PR #2065, opened 2024-05-23
feat(APIV2): useQuery proposal

AsToNlele:
~~While updating the `compliance-client` I broke the release process~~
There's a different issue with the `compliance-client`, so until it's fixed I've made an manual axios call.

This is just a proposal, open to changes

Added testing values to systems page
The query makes an initial call and is rerun on param change on refetch. In console you can see the results changing.

Mainly to discuss the `useQuery` hook.


src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
```diff
@@ -52,42 +53,77 @@ export const useFetchSystems = ({ query, onComplete, variables, onError }) => {
...
-      .query({
-        query,
-        fetchResults: true,
-        fetchPolicy: 'no-cache',
-        variables: {
-          perPage,
-          page,
-          ...(combinedVariables ? combinedVariables : {}),
-        },
+    return apiInstance
```
bastilian, line 59: I think maybe we should start with Reports or Policies first where we can really replace the `useQuery`. 

I would maybe suggest to doe the `SystemsTable` last and maybe also try to migrate it away from using a plain graphic client or later a direct axios instance.

PR #2077, opened 2024-06-18
chore(tests): Improve coverage reporting

bastilian:
This changes the `coverage.sh` script to also run the test suite (jest + cypress) to gather coverage and combines into a html report under `./coverage`. The three reports are also uploaded to `codecov` with separate flags, which will allow us to compare the reports also on `Codecov`.

It also cleans up includes/excludes for both coverage reports to be aligned and not include unnecessary files.

Babel and other dependencies needed to be updated to have everything working correctly, for some reason.

**How to test:**

1) Run `./coverage.sh`
2) Open the HTML reports in `coverage`, `coverage-cypress` & `coverage-jest`
3) Verify that the report makes sense and the percentages "add up" in the combined report.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -1,18 +1,53 @@
-#!/bin/sh -ex
+#!/bin/bash
+#
+# Merges coverage reports from cypress and jest tests
+# and generates an HTML report to view and json report to submit to codecov
+#
+case $(uname) in
+Darwin)
+        export CODECOV_BIN="https://uploader.codecov.io/latest/macos/codecov"
+        ;;
+Linux)
+        export CODECOV_BIN="https://uploader.codecov.io/latest/linux/codecov"
+        ;;
+esac
 
-# Merge coverage reports from cypress and jest tests
-mkdir -p reports
-cp cypress-coverage/coverage-final.json reports/from-cypress.json
-cp jest-coverage/coverage-final.json reports/from-jest.json
+rm -rf ./nyc_output
+rm -rf ./coverage-jest
+rm -rf ./coverage-cypress
+rm -rf ./coverage
 
-npx nyc merge reports .nyc_output/out.json
-npx nyc report --reporter json --report-dir coverage
+mkdir -p coverage/src
 
-# Upload the final coverage report to codecov https://docs.codecov.com/docs/codecov-uploader
-curl -Os https://uploader.codecov.io/latest/linux/codecov
-chmod +x codecov
+npm run test
+cp ./coverage-jest/coverage-final.json ./coverage/src/jest.json
 
-# Workaround for https://github.com/codecov/uploader/issues/475
-unset NODE_OPTIONS
+if [ -f cypress.config.js ]; then
+        npm run test:ct
+        cp ./coverage-cypress/coverage-final.json ./coverage/src/cypress.json
+else
+        echo "No cypress config found!"
+fi
 
-./codecov -t ${CODECOV_TOKEN} -f "coverage/coverage-final.json"
+npx nyc merge ./coverage/src ./coverage/coverage-final.json
+npx nyc report -t ./coverage --reporter html --report-dir ./coverage/html
+
+if [ -z "${TRAVIS}" ]; then
+        echo "Not on travis..."
+else
+        # Workaround for https://github.com/codecov/uploader/issues/475
+        unset NODE_OPTIONS
+        echo "Downloading codecov binary from $CODECOV_BIN"
+
+        curl -Os "$CODECOV_BIN"
+        chmod +x codecov
+
+        echo "Uploading to codecov"
+
+        ./codecov --verbose -F "combined" -t "$CODECOV_TOKEN" -f ./coverage/coverage-final.json
+        ./codecov --verbose -F "jest" -t "$CODECOV_TOKEN" -f ./coverage-jest/coverage-final.json
+
+        if [ -f cypress.config.js ]; then
+                ./codecov --verbose -F "cypress" -t "$CODECOV_TOKEN" -f ./coverage-cypress/coverage-final.json
+        fi
+fi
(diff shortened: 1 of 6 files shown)


coverage.sh
```diff
@@ -1,18 +1,32 @@
...
-# Upload the final coverage report to codecov https://docs.codecov.com/docs/codecov-uploader
-curl -Os https://uploader.codecov.io/latest/linux/codecov
-chmod +x codecov
+npx nyc merge ./coverage/src ./coverage/final.json
+npx nyc report -t ./coverage --reporter json --report-dir ./coverage/json
+npx nyc report -t ./coverage --reporter html --report-dir ./coverage/html
+
+case `uname` in
+  Darwin)
+    export CODECOV_BIN="https://uploader.codecov.io/latest/macos/codecov"
```
gkarat, line 16: The script is meant to be run by Travis which uses linux distributive. Why we need this change?

bastilian: So I can test it locally :D

package.json
```diff
@@ -135,25 +137,20 @@
...
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint config src",
     "lint:js:fix": "eslint config src --fix",
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
-    "prod": "NODE_ENV=production webpack serve --config config/prod.webpack.config.js",
-    "server:ctr": "node src/server/generateServerKey.js",
     "start": "fec dev",
     "start:proxy": "PROXY=true fec dev",
-    "travis:verify": "npm-run-all build lint test && npm run test:ct -- --browser chrome",
+    "travis:verify": "npm-run-all lint test && npm run test:ct -- --browser chrome && npm run build",
```
gkarat, line 148: Why the build step is put as the last? I would expect for CI to fail if the build was not successful. That also means there is no need to execute tests

config/cypress.webpack.config.js
```diff
@@ -25,6 +25,11 @@ module.exports = {
           },
         },
       },
+      {
+        test: /\.(?:js|mjs|cjs)$/,
+        exclude: /(node_modules|bower_components)/i,
+        use: ['babel-loader'],
+      },
```
gkarat, line 32: Isn't this provided from @redhat-cloud-services/frontend-components-config by default?

coverage.sh
gkarat: Wouldn't it mess up the coverage? Like if you run the script locally, will it update the coverage of the master branch? Or just the branch to which you are checked out locally?

bastilian: It would only attempt to upload, but fail because I won't provide a token. It is also only to verify that the binary is executable and that it finds the correct report.

package.json
bastilian: We actually don't have to build at all. We build it, because we used to push the result afterwards to the `-build` repo. But we don't do that anymore and we could actually remove this step. 

But in general, a build step should always be last, because it should be more like "you shouldn't build something for distributing, if the tests fail".

config/cypress.webpack.config.js
bastilian: Not anymore. It may have been in the past, but it doesn't anymore.

package.json
gkarat: I guess the intention here is to make sure the assets are buildable. Some compilation errors can occur and it has been following rather the logic: "if you can't build the app, you don't need to run tests"

coverage.sh
mkholjuraev: Is this something that the whole team can/should use?

bastilian: @mkholjuraev Yeah, I am thinking it would be good if we all can run this and it will have jest and cypress results separate, to see which tests cover what, as well as a third report with the combined results.

package.json
```diff
@@ -94,26 +94,33 @@
...
     "include": [
       "src/**/*"
     ],
     "exclude": [
-      "src/**/*.test.js"
+      "src/**/*.test.js",
+      "cypress/**/*.js",
+      "config/**/*.js",
+      "src/**/*.test.js",
+      "src/**/*.cy.js",
```
mkholjuraev, line 106: does not this cause some `cy` tests to be excluded?

package.json
```diff
@@ -140,6 +148,7 @@
     "jsdoc": "^4.0.3",
     "natsort": "^2.0.3",
     "npm-run-all": "^4.1.5",
+    "nyc": "^17.0.0",
```
gkarat, line 151: nit-pick: perhaps we could bump it to even newer version, `^17.1.0`

mkholjuraev (approved):

package.json
```diff
@@ -164,11 +174,10 @@
     "server:ctr": "node src/server/generateServerKey.js",
     "start": "fec dev",
     "start:proxy": "PROXY=true fec dev",
-    "verify": "npm-run-all build lint test",
+    "travis:verify": "npm run test:coverage",
```
gkarat, line 177: `verify` is reserved for ci.int pipeline. We should keep this script so that we have basic checks included in ci.int. Without sending coverage because it's covered by Travis.

You can also see that ci.int can't find the script and generates an error: https://ci.int.devshift.net/job/RedHatInsights-compliance-frontend-pr-check/1113/console.

mkholjuraev: Nice catch!

bastilian: Indeed, nice catch! Changed back.

package.json
bastilian: Yeah, that is the purpose. We don't need coverage of any test files, or want them to be included in any coverage calculation.

PR #2093, opened 2024-07-15
feat(Client): JS client setup and useQuery hook

AsToNlele:
Setup for the client and the custom `useQuery` hook.

Added some example hooks we could maybe use, so we don't import the `apiInstance` everytime in the main component but I'm open for changes.

The types for the `apiInstance` endpoints are currently wrong. The endpoints state they take an object, but it still takes params. I'm looking into that and will discuss with platform if needed.

Currently the parameters are set like this with the `useQuery` hook:
```js
const { data, error, loading, refetch } = useQuery(apiInstance.report, [id]);
```

Or like this with the `apiInstance`:
```js
const res = await apiInstance.report('e3d37ce7-ffd6-4b50-8b3d-29b3b180b648');
```

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/Utilities/hooks/useQuery.js
```diff
@@ -0,0 +1,67 @@
...
+    policySystems,
+    tailorings,
+    updatePolicy,
+    report,
+    reports,
+  },
+  instance
+);
+
+export const useQuery = (fn, ...params) => {
```
bastilian, line 30: We should add a test for this hook. The tests will also help as examples on how to use the hook properly.

AsToNlele:
I think I covered all the functionality in the tests, let me know

src/Utilities/hooks/useQuery/useQuery.js
```diff
@@ -0,0 +1,42 @@
...
+      setError(e);
+    } finally {
+      setLoading(false);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (
+      fnRef.current !== fn ||
+      JSON.stringify(paramsRef.current) !== JSON.stringify(params)
```
mkholjuraev, line 31: I believe we can make this simpler by not using a ref to store params. Instead, we can move JSON.stringify(params) into the dependency array. I feel we do not have to have the `fn` in the **ref** nor the dependency array. Let me know if there is any reason that I do not know, happy to learn. 

I would probably introduce AbortController to handle API request race conditions.

src/Utilities/hooks/api/useReport.js
```diff
@@ -0,0 +1,7 @@
+import useQuery, { apiInstance } from '../useQuery';
+
+export const useReport = (id) => {
+  const query = useQuery(apiInstance.report, id);
```
mkholjuraev, line 4: I think we do not have to store the result in another variable. Instead, we can directly return it.
```suggestion
  return useQuery(apiInstance.report, id);
```

src/Utilities/hooks/useQuery/useQuery.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react';
+
+const useQuery = (fn, ...params) => {
```
bastilian, line 3: We will also need to implement a `skip` option, which should prevent any request/state change to happen. Some `useQuery` hooks use this feature of apollo: https://github.com/RedHatInsights/compliance-frontend/blob/80128d099d9b996a8766dbf98b002fbfe3984039/src/SmartComponents/CreatePolicy/EditPolicyProfilesRules.js#L111

We can do this in another PR if we think it is too much for this PR, but I wanted to make a note to not forget about that. :)

bastilian: Another thing. It might be better to not collect the remaining arguments passed as `params`, but have a second options argument that allows to pass in a `params` option, which is passed along. this will also make it easier to add a `skip` option and other things for the `useQuery` hook itself.

AsToNlele: just one question, should a manual `refetch` override the skip or not?

src/Utilities/hooks/useQuery/useQuery.test.js
```diff
@@ -0,0 +1,236 @@
...
+  const apiInstance = {
+    systems: jest.fn(),
+  };
+
+  beforeEach(() => {
+    apiInstance.systems.mockReset();
+  });
+
+  it('keeps loading', () => {
+    const { result } = renderHook(() =>
```
mkholjuraev, line 14: Does this mean, that if the promise returns an empty object, the hook is in a loading state always? Should not it be showing loading state as long as the promise is in pending state?

src/Utilities/hooks/useQuery/useQuery.test.js
```diff
@@ -0,0 +1,236 @@
...
+  });
+
+  it('returns data', async () => {
+    const data = {
+      hello: 'there',
+    };
+    apiInstance.systems.mockResolvedValue(data);
+    const { result } = renderHook(() => useQuery(apiInstance.systems, [{}]));
+
+    expect(result.current).toMatchObject({
```
mkholjuraev, line 51: Do we have to verify the loading state in each test? I think it is enough if we check it only once. What do you think?

src/Utilities/hooks/useQuery/useQuery.test.js
```diff
@@ -0,0 +1,236 @@
...
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        data,
+        error: undefined,
+        loading: false,
+      });
+    });
+  });
+
+  it('calls api endpoint twice on refetch', async () => {
```
mkholjuraev, line 66: Maybe we should call this test as 'calls the API endpoint second time on **refetch**'. I first thought this test checks if the API is called twice when refect is triggered.

src/Utilities/hooks/api/useReports.js
```diff
@@ -0,0 +1,5 @@
+import useQuery, { apiInstance } from '../useQuery';
+
+export const useReports = () => {
+  return useQuery(apiInstance.reports, [{}]);
```
mkholjuraev, line 4: I am a bit confused as to why an empty object is passed inside an array.

src/Utilities/hooks/useQuery/useQuery.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react';
+
+const useQuery = (fn, [...params], skip = false) => {
```
mkholjuraev, line 3: hmm, `[...params]` I believe this is for passing parameters as individual function arguments to the Compliance client. Should we leave some comments to make this visible?

src/Utilities/hooks/useQuery/useQuery.test.js
AsToNlele: The promise doesn't return anything, it has an empty body.
I wanted to show that until you return something, it will be always loading

src/Utilities/hooks/useQuery/useQuery.test.js
AsToNlele: sure!

src/Utilities/hooks/api/useReports.js
AsToNlele: fixed

src/Utilities/hooks/useQuery/useQuery.js
AsToNlele: Yup, added type annotations to the hook and provided some examples as well.
Didn't convert it to typescript since we would need to suppress `any` types in the function parameters.
image

mkholjuraev (approved):
LGTM! I believe we have everything needed and clean now

PR #2103, opened 2024-07-25
fix(useQuery): object for options, refetch memoized

AsToNlele:
After all we need the refs, because otherwise, the `refetch` function changes every render and this code could be stuck in a loop:
```js
  useEffect(() => {
    refetch();
  }, [location, refetch]);
```

Included a test that verifies the reference is still the same.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Utilities/hooks/api/useReport.js
+++ b/src/Utilities/hooks/api/useReport.js
@@ -1,5 +1,5 @@
 import useQuery, { apiInstance } from '../useQuery';
 
 export const useReport = (id) => {
-  return useQuery(apiInstance.report, [id]);
+  return useQuery(apiInstance.report, { params: [id] });
 };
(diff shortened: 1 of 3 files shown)


mkholjuraev:
Would it be still a problem if we wrap the `refetch` function with useCallback?

bastilian:
Can we change the last two arguments of the `useQuery` hook to be an options object, which includes a `skip` option and an `params`, so we can easily add new option when needed without having to change any existing `useQuery` call or end up with a bunch of arguments.

We wanna be able to have something like this:

```js
const query = useQuery(apiInstance.systems, { params: ["param1", "param2"], skip: number > 5 })
```

AsToNlele:
> Would it be still a problem if we wrap the `refetch` function with useCallback?

yes, because the parameters wouldn't change and adding them in the dependency array would change the function

bastilian (approved):
LGTM! Thank you @AsToNlele!

PR #2104, opened 2024-07-25
feat(RHINENG-11636): Implement usePagination hook for async table

adonispuente:
This should be merged after the TableTools PR. Currently this just adds upon that 
feat(RHINENG-11636): Implement usePagination hook for async table 
## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/Frameworks/AsyncTableTools/hooks/useAsyncTableTools/useAsyncTableTools.js
```diff
@@ -13,9 +13,13 @@ import rowsBuilder from './rowsBuilder';
 const useAsyncTableTools = (items, columns, options = {}) => {
   // TODO only for development purposes remove before switching to async tables by default
   console.log('Async Table params:', items, columns, options);
+  const enhancedOptions = {
+    ...options,
+    numberOfItems: items.length,
```
bastilian, line 18: This will not work as the amount of items will always equal the `perPage`. We will pass down the `total` from the API response's `meta` section.

I think a good approach would be to add a `total` prop to the `AsyncTableToolsTable`, which will be passed as the `numberOfItems` option.

https://console.stage.redhat.com/docs/api/compliance/v2#operations-Reports-Reports

src/PresentationalComponents/ComplianceTable/serialisers.js
```diff
@@ -0,0 +1,3 @@
+// TODO correct the serialiser to transform state put into the tablestate to be API consumable
+export const paginationSerialiser = (state) =>
+  `offset=${state.page}&limit=${state.perPage}`;
```
bastilian, line 3: We need to change this to actually calculate the offset and limit. so on page 1, with 10 items, it should return something like `{ offset: 0, limit: 10}` and for page 2 `{ offset: 10, limit: 10}` etc.

PR #2110, opened 2024-07-30
feat(useQuery): serializer option

AsToNlele:
```js
const settings = {
  id: "id",
  title: "name",
  business_objective: "businessObjective.title",
  os_major_version: "osMajorVersion",
  profile_title: "policyType",
  ref_id: "refId",
  "nested.value.hehe": "lets.go.deep.err",
};

const originalResponse = {
  id: "8e06c7fc-29e0-4547-8db0-440604c0545e",
  title: "Odio non ea ut.",
  business_objective: "---",
  os_major_version: 7,
  profile_title: "my title",
  ref_id: "asd-123",
  nested: {
     value: {
        hehe: "hello there"
     }
  }
};
const {data} = useQuery(func, {serializer: settings})
// data = {
//  id: "8e06c7fc-29e0-4547-8db0-440604c0545e",
//  name: "Odio non ea ut.",
//  business_objective: { title: "---" },
//  osMajorVersion: 7,
//  policyTitle: "my title",
//  refId: "asd-123",
//  lets: {
//     go: {
//        deep: {
//           err: "hello there"
//        }
//     }
//  }
//
```

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Utilities/hooks/useQuery/useQuery.js
+++ b/src/Utilities/hooks/useQuery/useQuery.js
@@ -6,7 +6,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
  * @typedef {Object} Options
  * @property {Array} [params] - An array of parameters for the query function.
  * @property {boolean} [skip=false] - A boolean flag to skip the execution of the query function.
- *
+ * @property {Object} [serializer=undefined] - Settings to change the structure of the response
  *
  * @param {Function} fn - Function to execute
  * @param {Options} options - Includes options like params and skip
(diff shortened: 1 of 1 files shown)


src/Utilities/hooks/useQuery/useQuery.js
```diff
@@ -6,7 +6,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
  * @typedef {Object} Options
  * @property {Array} [params] - An array of parameters for the query function.
  * @property {boolean} [skip=false] - A boolean flag to skip the execution of the query function.
- *
+ * @property {Object} [serializer=undefined] - Settings to change the structure of the response
```
bastilian, line 9: What I'm imagining is that, we should rather just be able to pass in a function here. This function should be what we pass in places like the `Reports` page(s) and is basically what you have from line 68 to 98.

It would also separate the concerns better. `useQuery` has a "serialising" feature, but it doesn't need to know how this serialising is working. 

If we wanted to have a `useQuery` hook which already has the function you already wrote and applied it to the state before returning, I'd call it something like `useQueryWithAttributeSerialiser`, or even `useQueryWithComlianceSerialiser` since the serialising of the attributes would be specific to Compliance.

PR #2113, opened 2024-07-31
docs: Extend using AsyncTableToolsTable tutorial

bastilian:
This adds more documentation to several components, hooks and functions and extends the tutorial on using the `AsyncTableToolsTable` component.

It also changes the `useTableState` hook to work without a context provided. This was added to allow the tutorial to go from a simple example to adding everything one by one.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
+++ b/src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
@@ -13,10 +13,10 @@ import useAsyncTableTools from '../../hooks/useAsyncTableTools';
 /**
  *  This component is a wrapper around the Patternfly Table component(s), the FEC PrimaryToolbar and combines them with the `useAsyncTableTools` hook
  *
- *  @param props
- *  @param {Array} props.items An array or (async) function that returns an array of items to render or an async function to call with the tableState and serialised table state
- *  @param {Array} props.columns An array of column objects to render items with
- *  @param {Array} [props.filters] an array of filters
+ *  @param {object} props Component Props
+ *  @param {Array}  props.items An array or (async) function that returns an array of items to render or an async function to call with the tableState and serialised table state
+ *  @param {Array}  props.columns An array of column objects to render items with
+ *  @param {Array}  [props.filters] an array of filters
  *  @param {object} [props.options] An object of options that will be passed along to the `useAsyncTableTools` hook
  *  @param {object} [props.toolbarProps] Props to be passed on the `PrimaryToolbar` component
  *  @param {object} [props.tableHeaderProps] Props to be passed on the TableHeader component
@@ -24,7 +24,7 @@ import useAsyncTableTools from '../../hooks/useAsyncTableTools';
  *  @param {object} [props.tableToolbarProps] Props to be passed on the TableToolbar (bottom toolbar) component
  *  @param {object} [props.paginationProps] Props to be passed on the Pagination component
  *
- *  @returns {React.ReactElement}
+ *  @returns {React.ReactElement} Returns a `PrimaryToolbar` component, a Patternfly (v4) `Table` component and a `TableToolbarComponent` wrapped together
  *
  *  @tutorial using-async-table-tools
  *
@@ -37,6 +37,8 @@ const AsyncTableToolsTable = ({
   columns,
   filters,
   options,
+  // TODO I'm not sure if we need this level of customisation.
+  // It might actually hurt in the long run. Consider removing until we really have the case where we need this
   toolbarProps: toolbarPropsProp,
   tableHeaderProps,
   tableBodyProps,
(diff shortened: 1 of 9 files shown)


src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
```diff
@@ -13,18 +13,18 @@ import useAsyncTableTools from '../../hooks/useAsyncTableTools';
 /**
  *  This component is a wrapper around the Patternfly Table component(s), the FEC PrimaryToolbar and combines them with the `useAsyncTableTools` hook
  *
- *  @param props
- *  @param {Array} props.items An array or (async) function that returns an array of items to render or an async function to call with the tableState and serialised table state
- *  @param {Array} props.columns An array of column objects to render items with
- *  @param {Array} [props.filters] an array of filters
+ *  @param {object} props Component Props
+ *  @param {Array}  props.items An array or (async) function that returns an array of items to render or an async function to call with the tableState and serialised table state
```
mkholjuraev, line 17: Is not it better to have separate props for the items array and the async function that returns items?

src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
```diff
@@ -13,18 +13,18 @@ import useAsyncTableTools from '../../hooks/useAsyncTableTools';
...
+ *  @param {Array}  [props.filters] an array of filters
  *  @param {object} [props.options] An object of options that will be passed along to the `useAsyncTableTools` hook
  *  @param {object} [props.toolbarProps] Props to be passed on the `PrimaryToolbar` component
  *  @param {object} [props.tableHeaderProps] Props to be passed on the TableHeader component
  *  @param {object} [props.tableBodyProps] Props to be passed on the TableBody component
  *  @param {object} [props.tableToolbarProps] Props to be passed on the TableToolbar (bottom toolbar) component
  *  @param {object} [props.paginationProps] Props to be passed on the Pagination component
  *
- *  @returns {React.ReactElement}
+ *  @returns {React.ReactElement} Returns a `PrimaryToolbar` component, a Patternfly (v4) `Table` component and a `TableToolbarComponent` wrapped together
```
mkholjuraev, line 27: Maybe an irrelevant question, how difficult it will be to make this 'Framework' compatible with PF5? I am afraid we may get rather negative feedback if we again say we need to refactor useTableTools (useAsyncTableTools).

src/Frameworks/AsyncTableTools/docs/using-async-table-tools.md
```diff
@@ -1,19 +1,351 @@
...
+
+const filters = [
+  {
+    type: 'text',
+    label: 'Name',
+  }
+];
+
+const filtersSerialiser = (filterState) => ({
+    filter: Object.entries(filterState).reduce((filterStringParts, [filterId, filterValue]) =>
```
mkholjuraev, line 228: Can we provide a specific serializer for some filter that has distinct syntax? for example `tags` filter

bastilian: Sort of, but that is more a concern of the app filterSerialiser. The filter hook PR has this implemented in the filterSerialiser for Compliance: https://github.com/RedHatInsights/compliance-frontend/pull/2105/files#diff-3f7f8a4fe533d214a3b02f95a52a1d62a615b6c9601232461dadcbc83c40a987R56

src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
bastilian: It would probably be very easy. We would only need to change/replace this component to use the PF v5 Table. We can probably use most returns from the hooks as props still. And we can just take the `rows` and `cells` the hooks returns and turn them into table header and body rows.

src/Frameworks/AsyncTableTools/docs/using-async-table-tools.md
mkholjuraev: Nice!

src/Frameworks/AsyncTableTools/components/AsyncTableToolsTable/AsyncTableToolsTable.js
bastilian: Hm. I don't feel it would be better or worse. Just, if there were two props that are for the same purpose we would then at least have to decide which one would be used if both were present.

mkholjuraev (approved):
LGTM! Thank you for further improving the doc. I believe the PR is good to go.

PR #2116, opened 2024-08-01
feat(RHINENG-11639): implement useBulkSelect hook

mkholjuraev:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/ReportsTable/ReportsTable.js
+++ b/src/PresentationalComponents/ReportsTable/ReportsTable.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import propTypes from 'prop-types';
 import { COMPLIANCE_TABLE_DEFAULTS } from '@/constants';
 import { emptyRows } from '../../Utilities/hooks/useTableTools/Components/NoResultsTable';
@@ -20,6 +20,15 @@ const ReportsTable = ({ profiles }) => {
   const operatingSystems = uniq(
     profiles.map(({ osMajorVersion }) => osMajorVersion).filter((i) => !!i)
   );
+
+  useEffect(() => {
+    localStorage.setItem('insights:compliance:asynctables', 'true');
+  }, []);
+
+  const selectionTransformer = (row, selectedIds) => ({
+    ...row,
+    selected: selectedIds.includes(row.itemId),
+  });
 
   return (
     <TableToolsTable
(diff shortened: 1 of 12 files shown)


mkholjuraev:
@bastilian There is yet some minor cleanup and tests needed. But, I wanted to get initial feedback.

src/PresentationalComponents/ReportsTable/ReportsTable.js
```diff
@@ -21,6 +21,15 @@ const ReportsTable = ({ profiles }) => {
     profiles.map(({ osMajorVersion }) => osMajorVersion).filter((i) => !!i)
   );
 
+  useEffect(() => {
+    localStorage.setItem('insights:compliance:asynctables', 'true');
+  }, []);
+
+  const selectionTransformer = (row, selectedIds) => ({
```
bastilian, line 28: We can put this directly into the `useBulkSelect` hook here and pass it on as a "transformer" to the `rowsBuilder` function.

The current non async BulkSelect hook has something like this implemented: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/Utilities/hooks/useTableTools/useBulkSelect.js#L157

bastilian:
@mkholjuraev Awesome! Thank you! Looks already great to me! Only the select transformer can be moved, so we don't have to add this separate for each table!

src/PresentationalComponents/ReportsTable/ReportsTable.js
mkholjuraev: You are right, with this we have to pass the transformer on each page. What I was thinking was that we could pass any row transformers (e.g. selection and expansion) and they would be applied [sequentially](https://github.com/RedHatInsights/compliance-frontend/pull/2116/files#diff-1b4bc8a5c684410ae539b9e88458f86e27810151ee22e4bf3ca2326dd73be23cR16).

Also, this specific function is not needed. Because there is already one that is [exported](https://github.com/RedHatInsights/compliance-frontend/pull/2116/files/e72f43b10a4c455db17fb4e69988168453019322#diff-94f7f118c9c0682dd21f34502253c426150667f06f04549f034aaea3f0788725R37).

mkholjuraev:
@bastilian @RedHatInsights/team-interact  I have updated the PR and it is ready for another review.

src/Utilities/hooks/useFetchAllData.js
bastilian, line 1: This can probably removed or?

bastilian (approved):
Nice! LGTM! Thank you @mkholjuraev!

PR #2123, opened 2024-08-05
feat(RHINENG-11641): implement async useExpandable hook

mkholjuraev:
Implements async useExpandable hook. There is a known [issue](https://redhat-internal.slack.com/archives/C05494LH96F/p1722853892200309) related to useSelectionManager hook, I am working on a fix.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
+++ b/src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
@@ -18,6 +18,7 @@ import { itemDetailsRow } from './helpers';
  */
 
 const useExpandable = (options) => {
+  const enableExpandingRow = !!options?.detailsComponent;
   const { selection: openItems, toggle } = useSelectionManager([]);
 
   const onCollapse = (_event, _index, _isOpen, { itemId }) => toggle(itemId);
@@ -37,13 +38,15 @@ const useExpandable = (options) => {
     [openItems]
   );
 
-  return {
-    openItem,
-    openItems,
-    tableProps: {
-      onCollapse,
-    },
-  };
+  return enableExpandingRow
+    ? {
+        openItem,
+        openItems,
+        tableProps: {
+          onCollapse,
+        },
+      }
+    : {};
 };
 
 export default useExpandable;
--- a/src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
+++ b/src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
@@ -1,0 +1,49 @@
+import { useCallback } from 'react';
+import useSelectionManager from '../useSelectionManager';
+import { itemDetailsRow } from './helpers';
+
+/**
+ *  @typedef {object} AsyncTableProps
+ *  @property {object} toolbarProps Object containing PrimaryToolbar props
+ *  @property {object} tableProps Object containing Patternfly (deprecated) Table props
+ */
+
+/**
+ * Provides props for a Patternfly table to manage expandable items/rows.
+ *  @param {object} [options] an oblect containing detailComponent to render on expand
+ *  @returns {AsyncTableProps} An object of props meant to be used in the {@link AsyncTableToolsTable}
+ *
+ *  @category AsyncTableTools
+ *  @subcategory Hooks
+ */
+
+const useExpandable = (options) => {
+  const { selection: openItems, toggle } = useSelectionManager([]);
+
+  const onCollapse = (_event, _index, _isOpen, { itemId }) => toggle(itemId);
+
+  const openItem = useCallback(
+    (row, _selectedIds, index) => {
+      const expandedRow = [
+        {
+          ...row,
+          isOpen: (openItems || []).includes(row.itemId),
+        },
+        itemDetailsRow(row, index, options),
+      ];
+
+      return expandedRow;
+    },
+    [openItems]
+  );
+
+  return {
+    openItem,
+    openItems,
+    tableProps: {
+      onCollapse,
+    },
+  };
+};
+
+export default useExpandable;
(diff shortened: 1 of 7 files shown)


src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
```diff
@@ -0,0 +1,49 @@
...
+  const onCollapse = (_event, _index, _isOpen, { itemId }) => toggle(itemId);
+
+  const openItem = useCallback(
+    (row, _selectedIds, index) => {
+      const expandedRow = [
+        {
+          ...row,
+          isOpen: (openItems || []).includes(row.itemId),
+        },
+        itemDetailsRow(row, index, options),
```
gkarat, line 32: If options.detailsComponent is undefined, would it mean that we keep the value "false" here? Is it expected?

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
```diff
@@ -0,0 +1,49 @@
...
+ *  @returns {AsyncTableProps} An object of props meant to be used in the {@link AsyncTableToolsTable}
+ *
+ *  @category AsyncTableTools
+ *  @subcategory Hooks
+ */
+
+const useExpandable = (options) => {
+  const { selection: openItems, toggle } = useSelectionManager([]);
+
+  const onCollapse = (_event, _index, _isOpen, { itemId }) => toggle(itemId);
```
gkarat, line 23: Do we want to support collapse/expand also by the row index? Or it's strictly limited to the id of the item?

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.test.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(result.current).toEqual({
+      openItem: expect.any(Function),
+      openItems: [],
+      tableProps: {
+        onCollapse: expect.any(Function),
+      },
+    });
+  });
+
+  it('should expand & shrink  a row', () => {
```
gkarat, line 26:
```suggestion
  it('should expand & collapse a row', () => {
```

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.test.js
```diff
@@ -0,0 +1,47 @@
...
+    //expand
+    act(() => {
+      result.current.tableProps.onCollapse(null, 0, 'row title', {
+        itemId: 'test-id',
+      });
+    });
+
+    expect(result.current.openItems).toEqual(['test-id']);
+
+    //shrink
```
gkarat, line 38:
```suggestion
    // collapse
```

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.test.js
```diff
@@ -0,0 +1,47 @@
...
+      },
+    });
+  });
+
+  it('should expand & shrink  a row', () => {
+    const { result } = renderHook(() => useExpandable(defaultOptions));
+
+    //expand
+    act(() => {
+      result.current.tableProps.onCollapse(null, 0, 'row title', {
```
gkarat, line 31: nit-pick, the third argument is boolean:

```suggestion
      result.current.tableProps.onCollapse(null, 0, true, {
```

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.test.js
```diff
@@ -0,0 +1,47 @@
...
+        itemId: 'test-id',
+      });
+    });
+
+    expect(result.current.openItems).toEqual(['test-id']);
+
+    //shrink
+    act(() => {
+      result.current.tableProps.onCollapse(null, 0, 'row title', {
```
gkarat, line 40: same here
```suggestion
      result.current.tableProps.onCollapse(null, 0, false, {
```

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.test.js
gkarat, line 1: perhaps we want write a few more tests for `openItem`?

mkholjuraev: added

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
johnsonm325: @gkarat means that if `itemDetailsRow` returns false, then is the intention to return:
```
[
  {
    ...row,
    isOpen: (openItems || []).includes(row.itemId),
  },
  false,
]
```

mkholjuraev: Yes, that would result in false to be returned. `options.detailsComponent` is a required parameter to be passed.

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
```diff
@@ -0,0 +1,49 @@
...
+        },
+        itemDetailsRow(row, index, options),
+      ];
+
+      return expandedRow;
+    },
+    [openItems]
+  );
+
+  return {
```
bastilian, line 40: It might make sense to only return something here if the `options.detailsComponent` is set. Other hooks have something similar with `enable....` `const`s. 

This would probably also prevent the case above where when `itemDetailsRow` would be called it would return `false`.

src/Frameworks/AsyncTableTools/hooks/useExpandable/useExpandable.js
gkarat: Yes, I see it's required, but is it correct passing a boolean value instead of React component? @mkholjuraev

mkholjuraev: @gkarat Yup, I have made the necessary change to prevent this case.

PR #2129, opened 2024-08-12
fix(RHINENG-2136): Fix spacing in Create policy wizard review step

leSamo:
https://issues.redhat.com/browse/RHINENG-2136

To test:
1. Open the wizard to create a new SCAP policy
2. Go to the Review step
3. Observe the spacing between the rows
  - space below policy name (C2S for Red Hat Enterprise Linux 6) should be the same as spacing between (Policy type and Compliance threshold) and (Compliance threshold and Systems)
  - space above policy name (C2S for Red Hat Enterprise Linux 6) should be the larger than spacing below the policy name

## Before
Screenshot from 2024-08-12 23-33-28

## After
Screenshot from 2024-08-12 23-33-24


bastilian (approved):
LGTM! Thank you @leSamo!

PR #2130, opened 2024-08-14
feat(RHINENG-10391): migrate Reports page to REST

mkholjuraev:
This migrates the Reports page to use REST API with a flag. It also includes serialiser implementation using lodash set function.

To test:
1. Visit reports page with compliance-api-v2 flag disabled
2. Observe that there are no changes in the behavior and grapgQL is used to fetch data
3. Enable the flag
4. Observe that now REST api is used 
5. Observe that table renders all data
6. Observe that filtering, pagination and other functionalities yet disabled

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


package.json
```diff
@@ -139,6 +139,8 @@
     "jsdoc": "^4.0.3",
     "natsort": "^2.0.3",
     "npm-run-all": "^4.1.5",
+    "react-intl": "^6.6.8",
+    "redux-mock-store": "^1.5.4",
```
bastilian, line 143: Why is the mock-store needed now?

mkholjuraev: Is it because of the new TestWrapper from fee-components package. It requires these packages to be installed as a dev-dependency. Now, I think if mocking store and int provider is needed at all from the TestWrapper or should we make that component configurable so that apps can choose what wrappers to be added.

bastilian: Yeah, it would make sense to only enable these features when needed.

It would also be better to keep the dependencies of the TestWrapper also as dependencies of the fec/utils package rather than adding them as an app dependency.

mkholjuraev: Good, in that case let's keep the dips in fec/utils itself and leave redux and into providers as is. Making them configurable feels overkill.

mkholjuraev: PR in few: https://github.com/RedHatInsights/frontend-components/pull/2052

src/PresentationalComponents/ReportsTable/ReportsTable.js
```diff
@@ -52,7 +52,7 @@ const ReportsTable = ({ profiles }) => {
 };
 
 ReportsTable.propTypes = {
-  profiles: propTypes.array,
+  reports: propTypes.any,
```
gkarat, line 55:
```suggestion
  reports: propTypes.array.isRequired,
```

src/SmartComponents/Reports/Reports.js
```diff
@@ -75,35 +73,41 @@ export const Reports = () => {
...
+  return <ReportsBase {...{ data: profiles, error, loading, refetch }} />;
+};
+
+const ReportsWithRest = () => {
+  const params = {
+    params: ['', 100],
+  };
+  const location = useLocation();
+
+  let { data: { data } = {}, error, loading, refetch } = useReports(params);
```
gkarat, line 87:
```suggestion
  const location = useLocation();

  let { data: { data } = {}, error, loading, refetch } = useReports();
```

since from what I see in the implementation of useReports: it doesn't support any params.

src/SmartComponents/Reports/Reports.js
```diff
@@ -75,35 +73,41 @@ export const Reports = () => {
...
+
+  useEffect(() => {
+    refetch();
+  }, [location, refetch]);
+
+  if (data) {
+    data = dataSerialiser(data, dataMap);
+    error = undefined;
+    loading = undefined;
```
gkarat, line 96: I am not sure why it's necessary here? The error should be null or undefined if the request's succeeded. The same for loading: it should indicate with only boolean value if the request is complete.

src/SmartComponents/Reports/Reports.js
```diff
@@ -57,9 +28,36 @@ const ReportsHeader = () => (
   </PageHeader>
 );
 
-export const Reports = () => {
+export const ReportsBase = ({ data, loading, error }) => {
+  const showView = data && data.length > 0;
```
gkarat, line 32:
```suggestion
  const showView = data && data.length > 0;
```

is it correct to call `.length` on object here?

src/SmartComponents/Reports/Reports.js
```diff
@@ -57,9 +28,36 @@ const ReportsHeader = () => (
...
+};
+
+ReportsBase.propTypes = {
+  data: PropTypes.object,
+  error: PropTypes.string,
+  loading: PropTypes.bool,
+};
+
+//depricated component
+const ReportsWithGrahpQL = () => {
```
gkarat, line 59:
```suggestion
// deprecated component
const ReportsWithGraphQL = () => {
```

src/SmartComponents/Reports/Reports.js
bastilian: The StateView will show parts for keys that have something defined, so a `null` or a boolean would make them defined and cause the StateView to render. 

There are reasons for why the StateView behaves like this, but I can't remember at the moment. We could think about changing this behaviour at some point.

src/SmartComponents/Reports/Reports.js
bastilian: JSDoc provides a way to properly mark something as [deprecated](https://jsdoc.app/tags-deprecated), this will also pop up in any TypeScript thingies.

src/SmartComponents/Reports/Reports.js
mkholjuraev: the type of `data` is an array. Fixed in the propTypes.

bastilian:
@mkholjuraev Are the `package-lock.json` changes needed?

gkarat (commented):
Code looks good! Though found two suspicious things: 1) we do too many request in my opinion to fetch the reports list thru REST API: there are about 5 identical requests made for some reason, 2) the report description/grey text doesn't have proper key mapping and results in the default "report" text.

Screenshot 2024-08-27 at 11 09 36

Screenshot 2024-08-27 at 11 10 09

mkholjuraev:
@gkarat those requests should be gone now. They were due to the hook to refetch with location object as dependency. Also the report description is also fixed.

@bastilian the changes in package-lock were unexpected. They were due to the difference in npm version. I have fixed this also.

gkarat (approved):
Awesome, thank you @mkholjuraev

PR #2142, opened 2024-08-22
feat(APIV2): Delete Policy modal

AsToNlele:
[RHINENG-11723](https://issues.redhat.com/browse/RHINENG-11723)

Policy deletion using both APIs. Both should work.

How to test:
1. Test the V2 version
2. Enable the `compliance-api-v2` feature flag or manually hardcode `apiV2Enabled` to true in `CompliancePolicies` and `DeletePolicy`
3. Test the policy gets deleted and table is updated
4. Repeat for GraphQL version

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/DeletePolicy/DeletePolicy.js
+++ b/src/SmartComponents/DeletePolicy/DeletePolicy.js
@@ -8,6 +8,7 @@ import {
   ModalVariant,
   Text,
   Spinner,
+  Bullseye,
 } from '@patternfly/react-core';
 import useNavigate from '@redhat-cloud-services/frontend-components-utilities/useInsightsNavigate';
 import { DELETE_PROFILE } from 'Mutations';
@@ -19,43 +20,18 @@ import {
 } from 'PresentationalComponents';
 import { dispatchAction } from 'Utilities/Dispatcher';
 import usePolicyQuery from 'Utilities/hooks/usePolicyQuery';
+import usePolicyQuery2 from 'Utilities/hooks/usePolicyQuery/usePolicyQuery2';
+import useAPIV2FeatureFlag from '../../Utilities/hooks/useAPIV2FeatureFlag';
+import { apiInstance } from '../../Utilities/hooks/useQuery';
+import dataSerialiser from '../../Utilities/dataSerialiser';
+import { dataMap } from './constants';
 
-const DeletePolicy = () => {
+const DeletePolicyBase = ({ query, deletePolicy, onClose }) => {
   const [deleteEnabled, setDeleteEnabled] = useState(false);
-  const navigate = useNavigate();
-  const { policy_id: policyId } = useParams();
-  const { data, error, loading } = usePolicyQuery({
-    policyId,
-    minimal: true,
-  });
+  const { data, error, loading } = query;
   const {
     profile: { name, id },
   } = data || { profile: {} };
-  const onClose = () => {
-    navigate('/scappolicies');
-  };
-
-  const [deletePolicy] = useMutation(DELETE_PROFILE, {
-    onCompleted: () => {
-      dispatchAction(
-        addNotification({
-          variant: 'success',
-          title: `Deleted "${name}" and its associated reports`,
-        })
-      );
-      onClose();
-    },
-    onError: (error) => {
-      dispatchAction(
-        addNotification({
-          variant: 'danger',
-          title: 'Error removing policy',
-          description: error.message,
-        })
-      );
-      onClose();
-    },
-  });
 
   return (
     <ComplianceModal
@@ -72,7 +48,7 @@ const DeletePolicy = () => {
           aria-label="delete"
           isDisabled={!deleteEnabled}
           variant="danger"
-          onClick={() => deletePolicy({ variables: { input: { id } } })}
+          onClick={() => deletePolicy(id)}
         >
           Delete policy and associated reports
         </Button>,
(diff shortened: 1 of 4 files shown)


src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -107,8 +82,119 @@ const DeletePolicy = () => {
...
+  query: propTypes.shape({
+    loading: propTypes.bool,
+    error: propTypes.object,
+    data: propTypes.object,
+  }),
+  onClose: propTypes.func,
+  deletePolicy: propTypes.func,
+};
+
+const DeletePolicyV2 = ({ policyId, onClose }) => {
```
gkarat, line 95: nit-pick: Would be great if we could stay also consistent in our naming. Some PRs I saw before used Component**Rest** suffix instead of V2

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -107,8 +82,119 @@ const DeletePolicy = () => {
...
+};
+
+DeletePolicyGraphQL.propTypes = {
+  policyId: propTypes.string,
+  onClose: propTypes.func,
+};
+
+const DeletePolicyWrapper = () => {
+  const { policy_id: policyId } = useParams();
+  const apiV2Enabled = useAPIV2FeatureFlag();
```
gkarat, line 189: Feature flags are fetched at runtime. That means there is a snap time frame where the value would be false or undefined. Could we make sure that we also distinguish the "loading" state and render some simple spinner before deciding on which component we should use?

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
```diff
@@ -8,6 +8,18 @@ import TestWrapper from '@/Utilities/TestWrapper';
...
+
+jest.mock('../../Utilities/hooks/useQuery', () => ({
+  ...jest.requireActual('../../Utilities/hooks/useQuery'),
+  __esModule: true,
+  default: () => ({
+    data: { data: { title: 'Test Policy 1', id: 'teste_polict_ID' } },
+    error: undefined,
+    loading: undefined,
+  }),
+}));
```
gkarat, line 21: I think we can shorten this by using `mockImplementation`/`mockReturnValue` right inside `it`, like this https://github.com/RedHatInsights/compliance-frontend/blob/54bdc63913528585555b5efed1ca822d6712b3d3/src/SmartComponents/CompliancePolicies/CompliancePolicies.test.js#L32

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
```diff
@@ -42,6 +56,7 @@ describe('DeletePolicy', () => {
       },
     }));
     dispatchAction.mockImplementation(() => {});
+    useAPIV2FeatureFlag.mockImplementation(() => false);
```
gkarat, line 59:
```suggestion
    useAPIV2FeatureFlag.mockReturnValue(false);
```

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
```diff
@@ -59,3 +74,34 @@ describe('DeletePolicy', () => {
...
+  });
+
+  it('expect to render a modal to delete a policy', () => {
+    render(
+      <TestWrapper>
+        <DeletePolicy />
+      </TestWrapper>
+    );
+
+    screen.debug();
```
gkarat, line 98:
```suggestion
```

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
```diff
@@ -59,3 +74,34 @@ describe('DeletePolicy', () => {
...
+
+    expect(
+      screen.getByText(
+        'I understand this will delete the policy and all associated reports'
+      )
+    ).toBeInTheDocument();
+    expect(screen.getByText('Test Policy 1')).toBeInTheDocument();
+  });
+});
```
gkarat, line 107: What about adding an async test that would 1) render the modal, 2) click on the confirmation checkbox, 3) click on the Delete policy button and expect a callback (mocked) to be called?

gkarat (commented):
Some suggestions, but overall looks good. Tested with the flag set to true, all the GET and DELETE requests are sent as expected 👍🏼

adonispuente:
Functionality wise I wasnt able to find an issues and everything is working as intended! The ci is failing with a resource error that I think is unrelated, gonna slap a retest on it

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
AsToNlele: good idea

src/SmartComponents/DeletePolicy/DeletePolicy.test.js
AsToNlele: it worked, but the I needed to mock a named import from here and had to revert to this solution

src/SmartComponents/DeletePolicy/DeletePolicy.js
AsToNlele: Used a spinner that matches the previous ones so the user sees the same spinner while loading the app, made a test for it as well.

AsToNlele:
All the suggestions are added in a new commit, thanks for reviewing :)

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -107,8 +83,127 @@ const DeletePolicy = () => {
...
+    policyId,
+  });
+
+  const data = query.data?.data
+    ? dataSerialiser(query.data.data, dataMap)
+    : undefined;
+
+  const deletePolicy = (id) => {
+    try {
+      apiInstance.deletePolicy(id).then(() => {
```
mkholjuraev, line 107: as you have already wrapper the API call with try&catch, we do not have to use promise chaining `.then`. Actually, because of this, if APiI request fails error notification is not popping up.

src/SmartComponents/DeletePolicy/DeletePolicy.js
```diff
@@ -107,8 +83,127 @@ const DeletePolicy = () => {
...
+const DeletePolicyRest = ({ policyId, onClose }) => {
+  const query = usePolicyQuery2({
+    policyId,
+  });
+
+  const data = query.data?.data
+    ? dataSerialiser(query.data.data, dataMap)
+    : undefined;
+
+  const deletePolicy = (id) => {
```
mkholjuraev, line 105: This code is tested with both API requests, failing and passing.
```suggestion
 const deletePolicy = async (id) => {
    try {
      await apiInstance.deletePolicy(id);
      dispatchAction(
        addNotification({
          variant: 'success',
          title: `Deleted "${query.data?.data?.title}" and its associated reports`,
        })
      );
      onClose();
    } catch (e) {
      dispatchAction(
        addNotification({
          variant: 'danger',
          title: 'Error removing policy',
          description: e?.message,
        })
      );
      onClose();
    }
  };
```

mkholjuraev (approved):
LGTM! only one issue and then it is ready to be merged.

PR #2148, opened 2024-09-03
feat(RHINENG-11719): migrate report download modal to REST api

mkholjuraev:

Migrates the report download modal to REST api. The hooks used to fetch report data are also migrated within this effort.  

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/ReportDownload/ReportDownload.js
+++ b/src/SmartComponents/ReportDownload/ReportDownload.js
@@ -15,10 +15,10 @@ import { GET_PROFILE } from './constants';
 import ExportPDFForm from './Components/ExportPDFForm';
 import usePDFExport from './hooks/usePDFExport';
 import useExportSettings from './hooks/useExportSettings';
-import useAPIV2FeatureFlag from '../../Utilities/hooks/useAPIV2FeatureFlag';
-import { useReport } from '../../Utilities/hooks/api/useReport';
-import dataSerialiser from '../../Utilities/dataSerialiser';
-import { reportDataMap } from '../../constants';
+import { useReport } from '@/Utilities/hooks/api/useReport';
+import dataSerialiser from '@/Utilities/dataSerialiser';
+import { reportDataMap } from '@/constants';
+import GatedComponents from '@/PresentationalComponents/GatedComponents';
 
 // Provides that export settings modal accessible in the report details
 export const ReportDownloadBase = ({ report, loading, error, data }) => {
@@ -98,7 +98,7 @@ ReportDownloadBase.propTypes = {
   data: propTypes.object.isRequired,
 };
 
-export const ReportsWithGraphQL = () => {
+export const ReportDownloadGraphQL = () => {
   const { report_id: reportId } = useParams();
 
   const { data, loading, error } = useQuery(GET_PROFILE, {
@@ -128,10 +128,11 @@ export const ReportDownloadRest = () => {
   );
 };
 
-export const ReportDownload = () => {
-  const isRestApiEnabled = useAPIV2FeatureFlag();
-
-  return isRestApiEnabled ? <ReportDownloadRest /> : <ReportsWithGraphQL />;
-};
+export const ReportDownload = () => (
+  <GatedComponents
+    RestComponent={ReportDownloadRest}
+    GraphQLComponent={ReportDownloadGraphQL}
+  />
+);
 
 export default ReportDownload;
--- a/src/SmartComponents/ReportDownload/ReportDownload.js
+++ b/src/SmartComponents/ReportDownload/ReportDownload.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import propTypes from 'prop-types';
 import { Button, Spinner } from '@patternfly/react-core';
 import { useParams } from 'react-router-dom';
 import { useQuery } from '@apollo/client';
@@ -14,22 +15,22 @@ import { GET_PROFILE } from './constants';
 import ExportPDFForm from './Components/ExportPDFForm';
 import usePDFExport from './hooks/usePDFExport';
 import useExportSettings from './hooks/useExportSettings';
+import useAPIV2FeatureFlag from '../../Utilities/hooks/useAPIV2FeatureFlag';
+import { useReport } from '../../Utilities/hooks/api/useReport';
+import dataSerialiser from '../../Utilities/dataSerialiser';
+import { reportDataMap } from '../../constants';
 
 // Provides that export settings modal accessible in the report details
-export const ReportDownload = () => {
-  const { report_id: policyId } = useParams();
+export const ReportDownloadBase = ({ report, loading, error, data }) => {
+  console.log(report, 'debug policy');
   const navigate = useNavigate();
-  const { data, loading, error } = useQuery(GET_PROFILE, {
-    variables: { policyId },
-  });
-  const policy = data?.profile;
   const {
     exportSettings,
     setExportSetting,
     isValid: settingsValid,
   } = useExportSettings();
 
-  const exportPDF = usePDFExport(exportSettings, policy);
+  const exportPDF = usePDFExport(exportSettings, report);
   const exportFileName = `compliance-report--${
     new Date().toISOString().split('T')[0]
   }`;
@@ -48,7 +49,7 @@ export const ReportDownload = () => {
       key="export"
       label={buttonLabel}
       reportName={`Compliance:`}
-      type={policy && policy.name}
+      type={report && report.name}
       fileName={exportFileName}
       asyncFunction={exportPDF}
       buttonProps={buttonProps}
@@ -82,11 +83,52 @@ export const ReportDownload = () => {
           <Spinner />
         </StateViewPart>
         <StateViewPart stateKey="data">
-          <ExportPDFForm {...{ policy, setExportSetting, exportSettings }} />
+          <ExportPDFForm
+            {...{ policy: report, setExportSetting, exportSettings }}
+          />
         </StateViewPart>
       </StateViewWithError>
     </ComplianceModal>
   );
+};
+
+ReportDownloadBase.propTypes = {
+  report: propTypes.object.isRequired,
+  loading: propTypes.bool,
+  error: propTypes.object,
+  data: propTypes.object.isRequired,
+};
+
+export const ReportsWithGraphQL = () => {
+  const { report_id: reportId } = useParams();
+
+  const { data, loading, error } = useQuery(GET_PROFILE, {
+    variables: { policyId: reportId },
+  });
+
+  return (
+    <ReportDownloadBase report={data?.profile} {...{ loading, error, data }} />
+  );
+};
+
+export const ReportDownloadRest = () => {
+  const { report_id: reportId } = useParams();
+
+  const { data: { data } = {}, loading, error } = useReport(reportId);
+  const serialisedData = dataSerialiser(data, reportDataMap);
+  return (
+    <ReportDownloadBase
+      report={serialisedData}
+      reportId={reportId}
+      {...{ loading, error, data: serialisedData }}
+    />
+  );
+};
+
+export const ReportDownload = () => {
+  const isRestApiEnabled = true;
+
+  return isRestApiEnabled ? <ReportDownloadRest /> : <ReportsWithGraphQL />;
 };
 
 export default ReportDownload;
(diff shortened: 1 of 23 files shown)


src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
```diff
@@ -1,93 +1,69 @@
...
-        [...new Array(pages)].map(
-          (_, pageIdx) => () => fetchFunction(batchSize, pageIdx + 1, filter)
-        )
-      );
-
-      return results;
-    },
-  };
+    return prepareForExportGraphQL(exportSettings, systems, rules);
+  }, [JSON.stringify(report), JSON.stringify(exportSettings)]);
```
bastilian, line 20: These `JSON.stringify` look suspicious. Why are they needed?

src/SmartComponents/ReportDownload/ReportDownload.js
```diff
@@ -82,11 +82,52 @@ export const ReportDownload = () => {
...
+  return (
+    <ReportDownloadBase report={data?.profile} {...{ loading, error, data }} />
+  );
+};
+
+export const ReportDownloadRest = () => {
+  const { report_id: reportId } = useParams();
+
+  const { data: { data } = {}, loading, error } = useReport(reportId);
+  const serialisedData = dataSerialiser(data, reportDataMap);
```
bastilian, line 117: This should probably be memoised as it might cause unwanted rerenders.

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
mkholjuraev: They are just to make sure the callback is not recreated when report and exportSettings objects are reinitialised in re-renders, but do not change.

src/SmartComponents/ReportDownload/ReportDownload.js
mkholjuraev: Nice idea!

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
bastilian: @mkholjuraev Let's try and figure out what in these objects changes and try to only make that not change and not use `JSON.stringify` right away.

mkholjuraev: Does not an object gets reinitialised on every re-render and that cause the hook to be retriggered because of access by reference?

mkholjuraev: I do not see anything that gets changes in both exportSettings and reports as both of them do not contain function properties.

src/SmartComponents/ReportDownload/hooks/apiQueryHooks.js
```diff
@@ -0,0 +1,177 @@
...
+        filter: `policy_id = ${policyId}`,
+        policyId,
+      },
+    });
+
+  return async () =>
+    (await fetchFunction()).data.profiles?.edges.flatMap(
+      (edge) => edge.node.topFailedRules
+    );
+};
```
bastilian, line 107: All of the functions defined and returned in here should be wrapped in `useCallback`s

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
```diff
@@ -1,93 +1,69 @@
...
-        [...new Array(pages)].map(
-          (_, pageIdx) => () => fetchFunction(batchSize, pageIdx + 1, filter)
-        )
-      );
-
-      return results;
-    },
-  };
+    return prepareForExportGraphQL(exportSettings, systems, rules);
+  }, [report, exportSettings]);
```
bastilian, line 20: This is missing dependencies.

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
```diff
@@ -1,93 +1,69 @@
...
-      (edge) => edge.node.topFailedRules
+    return prepareForExportRest(
+      exportSettings,
+      compliantSystems,
+      nonCompliantSystems,
+      unsupportedSystems,
+      neverReported,
+      topTenFailedRules
     );
+  }, [report, exportSettings]);
```
bastilian, line 45: Also missing the two fetch functions as dependencies.

src/SmartComponents/ReportDownload/hooks/apiQueryHooks.js
mkholjuraev: I believe now we are overusing the useCallback. Why? because the useSystemsFetch was already present without any useCallback and it was working nice.

bastilian: I know you have not written these, but we still need to fix them in order to comply with the Rules of Hooks.

mkholjuraev:
@RedHatInsights/team-interact this is ready for another review

src/SmartComponents/ReportDownload/ReportDownload.js
```diff
@@ -82,11 +82,56 @@ export const ReportDownload = () => {
...
 };
 
+ReportDownloadBase.propTypes = {
+  report: propTypes.object.isRequired,
+  loading: propTypes.bool,
+  error: propTypes.object,
+  data: propTypes.object.isRequired,
+};
+
+export const ReportsWithGraphQL = () => {
```
gkarat, line 101: nit-pick: 
```suggestion
export const ReportDownloadGraphQL = () => {
```

src/SmartComponents/ReportDownload/ReportDownload.js
```diff
@@ -82,11 +82,56 @@ export const ReportDownload = () => {
...
+  const serialisedData = useMemo(
+    () => dataSerialiser(data, reportDataMap),
+    [data]
+  );
+
+  return (
+    <ReportDownloadBase
+      report={serialisedData}
+      reportId={reportId}
+      {...{ loading, error, data: serialisedData }}
```
gkarat, line 126: nit-pick:
```suggestion
      loading={loading}
      error={error}
      data={serialisedData}
```

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
```diff
@@ -1,93 +1,69 @@
...
-      fetchResults: true,
-      fetchPolicy: 'no-cache',
-      variables: {
-        perPage,
-        page,
-        filter: `policy_id = ${policyId}`,
-        policyId,
-      },
-    });
+  return useCallback(async () => {
```
gkarat, line 27: Hooks can't be called in component's `return`. They should be placed in the component body

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
```diff
@@ -1,93 +1,69 @@
...
-  const { isResolving: isLoading, resolve } = usePromiseQueue();
+const useExportDataGraphQL = (report, exportSettings) => {
+  const fetchSystems = useSystemsFetch(report);
+  const fetchRules = useFetchRules(report);
 
-  return {
-    isLoading,
-    fetchBatched: (fetchFunction, total, filter, batchSize = 50) => {
-      const pages = Math.ceil(total / batchSize) || 1;
+  return useCallback(async () => {
```
gkarat, line 15: Same thing here

gkarat (commented):
This also results in something went wrong when I try to open the report download modal:

image

src/constants.js
```diff
@@ -249,3 +249,16 @@ export const paletteColors = {
 export const backgroundColors = {
   light300: '#f0f0f0', //'--pf-global--BackgroundColor--light-300',
 };
+
+export const reportDataMap = {
+  title: 'policy.name',
+  business_objective: 'businessObjective',
+  compliance_threshold: 'complianceThreshold',
+  os_major_version: 'osMajorVersion',
+  profile_title: 'policyType',
```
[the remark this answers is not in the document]
LightOfHeaven1994: @gkarat I think it's correct one, I just took a look on GQL profiles request 
image

LightOfHeaven1994:
> This also results in something went wrong when I try to open the report download modal:
> 
> image

I think this happens because we lost Report IDs after one of PRs. https://github.com/RedHatInsights/compliance-frontend/pull/2158 this should fix it

src/SmartComponents/ReportDownload/ReportDownload.js
```diff
@@ -82,11 +82,56 @@ export const ReportDownload = () => {
...
+    />
+  );
+};
+
+export const ReportDownload = () => {
+  const isRestApiEnabled = useAPIV2FeatureFlag();
+
+  return isRestApiEnabled ? <ReportDownloadRest /> : <ReportsWithGraphQL />;
```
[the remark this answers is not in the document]
mkholjuraev: Good idea! Will do

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
mkholjuraev: BTW, I was hurried to answer. I do not see any issue with this specific use of useCallback. Because it is used inside a custom hook not in a component direactly which is supposed to be called on top of a react component.

bastilian: I don't think it is against the Rules to return the hook right away. The only place I could see issues it if the hook is called within an object. Like:


```js
const useExample = () => {
   ....

  return {
    someProp: {
      prop: useCallback(...)
    }
  }
}
```

This I believe might cause that return to always be considered as changed on every render.

gkarat: I also believe that this doesn't need useCallback here. In any case, currently, it has 4 dependencies as objects and functions which will always perform shallow comparison. There is no difference as it would be not wrapped in useCallback. Let's not focus on that at the moment and just make sure the component returns this async function.

src/SmartComponents/ReportDownload/hooks/useQueryExportData.js
gkarat: Let's remove useCallback here

gkarat (commented):
When opening the modal, we are missing policy title
image

also, the result PDFs are quite different

mkholjuraev:
@eshamard @bastilian In report download modal, graphQL fetches `name`, `policy.name`, `profileType` that are the same according what I see from the API result. However, with REST api misses `policyType` field. Instead it has `type` field that is meaningless. I have a [workaround](https://github.com/RedHatInsights/compliance-frontend/pull/2148/files#diff-8e3b187293d8e7902d08eff3c650466860fe527b0c7da8ec32c1ee4a450f97c5R259) to save time and not to get unblocked. Let me know if it is feasable.

bastilian:
@mkholjuraev I think the `policyType` is now the "profile_title". @LightOfHeaven1994 Can maybe confirm that.

LightOfHeaven1994:
> policyType
@mkholjuraev 
yes, you are right. PDF report needs Policy/Report name and Type. 
Policy/Report name is mutable field as when you are creating policy you can change policy name to something custom, while policy type is immutable field and coming from parent profile.
For visibility:
image

src/Utilities/dataSerialiser.js
```diff
@@ -11,13 +11,41 @@ const serialiseObject = (object, map) => {
...
+ * }
+ *
+ * const data = {
+ *  ref_id: 'some-ref-id',
+ *  profile_title: 'some-profile-title'
+ * }
+ *
+ * const result = dataTransformer(apiInstance.policies);
+ * console.log(result); //outputs  => { refId: 'some-ref-id', name: 'some-profile-title', policyType: 'some-profile-title' }
+ */
```
bastilian, line 48: ❤️

LightOfHeaven1994 (approved):
Business objective is not taken into PDF report when BO exists for Policy. I just found out that it's the same with GraphQL, but I think we can fix it here right away, wdyt @mkholjuraev ?

Other than that, looks good

mkholjuraev:
@LightOfHeaven1994 thank you! I am preparing a separate PR for fixing the Business objective.

PR #2151, opened 2024-09-06
feat(RHINENG-10394): Migrate System Detail page to REST API

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-10394.

This migrates the system details page to use Compliance REST API. Only the header and the policies cards are covered by this PR. The rules table will be updated in a separate one.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/PresentationalComponents/SystemPolicyCards/components/LoadingPolicyCards.js
```diff
@@ -2,15 +2,29 @@ import React from 'react';
 import { GridItem, Card, CardBody } from '@patternfly/react-core';
 import { Instagram } from 'react-content-loader';
 
-const LoadingPolicyCards = () =>
-  [...Array(3)].map((_item, i) => (
-    <GridItem span={4} key={i}>
+export const LoadingPolicyCard = () => (
```
bastilian, line 5: Can't the old Component not be extended to cover both cases and not have duplicate code?

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -111,11 +111,18 @@ SystemDetailsRest.propTypes = {
...
 
-export const SystemDetailsWrapper = (children) => {
+export const SystemDetailsWrapper = (props) => {
   const restApiEnabled = useAPIV2FeatureFlag();
-  const Component = restApiEnabled ? SystemDetailsRest : SystemDetailsGraphQL;
-
-  return <Component {...children} />;
+  const Component =
+    restApiEnabled === undefined ? (
+      <></>
```
bastilian, line 118: `React.Fragment` can be used here, or not?

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -30,7 +32,52 @@ const QUERY = gql`
...
+
+SystemDetailBase.propTypes = {
+  data: propTypes.object,
+  error: propTypes.object,
+  loading: propTypes.bool,
+  systemName: propTypes.string,
+  inventoryId: propTypes.string,
+};
+
+const SystemDetailsGraphQL = ({ route }) => {
```
gkarat, line 80: TODO: add JSDoc for these component

src/SmartComponents/SystemDetails/SystemDetails.js
```diff
@@ -40,31 +87,42 @@ export const SystemDetails = ({ route }) => {
...
+
+SystemDetailsRest.propTypes = {
+  route: propTypes.object.isRequired,
+};
+
+export const SystemDetailsWrapper = (props) => {
+  const restApiEnabled = useAPIV2FeatureFlag();
+  const Component =
+    restApiEnabled === undefined ? (
+      <React.Fragment />
```
mkholjuraev, line 118: Here as well, we usually show a spinner. I have a [component](https://github.com/RedHatInsights/compliance-frontend/pull/2148/files#diff-23cb5de09b869d7b20ec4d78c36756ddfa8fca0b9c71aa3bede14fccebec7ea9) that can be re-used here. But it is in review state at the moment.

src/SmartComponents/SystemPolicyCards/SystemPolicyCards.js
```diff
@@ -0,0 +1,58 @@
...
+    reportId: policy.id,
+    filter: `system_id=${inventoryId}`,
+  });
+
+  return loading ? (
+    <LoadingPolicyCards count={1} />
+  ) : (
+    <SystemPolicyCardPresentational
+      policy={dataSerialiser({ ...data.data[0], ...policy }, dataMap)}
+      style={{ height: '100%' }}
```
mkholjuraev, line 25: Why do we need to give this custom css height? Can we somehow avoid it?

src/SmartComponents/SystemPolicyCards/fixtures.js
```diff
@@ -0,0 +1,85 @@
+export const fixturesSystemReports = {
+  data: [
+    {
+      id: '0de744d1-b00d-4b6c-a524-de0eb94dbe97',
```
mkholjuraev, line 4: We started to use fake data factories. Do you think that we can use here as well?

mkholjuraev:
@gkarat gentle ping on this :)

src/PresentationalComponents/SystemPolicyCard/SystemPolicyCard.js
```diff
@@ -26,20 +26,12 @@ const SystemPolicyCard = ({ policy, style, onClick, isSelected }) => {
     policyType,
   } = policy;
   const passedPercentage = ((value, fixed = 0, withPercent = true) =>
-    value.toFixed(fixed) + (withPercent ? '%' : ''))(score);
+    Number(value).toFixed(fixed) + (withPercent ? '%' : ''))(score);
```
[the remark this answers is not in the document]
gkarat: @mkholjuraev, with the REST component, the percentage value is string for some reason (I guess after the serialiser). This Number constructor ensures the value is always of the number type before calling `toFixed`.

src/SmartComponents/SystemDetails/ComplianceDetail.js
```diff
@@ -11,6 +11,8 @@ import { ErrorCard } from 'PresentationalComponents';
 import natsort from 'natsort';
 
 import EmptyState from './EmptyState';
+import useAPIV2FeatureFlag from '../../Utilities/hooks/useAPIV2FeatureFlag';
```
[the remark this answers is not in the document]
gkarat: I know it's inconsistent with the original code base, but the relative path works better for automation like Ctrl + click in VScode allowing to quickly jump to the module file. There is a lot of places where we started mixing this, so I am not sure it's a real problem.

src/SmartComponents/SystemPolicyCards/SystemPolicyCards.js
gkarat: This was taken from the original code (I didn't add this). I have tested the view without this style property, and all looks good. I will remove it.

src/SmartComponents/SystemPolicyCards/fixtures.js
gkarat: Good note, fixed

LightOfHeaven1994 (commented):
from functionality point of view, everything looks fine to me, great work!!!

mkholjuraev (commented):
@gkarat jest tests are failing and they seem valid failures.

src/SmartComponents/SystemDetails/ComplianceDetail.js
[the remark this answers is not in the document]
bastilian: > better for automation like Ctrl + click in VScode allowing to quickly jump to the module file

This may just be an issue with VSCode not being properly configured to resolve aliases. In any case we should use the aliases in code.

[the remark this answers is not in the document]
bastilian: I thought maybe it'd be good to put the general guidance on how to utilise the aliases when importing, to the README. 

But in general think of the aliases as modules, if you are within a "module", for example in the PresentationalComponents and you want to import another component within the PresentationalComponents, you should use a relative path. If it would be a hook or helper from the Utilities "module" like here an import path using the `Utilities` alias would be better.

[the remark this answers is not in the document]
bastilian: There is also a benefit that may not be obvious, but using the aliases does help webpack bundling, because it doesn't need to associate multiple paths to one module/file.

src/SmartComponents/SystemPolicyCards/SystemPolicyCards.test.js
```diff
@@ -67,7 +67,10 @@ describe('SystemPolicyCards', () => {
     );
     screen.getByText(
       `${testResultsData[0].failed_rule_count} rule${
-        testResultsData[0].failed_rule_count > 1 ? 's' : ''
+        testResultsData[0].failed_rule_count > 1 ||
+        testResultsData[0].failed_rule_count === 0
```
mkholjuraev, line 71: would not this result in plural when the rule count is 0? Could this cause an issue?

mkholjuraev (approved):
LGTM! Works fine now. Just one question before merging.

src/SmartComponents/SystemPolicyCards/SystemPolicyCards.test.js
gkarat: Yes, this can result in a value of 0 or more than 1 which means we have to use plural. This was fixed because the last test failed after a factory generate 0 failed rule count

PR #2143, opened 2024-08-29
Policies async

AsToNlele:
WIP

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #2152, opened 2024-09-06
feat(RHINENG-10392): Migrate systems page and component to REST

mkholjuraev:

This PR migrates Compliance Systems page and at the same time migrates all the required API fetches in SystemTable component that is shared by multiple pages/components to use REST API when the flag is enabled. The flag to determine between GraphQL and REST should be provided from consumer components of SystemsTable. Currently
tag filtering is disabled completely, while bulk selection uses only GraphQL due to backend issues

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


mkholjuraev:
Working on those test files. They seem to be because of the compliance-client version mismatch.

src/SmartComponents/SystemsTable/SystemsTableRest.js
```diff
@@ -0,0 +1,285 @@
...
+  });
+
+  useInventoryUtilities(inventory, selectedIds, activeFilterValues);
+
+  const onComplete = (result) => {
+    setTotal(result.meta.totalCount);
+    setItems(result.entities);
+    setPerPage(result.perPage);
+    setIsLoaded(true);
+    setCurrentTags && setCurrentTags(result.meta.tags);
```
gkarat, line 126: Why we need to check for setCurrentTags to be defined?

src/SmartComponents/SystemsTable/SystemsTableRest.js
```diff
@@ -0,0 +1,285 @@
...
+  const fetchSystems = useFetchSystemsV2({
+    onComplete,
+    fetchApi,
+    systemFetchArguments,
+  });
+
+  const getEntities = useGetEntities(fetchSystems, {
+    selected: selectedIds,
+    columns,
+  });
```
gkarat, line 149: For useFetchSystemsV2, you pass arguments with the single object, but for useGetEntities, the arguments are separated. Is there some pattern we could stick to for all hooks? I think we also need to consider performance and make sure the hooks are not unnecessarily updated

src/SmartComponents/SystemsTable/SystemsTableRest.js
```diff
@@ -0,0 +1,285 @@
...
+    filter: systemsFilter,
+    selected: selectedIds,
+    total,
+    fetchArguments: {
+      ...systemFetchArguments,
+    },
+  });
+
+  return (
+    <StateView
```
gkarat, line 161: We are having almost identical duplicates of this return value (check SystemsTableGraphQL component). I guess we could create some sort of SystemsTableBase that would be rendered by both GraphQL and REST wrapper components. What do you think?

src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
```diff
@@ -91,4 +93,41 @@ export const useFetchSystems = ({ query, onComplete, variables, onError }) => {
   };
 };
 
-export default useFetchSystems;
+export const useFetchSystemsV2 = ({
```
gkarat, line 96: Can't this be merged with useSystemsQuery hook? Otherwise, I am not sure what is the purpose of useSystemsQuery

src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
```diff
@@ -91,4 +93,41 @@ export const useFetchSystems = ({ query, onComplete, variables, onError }) => {
...
+      } catch (error) {
+        if (onError) {
+          onError(error);
+          return { entities: [], meta: { totalCount: 0 } };
+        } else {
+          throw error;
+        }
+      }
+    },
+    [systemFetchArguments]
```
gkarat, line 131: What if onComplete/onError/fetchApi change? We will not update the returned function?

AsToNlele:
Could you please commit the `package.json` with the updated `compliance-client` package as well?

AsToNlele (commented):
While going through the pagination I noticed the pages are not moving by the per page value (like 50), but the offset is incrementing by one, so we're moving by a single item between pages.

The offset is set to 1 by default, so we will always miss one item.
Even when you search for a single item you don't see it.

You can get the serialized state from the `useSerialisedTableState()` hook like [here](https://github.com/RedHatInsights/compliance-frontend/pull/2143/files#diff-46075fd63ea445df545ad357cafa7b9f689443bc2cec6d83bd3c7e532bdc08ebR105-R120)

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
```diff
@@ -23,12 +26,43 @@ export const QUERY = gql`
...
+      undefined,
+      perPage,
+      page,
+      combinedVariables.sortBy,
+      combinedVariables.filter
+    )
+    .then(({ data: { data = [], meta = {} } = {} } = {}) => ({
+      data: dataSerialiser(data, dataMap),
+      meta,
+    }));
```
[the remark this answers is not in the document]
bastilian: @mkholjuraev In what way does using a/the hook affect performance?

src/SmartComponents/SystemsTable/SystemsTableRest.js
```diff
@@ -0,0 +1,284 @@
...
+    perPage,
+    onSelect: onSelectProp,
+    preselected: preselection,
+    fetchArguments: systemFetchArguments,
+    currentPageIds: items.map(({ id }) => id),
+  });
+
+  useInventoryUtilities(inventory, selectedIds, activeFilterValues);
+
+  const onComplete = (result) => {
```
bastilian, line 121: Pwetty sure this should be within a `useCallback`.

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -286,15 +324,31 @@ export const useSystemBulkSelect = ({
     },
   });
 
+  const fetchSystemsRest = useFetchSystemsV2({
+    dataMap,
+    systemFetchArguments: fetchArguments,
+    onError: () => {
```
bastilian, line 330: This anonymous function might cause the `useCallback`[0] below to be invalidated on every render and cause rerenders.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/2152/files#diff-658b1e510ac2f7ad7576115c6e6e99cb214f0ec60e1b73bea71e2b926bd14a44R101

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -205,18 +221,38 @@ export const useSystemsExport = ({
...
-      ...fetchArguments.variables,
+      ...fetchArguments?.variables,
       ...(fetchArguments.tags && { tags: fetchArguments.tags }),
       filter: selectionFilter
-        ? `${fetchArguments.variables.filter} and (${selectionFilter})`
-        : fetchArguments.variables.filter,
+        ? `${fetchArguments.variables?.filter} and (${selectionFilter})`
+        : fetchArguments.variables?.filter,
+    },
+    onError: () => {
```
bastilian, line 239: Same as below. with the `onError`

src/SmartComponents/SystemsTable/SystemsTableGraphQL.js
```diff
@@ -0,0 +1,314 @@
...
+    perPage,
+    onSelect: onSelectProp,
+    preselected: preselection,
+    fetchArguments: systemFetchArguments,
+    currentPageIds: items.map(({ id }) => id),
+  });
+
+  useInventoryUtilities(inventory, selectedIds, activeFilterValues);
+
+  const onComplete = (result) => {
```
bastilian, line 154: Same as below.

src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
bastilian: In any case, this hook needs a test. :)

src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
```diff
@@ -91,4 +92,41 @@ export const useFetchSystems = ({ query, onComplete, variables, onError }) => {
...
+
+        const serialisedData = {
+          entities: data,
+          meta: {
+            ...(requestVariables?.tags && { tags: requestVariables.tags }),
+            totalCount: meta.total || 0,
+          },
+        };
+
+        onComplete && onComplete(serialisedData);
```
bastilian, line 119: `onComplete?.(serialisedData)`

src/SmartComponents/ComplianceSystems/ComplianceSystems.js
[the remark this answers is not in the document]
mkholjuraev: Using useSystemsQuery in this sense means the use of useQuery hook with [3 states](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/Utilities/hooks/useQuery/useQuery.js#L31) to keep the API request status. As the function is used by the Inventory module and the module does not use those states, useQuery is not a good idea. Here plain JS function is enough.

src/SmartComponents/SystemsTable/SystemsTableRest.js
mkholjuraev: I just copied this from existing component and did not notice the need :). I will wrap it

src/SmartComponents/SystemsTable/hooks/useFetchSystems.js
mkholjuraev: This was already as I have it now before. But, I will make the suggested change.

src/SmartComponents/SystemsTable/SystemsTableRest.js
mkholjuraev: Now, useFetchSystemsV2 has the same pattern as useGetEntities

src/SmartComponents/SystemsTable/SystemsTable.js
```diff
@@ -1,317 +1,24 @@
...
-            ],
-          })}
-        />
-      </StateViewPart>
-    </StateView>
-  );
-};
+import { Bullseye, Spinner } from '@patternfly/react-core';
+
+const SystemsTable = (props) => (
```
bastilian, line 7: Is it not possible to leave the majority of the `SystemsTable` intact and only pass in the things that are requested differently from wrapper components, like we do with `Reports` and `Policies`?

mkholjuraev: This was my first approach. But later, I had to change this. Because there are a lot more differences in how the SystemsTable component's behaviour when it comes to GraphQL and Rest API. The first approach resulted in bloated component at the end. Now, the two components are purely responsible for their type and later when we finish the migration, the GraphQL component can just be removed. The other component just works out of the box.

src/SmartComponents/SystemsTable/hooks/useFetchSystems.test.js
```diff
@@ -185,3 +185,69 @@ describe('useFetchSystems', () => {
     });
   });
 });
+
+describe.only('useFetchSystemsV2', () => {
```
gkarat, line 189:
```suggestion
describe('useFetchSystemsV2', () => {
```

src/SmartComponents/SystemsTable/SystemsTableRest.js
```diff
@@ -0,0 +1,294 @@
...
+  const preselection = useMemo(
+    () => preselectedSystems.map(({ id }) => id),
+    [preselectedSystems]
+  );
+
+  const {
+    selectedIds,
+    tableProps: bulkSelectTableProps,
+    toolbarProps: bulkSelectToolBarProps,
+  } = useSystemBulkSelect({
```
gkarat, line 111: We are not passing `query` to `useSystemBulkSelect`. This is most probably a reason why selection doesn't work. When I select one system from the table, I get infinite spinner and the error in the console:

```
Uncaught (in promise) Invariant Violation: An error occurred! For more details, see the full error text at https://go.apollo.dev/c/...
```

gkarat (commented):
Also, can't export CSV or JSON


https://github.com/user-attachments/assets/7a0c1c70-a422-4d7b-813d-d8a7601a666f

gkarat (commented):
Comparing to the current GraphQL implementation, I see a visual difference for the Last seen column: my system reported last time around 30 mins ago, and this shouldn't be marked with red:

image

src/SmartComponents/SystemsTable/SystemsTableRest.js
mkholjuraev: The bulks select is disabled due to the API shortcoming. RHINENG-12138

gkarat (commented):
I have unresolved some of the threads for useFetchSystems.js and useFetchSystems.test.js since they are not addressed in the last commits.

Also, I see that we have disabled the bulk select which is expected. But when I select individual items, the select checkbox get checked and also disabled for some reason.

image

gkarat (approved):

PR #2176, opened 2024-09-30
feat(RHINENG-12789): Migrate the first step of Create policy wizard

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-12789.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   ApolloProvider,
   ApolloClient,
@@ -19,12 +20,16 @@ const client = new ApolloClient({
   cache: new InMemoryCache(),
 });
 
-const AppEntry = () => (
-  <Provider store={init(process?.env?.NODE_ENV || 'production').getStore()}>
+const AppEntry = ({ environment = 'production' }) => (
+  <Provider store={init(environment).getStore()}>
     <ApolloProvider client={client}>
       <App />
     </ApolloProvider>
   </Provider>
 );
+
+AppEntry.propTypes = {
+  environment: PropTypes.bool,
+};
 
 export default AppEntry;
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {
   ApolloProvider,
   ApolloClient,
@@ -20,16 +19,12 @@ const client = new ApolloClient({
   cache: new InMemoryCache(),
 });
 
-const AppEntry = ({ logger }) => (
-  <Provider store={init(logger).getStore()}>
+const AppEntry = () => (
+  <Provider store={init(process?.env?.NODE_ENV || 'production').getStore()}>
     <ApolloProvider client={client}>
       <App />
     </ApolloProvider>
   </Provider>
 );
-
-AppEntry.propTypes = {
-  logger: PropTypes.any,
-};
 
 export default AppEntry;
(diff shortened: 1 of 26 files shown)


src/store/index.js
```diff
@@ -2,14 +2,22 @@ import { getRegistry } from '@redhat-cloud-services/frontend-components-utilitie
...
-    promiseMiddleware,
-    ...middleware.filter((item) => typeof item !== 'undefined'),
-  ]);
+  registry = getRegistry(
+    {},
+    [
+      promiseMiddleware,
+      ...middleware.filter((item) => typeof item !== 'undefined'),
+    ],
+    (process.env.NODE_ENV !== 'production' &&
```
bastilian, line 16: Watch out, there is no `process.env.NODE_ENV` in a browser. I think this would be better done by extending the `DevAppEntry.js` to pass down a debug option to enable the redux dev tools.

gkarat: I have tried to revisit this at https://github.com/RedHatInsights/compliance-frontend/pull/2176/commits/0164e1d66967121a453f6c0ce7831bf0236efc33, can you take a look?

LightOfHeaven1994 (commented):
I took a look on functionality, generally it works. But I think we can slightly improve how loading works:

1. Every time I click on RHEL X button, app sends backend request. Even if I have already clicked once, it will still send a request. Do you think we can cache response or smth like that? As supported_profiles endpoint is immutable.
2. From what I saw, table loads in couple seconds, while we are waiting, do you think it makes sense to add loading state or so? see attached recording

[Screencast from 2024-10-07 12-11-10.webm](https://github.com/user-attachments/assets/919160d4-afb2-4521-86b9-fde6b1197dc3)

mkholjuraev:
Let's integrate to AsyncTableTools on the fly, WDYT @gkarat ?

src/SmartComponents/CreatePolicy/CreateSCAPPolicy/CreateSCAPPolicyRest.test.js
```diff
@@ -0,0 +1,175 @@
...
+        <CreateSCAPPolicyRest change={change} selectedOsMajorVersion={8} />
+      </TestWrapper>
+    );
+
+    await userEvent.click(
+      screen.getByRole('radio', {
+        name: /select row 0/i,
+      })
+    );
+    expect(change).toBeCalledTimes(4);
```
johnsonm325, line 173: Is this expected? Is it possible that 3 of the 4 times is coming from tests above? How do we fix this? Should we expect change to be called in previous tests?

gkarat: @johnsonm325, not really, once a policy selected, it literally makes 4 `change` calls: https://github.com/RedHatInsights/compliance-frontend/blob/c1b56e75ba87f81e5ea5b1689c59b5a393c75f3c/src/SmartComponents/CreatePolicy/CreateSCAPPolicy.js#L94-L97.

LightOfHeaven1994 (commented):
@gkarat we are still missing limit=100 for PolicyType table + seems travis test failure is relevant. Other than that looks good

AsToNlele (approved):
api is called correctly, tried the `process.env.NODE_ENV` as well, filtering works as well
LGTM

src/AppEntry.js
```diff
@@ -20,16 +19,12 @@ const client = new ApolloClient({
   cache: new InMemoryCache(),
 });
 
-const AppEntry = ({ logger }) => (
-  <Provider store={init(logger).getStore()}>
+const AppEntry = () => (
+  <Provider store={init(process?.env?.NODE_ENV || 'production').getStore()}>
```
bastilian, line 23: We should really not base any conditions on `node` specific constants in browser code. We can instead of this just pass an `env` prop to `AppEntry` in `DevAppEntry`.

mkholjuraev: I agree, now I remember we faced issues related to the redux logger middleware due to improper env setup in Inventory app.

PR #2192, opened 2024-10-03
feat(RHIENG-10400): migrate policy systems table to REST

mkholjuraev:

This migrates the Policy systems table to Rest API. Due to API shortcomings, ssgVersion column has been hidden. Also, all the disabled functionalities in SystemsTable component applies to this page as well. They are selection, tag filter.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Utilities/hooks/api/useOperatingSystems.js
+++ b/src/Utilities/hooks/api/useOperatingSystems.js
@@ -1,7 +1,7 @@
 import useQuery, { apiInstance } from '../useQuery';
 
 const useOperatingSystemsQuery = (filters) => {
-  return useQuery(apiInstance.systemsOS, { params: [undefined, filters] });
+  return useQuery(apiInstance.systemsOS, { params: [null, filters] });
 };
 
 export default useOperatingSystemsQuery;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "compliance-frontend",
   "version": "1.1.0",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
@@ -42,6 +42,7 @@
         "redux-form": "^8.3.10",
         "redux-logger": "^3.0.6",
         "redux-promise-middleware": "^6.1.2",
+        "semver": "^7.6.3",
         "use-deep-compare-effect": "^1.8.1"
       },
       "devDependencies": {
@@ -202,6 +203,14 @@
         "url": "https://opencollective.com/babel"
       }
     },
+    "node_modules/@babel/core/node_modules/semver": {
+      "version": "6.3.1",
+      "dev": true,
+      "license": "ISC",
+      "bin": {
+        "semver": "bin/semver.js"
+      }
+    },
     "node_modules/@babel/eslint-parser": {
       "version": "7.25.1",
       "dev": true,
@@ -217,6 +226,14 @@
       "peerDependencies": {
         "@babel/core": "^7.11.0",
         "eslint": "^7.5.0 || ^8.0.0 || ^9.0.0"
+      }
+    },
+    "node_modules/@babel/eslint-parser/node_modules/semver": {
+      "version": "6.3.1",
+      "dev": true,
+      "license": "ISC",
+      "bin": {
+        "semver": "bin/semver.js"
       }
     },
     "node_modules/@babel/generator": {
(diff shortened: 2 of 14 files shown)


LightOfHeaven1994 (commented):
We are also missing:

1. Tags filtering (searching by tags)
2. Select all (just wanted to mention, I know it's going to be covered later)

mkholjuraev:
@LightOfHeaven1994 Thanks for the initial review. Yes,  as I mentioned in the PR description, tag filtering, selection are disabled. Regarding the OS endpoint issue, if I understand correctly, each page should use different endpoints for fetching OS versions right?

LightOfHeaven1994:
> @LightOfHeaven1994 Thanks for the initial review. Yes, as I mentioned in the PR description, tag filtering, selection are disabled. Regarding the OS endpoint issue, if I understand correctly, each page should use different endpoints for fetching OS versions right?

yes, each endpoint has it's own /os_versions . It's visible in the APIv2 doc

mkholjuraev:
Needs to be reviewed after: https://github.com/RedHatInsights/insights-inventory-frontend/pull/2262

src/Utilities/hooks/useAPIV2FeatureFlag.js
```diff
@@ -1,5 +1,5 @@
 import useFeatureFlag from './useFeatureFlag';
 
-const useAPIV2FeatureFlag = () => useFeatureFlag('compliance-api-v2');
```
AsToNlele, line 3: Let's not forget to revert this

src/Utilities/hooks/api/useOperatingSystems.js
```diff
@@ -1,7 +1,7 @@
 import useQuery, { apiInstance } from '../useQuery';
 
 const useOperatingSystemsQuery = (filters) => {
-  return useQuery(apiInstance.systemsOS, { params: [undefined, filters] });
+  return useQuery(apiInstance.systemsOS, { params: [null, filters] });
```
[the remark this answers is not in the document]
mkholjuraev: @LightOfHeaven1994 as of the last commit, we are using proper endpoints to fetch operating systems.

src/Utilities/hooks/useAPIV2FeatureFlag.js
mkholjuraev: ah, very true :)

mkholjuraev: Done!

LightOfHeaven1994 (commented):
When we are sorting by os_version on Policy details Systems page I see that we are sending `sort_by=os_major_version=X and sort_by=os_minor-version=Y` - we don't need that on that page since major version is constant. We need to sort only by os_minor version here

image

mkholjuraev:
@LightOfHeaven1994 

> we don't need that on that page since major version is constant. We need to sort only by os_minor version here

As we share the same code under the hood for several pages and this does not affect the actual behaviour, I would suggest not to add additional conditions/complexity to the sorting. It is okay to have both major and minor versions passed into request. What do you think?

LightOfHeaven1994:
> @LightOfHeaven1994
> 
> > we don't need that on that page since major version is constant. We need to sort only by os_minor version here
> 
> As we share the same code under the hood for several pages and this does not affect the actual behaviour, I would suggest not to add additional conditions/complexity to the sorting. It is okay to have both major and minor versions passed into request. What do you think?

agree, let's leave it like this for now. We can always return to it later.

LightOfHeaven1994 (approved):
I think we are good now, great work!

AsToNlele (commented):
We're missing empty state when the policy has no systems.

Rest API
image

GraphQL
image

Other than that well done!

mkholjuraev:
@AsToNlele I am aware of the issue and it is tricky to resolve due to the difference in the API results. I am happy to explain/discuss it at the standup.

mkholjuraev:
@AsToNlele update to my prev comment:

Actually, the empty state is not shown in both stage with GraphQL as well. You saw the empty state because the policy had no added systems at all. In this case, the empty state with the warning is shown. In case, policy has at least one system and there is an active filter that results in an empty state, then the table would not have an empty state in it even before the migration. Thus, this is not a regression.

PR #2210, opened 2024-10-18
feat(RHINENG-14071): add sentry to compliance

adonispuente:
This adds the ability to upload source maps from compliance into sentry. It also adds the pointer to tell sentry which project to send the error too.

Testing is a bit tricky, as currently you need to run chrome-prod in tandem with inventory to test the actual error. You will need an authKey present (reach out to me for the file).

in compliance, run `SENTRY_AUTH_TOKEN=IWILLGIVETHISTOYOU  npm run static`,
Create an error 
You can create an easy error by adding `<button onClick={() => methodDoesNotExist()}/>` anywhere and clicking the button.

In chrome youd need to have a prod account ->
 change webpack.config.js file env var too 'prod-stable' ->
 add this to routes` ...(true && { '/apps/inventory': { host: 'http://localhost:8003', }, }),`->
  and `run npm run dev`.

Observer that the error then goes to the compliance-rhel project and has valuable data inside.

Here is an example: https://red-hat-it.sentry.io/issues/6002273624/?project=4508083145408512&referrer=onboarding-first-event-indicator


bastilian (approved):
LGTM! Thank you @adonispuente!

PR #2214, opened 2024-10-22
feat(RHINENG-13052): Migrate PolicyRules Page to v2

adonispuente:
Implements https://issues.redhat.com/browse/RHINENG-13052.

This migrates policy rules to use apiv2. Currently the route to get to the default rules page is unavailable when apiv2 and async table FF are turned on. 
So in order to test, simply navigate to the url : https://stage.foo.redhat.com:1337/insights/compliance/scappolicies/7aafaf4b-705c-411b-b39d-220da89ee254/default_ruleset#SIDs=&tags= 

Currently, the ID on the page is hardcoded (so even inserting https://stage.foo.redhat.com:1337/insights/compliance/scappolicies/blabla/default_ruleset#SIDs=&tags=_ will work) This will change once the modals are migrated. 

Once on the page just verify that both tables load in proper data. 
Ive changed the routes, so right now securityGuideData is optional. but once v2 is completed mirgrated that param will always be in the url

This currently Mimics Tailorings Tab. Investigating why sorting is so wonky.
Tree view still has no twigs and leaves
Not able to trigger fresh data off table state so no new request are firing when changing filters/pagination


src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -3,69 +3,99 @@ import { useQuery, gql } from '@apollo/client';
...
+import { Spinner} from '@patternfly/react-core';
+import { usePolicyRulesList } from './usePolicyRulesList';
+import useAPIV2FeatureFlag from '../../Utilities/hooks/useAPIV2FeatureFlag';
+import PolicyRulesHeader from './PolicyRulesHeader';
+import { useProfileTree } from './useProfileTree';
 
 const PROFILES_QUERY = gql`
-  query PR_Profile($policyId: String!) {
-    profile(id: $policyId) {
+query PR_Profile($policyId: String!) {
```
bastilian, line 16: This query should not change at all.

src/SmartComponents/PolicyRules/usePolicyRulesList.js
```diff
@@ -0,0 +1,8 @@
+import useQuery, { apiInstance } from '../../Utilities/hooks/useQuery';
+
+export const usePolicyRulesList = (security_guide_id, profile_id) => {
+  return useQuery(() => {
+    return apiInstance.profileRules(security_guide_id, profile_id);
```
bastilian, line 5: We should make this similar to how the other "api hooks" look, with the apiInstance things for the first argument and the parameters as the second.

bastilian:
@adonispuente Let's update the link here[0] and make sure it can show up properly when enabled.

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/Tailorings/components/TabHeader.js#L47

adonispuente:
@bastilian Ive added the optional param in the routes page, and made the link render a conditional id based of v2 FF. 
I also adopted the new hook schema and sent that up.

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -1,13 +1,20 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useQuery, gql } from '@apollo/client';
 import { useParams } from 'react-router-dom';
 import propTypes from 'prop-types';
-import PageHeader, {
-  PageHeaderTitle,
-} from '@redhat-cloud-services/frontend-components/PageHeader';
+import PageHeader from '@redhat-cloud-services/frontend-components/PageHeader';
 import RulesTable from '../../PresentationalComponents/RulesTable/RulesTable';
```
mkholjuraev, line 6: let use relative import syntax:  
```suggestion
import RulesTable from '@/PresentationalComponents/RulesTable/RulesTable';
```

 there are other occurrences  below.

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -35,9 +42,8 @@ const PROFILES_QUERY = gql`
   }
 `;
 
-const PolicyRules = () => {
+const PolicyRulesGraphQL = ({ v2 }) => {
```
mkholjuraev, line 45: Super nitpick: how about using more meaningful name for this flag. For example: `apiV2` or `isRestEnabled`

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -49,50 +55,182 @@ const PolicyRules = () => {
...
+      rules={data?.profile?.rules}
+    />
+  );
+};
+
+PolicyRulesGraphQL.propTypes = {
+  v2: propTypes.bool,
+};
+
+const PolicyRulesRest = ({ v2 }) => {
```
mkholjuraev, line 75: In general, I did not quite understand the data fetching implemented in this component. Questions to clarify:
1. Are policy rules batch fetched? If so, I would suggest to move all the logic into a separate hook and preferably to a new file. Also, using [useFetchBatched](https://github.com/RedHatInsights/frontend-components/blob/e66918a2eb86bfcdbae024e9997bb6ae60c9c8b6/packages/utils/src/useFetchBatched/useFetchBatched.ts#L6) hook would be ideal. 
2. Why not wrap this component with `TableStateProvider` here directly, instead of `PolicyRulesBase`. Because, this is solely REST related thingy and `PolicyRulesBase` does not have to deal with it.

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -49,50 +55,182 @@ const PolicyRules = () => {
...
+      // Check if there is more data to fetch based on meta info
+      if (data.meta && data.meta.offset + data.meta.limit < data.meta.total) {
+        // Trigger the next page by increasing the offset
+        setOffset((prevOffset) => prevOffset + limit);
+      }
+      setIsPageLoading(false);
+    }
+  }, [data, loading, setOffset]);
+
+  if (isPageLoading || rulesTreeLoading || ruleGroupsLoading) {
```
mkholjuraev, line 126: Let's use [StateView](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/StateView/StateView.js#L12) component for displaying different states. Example: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CompliancePolicies/CompliancePolicies.js#L57

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -49,50 +55,182 @@ const PolicyRules = () => {
...
+    <React.Fragment>
+      <PolicyRulesHeader
+        name={headerName}
+        benchmarkVersion={benchmarkVersion}
+        osMajorVersion={osMajorVersion}
+      />
+
+      <div className="pf-v5-u-p-xl" style={{ background: '#fff' }}>
+        {v2 ? (
+          <TableStateProvider>
```
mkholjuraev, line 181: How about, passing already rendered components (RulesTableRest, RulesTable) into `PolicyRulesBase`, instead of rendering them with correct props here? With this, you would not have to pass many props as those props would have already be rendered in corresponding components. You could also use [GatedComponents.js](https://github.com/RedHatInsights/compliance-frontend/blob/19faaef241abaf125a778dc02ed36820b672e093/src/PresentationalComponents/GatedComponents.js#L6) to avoid passing feature flag.

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -49,50 +55,182 @@ const PolicyRules = () => {
...
+  profile: propTypes.array,
+  rules: propTypes.array,
+  osMajorVersion: propTypes.string,
+  benchmarkVersion: propTypes.string,
+  headerName: propTypes.string,
+  ruleTree: propTypes.any,
+};
+
+const PolicyRulesWrapper = () => {
+  const apiV2Enabled = useAPIV2FeatureFlag();
```
mkholjuraev, line 222: Here also, we could use `GatedComponents`

src/SmartComponents/PolicyRules/PolicyRules.js
adonispuente: for 1. I grabbed the batching set up from https://github.com/RedHatInsights/compliance-frontend/blob/21825e634bba097eac8f96109713a4c8ce56bca6/src/PresentationalComponents/Tailorings/hooks/useSecurityGuideData.js#L27. Ill try to move it to a seperate file. Also, it looks like compliance has its own [useFetchBatched](https://github.com/RedHatInsights/compliance-frontend/blob/21825e634bba097eac8f96109713a4c8ce56bca6/src/SmartComponents/SystemsTable/hooks.js#L71),  though it looks like the same thing, Should we just delete the compliance one all together? 

Ill attempt to use the suggested version of batching though

2. Good point, done

adonispuente: Ive opted to go with the implementation there, but removed the redundant useEffect logic

mkholjuraev:
Thank you @adonispuente . Some issues I see:
1. export button is missing
2. both Name and Severity filters do not work
3. Somehow when I dive into tree view and select  'system settings => Selinux =>  third row (near caret icon)', I get 'valueDefinitions is undefined' error and page breaks
4. Somehow, all of the last tree level has empty cells and unknown, manual values rendered. I think it is wrong.
image

LightOfHeaven1994:
Additionally to @mkholjuraev comments, sorting doesn't work too. Maybe we can fix rule tree issues later as I think Sebastian was working on enabling it and then it will be easy to have working rule tree. But it's up to you

mkholjuraev (changes requested):
I still see some of the issues:

1. both Name and Severity filters do not work yet and applying a filter suddenly changes to table view from tree view.
2. Somehow when I dive into tree view and select 'system settings => Selinux => third row (near caret icon)', now I get empty description cell view.
3. Somehow still , all of the last tree level has empty cells and unknown, manual values rendered. I think it is wrong.

adonispuente:
@mkholjuraev Sorry, I should have mentioned the issues inside another comment instead of udpating the post, youre correct. Currently the filters should work though, its just they only work with the original 10 items requested. (no longer fetching all rules). Sorting doesnt work well either but thats how it is in RulesRest everywhere. the 'Leaves' and 'twigs' are always empty arrays now and i am also investigating that

src/SmartComponents/PolicyRules/PolicyRules.js
```diff
@@ -35,64 +41,141 @@ const PROFILES_QUERY = gql`
...
+        <PageHeader>
+          <Spinner />
+        </PageHeader>
+      </StateViewPart>
+      <StateViewPart stateKey="data">
+        <PolicyRulesBase
+          name={data?.profile?.name}
+          benchmarkVersion={data?.profile?.benchmark?.version}
+          osMajorVersion={data?.profile?.osMajorVersion}
+          component={
```
bastilian, line 67: Maybe this can use `children`. Same below. https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children

LightOfHeaven1994 (approved):
I took a look and generally looks good. We are getting closer to the release and I don't want to hold it more. There are couple things that can be fixed later:

1. Sorting (but it was an issue before with GQL)
2. Table tree reloads after opening each node
3. If I use search by name + expand some rules - page crashes (Search by "Sel", find rule "Configure SELinux Policy" and try to expand it) -  I will create a separate jira to track this

PR #2223, opened 2024-10-24
fix(RHINENG-13404): Fix the total number in selected items

gkarat:
This fixes the bulk select showing incorrect number of total selected items when users click on "Select all". The numbers mismatched because of the different fetch arguments being passed to the bulk select hook and to the systems table.

## How to test

1. Go to the /systems page or any other page that shows systems
2. Select all items using the bulk select component
3. Make sure the number of selected items is the same as in pagination

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


LightOfHeaven1994 (approved):
I see selection number is displayed correctly now, ack

mkholjuraev (approved):
It seems I was wrong with the race conditions. I am sure that I faced such thing while working on the filtering on SystemsTable component. but at that time, I was using a state instead of a ref. Thus, race condition happened. But, as of the refs it is not happening. 

So, let's remove the `fetchArgumentsRef` completely, and use only `combinedFetchArgumentsRef`. We need to also provide it to useExport hook.

This fixes the issue. tested locally. Thus approving.

PR #2226, opened 2024-10-24
feat(Api): Enable passing params object

bastilian:
This PR changes the API functions/hooks to allow better passing of parameters.

The changes made to allow this:

* `useQuery` now accepts both, an object or an array as "parameters"
  * An object will be passed as the "params" object for the javascript-client API functions
  * An Array will be passed as the arguments for the API function
* The `useComplianceApi` hook is introduced to return a javascript-client API function with an axios instance having the platform interceptors, based on the api endpoint function name
* The `useComplianceQuery` now used the `useComplianceApi` hook to get an api function instance, and now **optionally** fetches data using the table state
* The api hooks, like `useReports`, `useSystems` etc. have been updated to use the `useComplianceQuery` hook
  * The hooks are now also consistent in form and consolidated in the `api` folder
* API v1 hooks have been moved to separate folder
* (minor) Aliases have been enabled for TypeScript to allow better editor integration

**How to test:**

Verify everything is still working as expected.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/Utilities/hooks/api/useComplianceApi.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const apiEndpoint = useMemo(() => {
+    const apiInstance = APIFactory(BASE_URL, complianceApi, axios);
+    return endpoint ? apiInstance[endpoint] : apiInstance;
+  }, [axios, endpoint]);
+
+  useEffect(() => {
+    if (endpoint && !apiEndpoint) {
+      console.error(`Endpoint ${endpoint} does not exist!`);
+      console.log('Available endpoints:', Object.keys(complianceApi));
```
gkarat, line 31: `throw`ing an error would be more appropriate IMO for more immediate feedback.

src/Utilities/hooks/api/useComplianceQuery.js
```diff
@@ -0,0 +1,93 @@
...
+ * @example
+ *
+ * const policiesApi = useComplianceQuery('reports', {
+ *  params: { filter: 'name NOT null'}
+ * })
+ *
+ */
+const useComplianceQuery = (
+  endpoint,
+  { params, ignoreTableState = false, ...options } = {}
```
gkarat: nit-pick: I would prefer to have ignoreTableState set to true by default. Isn't half or even majority of available endpoints are not table related (= have paginated/filtered/sorted list responses)?

src/Utilities/hooks/api/useComplianceQuery.js
```diff
@@ -0,0 +1,93 @@
...
+      sortBy,
+      ...params,
+      filter,
+    }),
+    [limit, offset, filter, sortBy, params]
+  );
+
+  const query = useQuery(apiEndpoint, {
+    params: !ignoreTableState ? paramsFromState : params,
+    skip: !serialisedTableState,
```
gkarat: What is the correlation between skipping the initial fetch and the fact that the query doesn't use the serialised table state? Do we want just to support `skip` parameter by `useComplianceQuery` and directly pass it here?

src/Utilities/hooks/api/useComplianceQuery.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const filter = useMemo(
+    () =>
+      params?.filter
+        ? `(${params.filter})${filters ? ` AND ${filters}` : ''}`
+        : filters,
+    [params, filters]
+  );
+
+  const paramsFromState = useMemo(
```
gkarat, line 90: nit-pick: perhaps, rename `paramsFromState` to `tableExtendedParams` or something similar? If I understand right, we add additional params like limit, offset, etc. if the query response is supposed to be a list/table reponse.

src/Utilities/hooks/useQuery/useQuery.js
```diff
@@ -73,8 +80,18 @@ const useQuery = (fn, options = {}) => {
...
-    [fn, fetchFn]
+    (fetchParams, setDataState) =>
+      fetchFn(
+        fn,
+        !Array.isArray(fetchParams)
+          ? {
+              ...params,
+              ...fetchParams,
+            }
+          : fetchParams,
```
gkarat, line 91: Can you help me understand why we feed here only `fetchParams`? Are we ignoring `params` in case `fetchParams` is an array?

src/Utilities/hooks/api/useComplianceQuery.js
gkarat: (sorry, was commenting on outdated code)

src/Frameworks/AsyncTableTools/hooks/useTableState/hooks/useContextOrInternalStateAndRefs.js
```diff
@@ -2,22 +2,23 @@ import { useEffect, useContext } from 'react';
...
+    if (!context) {
       console.warn(
         '%cNo context provided for useTableState.\n' +
           'Using internal state and refs.\n' +
           'Make sure a TableStateProvider is available!',
         'background: red; color: white; font-weight: bold;'
       );
     }
-  }, [contextState]);
+  }, [context]);
```
mkholjuraev, line 21: Is everything in `context` stable? Have we checked it already?

src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -53,11 +53,8 @@ export const useOsMinorVersionFilter = (showFilter, fetchArguments = {}) => {
     : [];
 };
 
-export const useOsMinorVersionFilterRest = (
-  showFilter,
-  fetchArguments = {}
-) => {
-  let { data = [] } = useOperatingSystemsQuery(fetchArguments);
+export const useOsMinorVersionFilterRest = (showFilter, params = {}) => {
```
mkholjuraev, line 56: Nitpick: we can remove this redundant function. Because, we are already using inventory OS filter.

src/Utilities/hooks/api/useComplianceQuery.test.js
```diff
@@ -1,19 +1,23 @@
...
-import usePagination from '../../../Frameworks/AsyncTableTools/hooks/usePagination';
+import { useSerialisedTableState } from '@/Frameworks/AsyncTableTools/hooks/useTableState';
+import usePagination from '@/Frameworks/AsyncTableTools/hooks/usePagination';
 import {
   paginationSerialiser,
   sortSerialiser,
-} from '../../../PresentationalComponents/ComplianceTable/serialisers';
-import TableStateProvider from '../../../Frameworks/AsyncTableTools/components/TableStateProvider';
-import useTableSort from '../../../Frameworks/AsyncTableTools/hooks/useTableSort';
+} from '@/PresentationalComponents/ComplianceTable/serialisers';
```
mkholjuraev, line 7: Some places we do not use any specific prefix in the import path like `Utilities/hooks/api/useOperatingSystems`. Here we are using `@` as an import path prefix. What is the difference between them?

I believe, we can use both of them. Does it make sense to stick to one?

src/Frameworks/AsyncTableTools/hooks/useTableState/hooks/useContextOrInternalStateAndRefs.js
bastilian: This is passing through the table context, so React should be able to properly handle any updates to it.

src/Utilities/hooks/api/useComplianceQuery.test.js
bastilian: There is only a minor difference. The `@/` will be replaced with `./src/`, while `Utilities/` will be replaced with `./src/Utilities/`, so using `@/Utilities` will also result in `./src/Utilities` and both can be used.

It might make sense to make the `@/` more preferred, but the benefits might only be very minor, like having to keep less aliases in sync between jest, webpack, typescript etc.

src/Utilities/hooks/useQuery/useQuery.js
bastilian: Yes, we are ignoring any params from the options here, if an array is passed, and will be passed on as the function arguments to the api function.

We can only safely merge the two if it is an object. We could attempt to make a union of the params if it is an array, but since we have already issues with having to keep params in the correct place, this might be very messy and just cause issues. 

I tried to mention the behaviour of the two functions in the docs, let me know if that helps to understand this.

src/Utilities/hooks/api/useComplianceApi.js
```diff
@@ -0,0 +1,38 @@
...
+const useComplianceApi = (endpoint) => {
+  const axios = useAxiosWithPlatformInterceptors();
+
+  const apiEndpoint = useMemo(() => {
+    const apiInstance = APIFactory(BASE_URL, complianceApi, axios);
+    return endpoint ? apiInstance[endpoint] : apiInstance;
+  }, [axios, endpoint]);
+
+  useEffect(() => {
+    if (endpoint && !apiEndpoint) {
```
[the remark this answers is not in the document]
bastilian: I'm not really sure what you mean. Do you mean to only check if an endpoint exists when running with `stage.foo.`?

src/Utilities/hooks/api/useComplianceQuery.js
bastilian: The `skip` option or any other `useQuery` option is not supported, it'll spread in the options now. The skipping before the serialised state is initialised is a measure to ensure only a request with params is made, other wise it would send a request without any params, and right after another one with params.

src/SmartComponents/SystemsTable/hooks.js
bastilian: Removed it.

bastilian:
@gkarat @mkholjuraev Thank you for reviewing! The comments should be addressed now, mind taking another look!

src/Utilities/hooks/api/useComplianceQuery.js
gkarat: I see
```
  const query = useQuery(apiEndpoint, {
    params: useTableState ? paramsFromState : params,
    skip: useTableState && !serialisedTableState,
    ...options,
  });
```
which would basically allow to override the `skip` parameter. But as you are saying, it's not supported

gkarat:
Looks good to me, but I want to clarify the situation with the initial skip (see the thread above).

src/Utilities/hooks/api/useComplianceQuery.js
bastilian: Oh, the "not supported" is a typo. Not sure how that happened. Any option passed to the `useComplianceQuery` hook will override whatever option it passes to `useQuery`, including the `skip` option. So an initial request is also skipped.

gkarat (approved):
LGTM, thanks for this huge improvement @bastilian. I have quickly verified the behaviour locally deployed, haven't spotted any issues.

PR #2234, opened 2024-11-01
feat(RHINENG-11549): Report detail - System Table migration

AsToNlele:
[RHINENG-11549](https://issues.redhat.com/browse/RHINENG-11549) and [RHINENG-10649](https://issues.redhat.com/browse/RHINENG-10649)

System table is divided into two tabs:

## Reporting table
Fetches data from `/reports/{report_id}/test_results`
These systems can be remediated

### Remediation
Fetches all rules for each test result from `/reports/{report_id}/test_results/{test_result_id}/rule_results`

## Never reported table
Fetches systems from `/reports/{report_id}/systems?filter=never_reported=true`

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/ComplianceRemediationButton/hooks.js
+++ b/src/PresentationalComponents/ComplianceRemediationButton/hooks.js
@@ -7,6 +7,91 @@ import {
   DEFAULT_CONNCURRENT_REQUESTS_FOR_ISSUES,
 } from './constants';
 import { remediationData } from './helpers';
+import { apiInstance } from '../../Utilities/hooks/useQuery';
+import pAll from 'p-all';
+
+export const useIssuesFetchRest = (reportId, testResults) => {
+  const { isResolving, results, resolve } = usePromiseQueue(
+    DEFAULT_CONNCURRENT_REQUESTS_FOR_ISSUES
+  );
+  const fetch = useCallback(async () => {
+    // Get rules of all test results
+    const results = await resolve(
+      [...testResults].map((tr) => () => batchFetch(tr, reportId, 100))
+    );
+
+    const flattenResults = results.flatMap((x) => x);
+
+    // Save memory by using a hash map
+    const hashMap = {};
+
+    flattenResults.forEach((res) => {
+      // Add each response to hash map
+      res.data.data.forEach((rule) => {
+        const remediation_issue_id = rule.remediation_issue_id;
+        const system_id = rule.system_id;
+        const result = rule.result;
+
+        if (!remediation_issue_id || result !== 'fail') return;
+
+        if (!Object.hasOwn(hashMap, remediation_issue_id)) {
+          hashMap[remediation_issue_id] = [system_id];
+        } else {
+          hashMap[remediation_issue_id] = [
+            ...hashMap[remediation_issue_id],
+            system_id,
+          ];
+        }
+      });
+    });
+
+    return hashMap && remediationDataRest(hashMap);
+  }, [reportId, testResults]);
+
+  const remediationDataRest = (hashMap) => ({
+    issues: Object.entries(hashMap).map(([remedIssue, systems]) => ({
+      id: remedIssue,
+      systems: systems,
+      description: undefined,
+    })),
+  });
+
+  const batchFetch = async (testResultId, reportId, limit = 100) => {
+    const firstRes = await buildFetch(testResultId, reportId, 0, limit);
+
+    const total = firstRes.data.meta.total || 0;
+
+    const neededRequests = Math.ceil(total / limit) - 1;
+
+    if (neededRequests == 0) {
+      return [firstRes];
+    }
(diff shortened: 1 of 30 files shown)


LightOfHeaven1994 (commented):
What I found when testing locally:

1. Export (without system selection) doesn't work, I see backend request is made correctly but actual download doesn't happen. Toast notifications are shown.
2. Sort test results by Name doesn't work as I see we try to sort it by `name` but it should be `display_name`
3. We are missing some filters for reporting tab - `Compliace (compliant, non-compliant, not supported)` ,  `compliance score`, `ssg_version`, `failed rule severity`
4. Filtering by tags doesn't work as it's not passed to backend

src/PresentationalComponents/ComplianceRemediationButton/hooks.js
```diff
@@ -7,6 +7,90 @@ import {
...
+import pAll from 'p-all';
+
+export const useIssuesFetchRest = (reportId, testResults) => {
+  const { isResolving, results, resolve } = usePromiseQueue(
+    DEFAULT_CONNCURRENT_REQUESTS_FOR_ISSUES
+  );
+  const fetch = useCallback(async () => {
+    // Get rules of all test results
+    const results = await resolve(
+      [...testResults].map((tr) => () => batchFetch(tr, reportId, 100))
```
mkholjuraev, line 20: Nitpick: it looks like batch fetching logic here may be replaced with the shared[useFetchBatched](https://github.com/RedHatInsights/frontend-components/blob/7b58162bb1e8d77a0b8c61f096c64f8cb4611a17/packages/utils/src/useFetchBatched/useFetchBatched.ts#L6)

src/PresentationalComponents/ComplianceRemediationButton/hooks.js
```diff
@@ -7,6 +7,90 @@ import {
...
+    );
+
+    const flattenResults = results.flatMap((x) => x);
```
mkholjuraev, line 23: Nitpick: I think what we need is `flat` rather `flatMap`.

src/PresentationalComponents/ComplianceRemediationButton/hooks.js
AsToNlele: uh oh

AsToNlele: I can't use a shared instance of the `usePromiseQueue` since each test result would need it's own. So I'll keep it like this.

mkholjuraev: hm, it seems I put my comment in confusing line. Rather, better line would have been on line [56](https://github.com/RedHatInsights/compliance-frontend/pull/2234/files#diff-0c27f43f2ed38d7255c266432ea7f865a124c3c0865002bc160ed72a9d14523d). What I wanted to point out is that using the shared hook, you would not need to recreate batch fetching logic from scratch. The shared hook would return callback function that can be invoked per your test results. There would not be any need to use `usePromiseQueue`. But, yes you have the hands-on experience with this case, feel free to ignore if what I say does not work out.

LightOfHeaven1994 (commented):
1. If I apply compliant filter and set non-compliant & not supported I get a bit weird filter string (see screenshot). I think it should be `(compliant=true OR supported=false)`

image

2. I was able to proceed with remediating unsupported system and it's visible on Remediate modal and from backend requests, We should ignore unsupported systems
image

other than that, looks good

adonispuente (commented):
After running locally and trying to break I wasnt to find any additional issues than already mentioned

src/constants.js
```diff
@@ -299,6 +337,23 @@ export const systemsDataMapper = {
...
+  id: 'id',
+  insights_id: 'insightsId',
+  tags: 'tags',
+  updated: 'updated',
+  score: ['complianceScore', 'score'],
+  failed_rule_count: 'rulesFailed',
+  security_guide_version: 'version',
+  supported: 'supported',
+  end_time: 'lastScanned',
+  system_id: 'system_id',
```
LightOfHeaven1994, line 354: We are missing `compliant` here

LightOfHeaven1994 (commented):
@AsToNlele filtering works fine now. We are missing 1 piece in data mapper and also Remediation still doesn't work as expected. I'm able to select unsupported system and remediate it.

LightOfHeaven1994 (approved):
Let's merge it and I will test it on stage properly, if I found something - I will create follow up cards. Don't want to hold this PR for a long time here

PR #2252, opened 2024-11-07
feat(RHINENG-13595): Migrate the rules table on the system details page

gkarat:
Implements https://issues.redhat.com/browse/RHINENG-13595.

This migrates the policy tabs and the rule (rule results) table on the system details view to use REST API endpoints.

Known issues (will be implemented/fixed in a separate PR):
- The tree table view is not yet implemented, only the rows table view must be available for now
- The Remediate button is kept disabled, we require items selection to be improved first

## How to test

Verify that all the reporting policies for the systems are reflected in a separate tab. The rules table should correctly filter, sort and paginate items. All the cells (name, policy, severity, rule state, remediation) must be correctly displayed. Also, check that the rule details are expandable.

## Screenshots

image

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/Frameworks/AsyncTableTools/hooks/useTableSort/useTableSort.js
```diff
@@ -34,9 +34,19 @@ import { TABLE_STATE_NAMESPACE } from './constants';
...
+    serialisers,
+    onSort: onSortOption,
+    detailsComponent,
+  } = options;
+  const hasExpandable = detailsComponent !== undefined;
   const serialiser = useCallback(
-    (state) => options.serialisers.sort(state, columns),
+    (state) =>
+      options.serialisers.sort(state, [
+        ...(hasExpandable ? [{}] : []),
```
bastilian, line 47: The issue this is fixing, should better be fixed in the `columnOffset` function to consistently get the proper index.

gkarat: Changed in the last commits

src/SmartComponents/SystemDetails/Columns.js
```diff
@@ -0,0 +1,54 @@
...
+import { renderComponent } from 'Utilities/helpers';
+import {
+  Rule,
+  Policy as PolicyCell,
+  Severity as SeverityCell,
+  Passed as PassedCell,
+  RemediationColumnCell,
+} from './Cells';
+
+export const Name = {
```
bastilian, line 11: Let's use the Columns and Cells from the `RulesTable` imported like the other `RulesTable`s do. It should not be difficult to prepare the columns and cells for the REST API response or map them via a `data mapper`. Some of the columns and cells are already prepared for the REST/rule object.

src/PresentationalComponents/ComplianceTable/serialisers.js
```diff
@@ -90,5 +93,6 @@ export const filtersSerialiser = (state, filters) => {
  *  ];
  *
  */
-export const sortSerialiser = ({ index, direction } = {}, columns) =>
-  columns[index]?.sortable && `${columns[index].sortable}:${direction}`;
+export const sortSerialiser = ({ index, direction } = {}, columns, options) =>
```
bastilian, line 96: Let's leave this out and try to fix the sorting in another separate PR and ensure the `useTableSort` hook returns always the expected index corresponding to the passed columns. This may fix the sorting under this context, but it will probably break it in a different one.

src/SmartComponents/SystemDetails/useTestResults.js
```diff
@@ -0,0 +1,60 @@
+import { useEffect, useState } from 'react';
+import useSystemReports from '../../Utilities/hooks/api/useSystemReports';
+import useReportTestResults from '../../Utilities/hooks/api/useReportTestResults';
+
+// custom hook to fetch test results for a system
+const useTestResults = (systemId) => {
+  const [testResultsLoading, setTestResultsLoading] = useState(true);
+  const [testResults, setTestResults] = useState(undefined);
```
bastilian, line 8: This is basically duplicating what the `useReportTestResults` would return via the `useQuery` hook with `data`, `loading` and `error`.

src/SmartComponents/SystemDetails/useTestResults.js
```diff
@@ -0,0 +1,60 @@
...
+        )
+      );
+      setTestResultsLoading(false);
+    };
+
+    if (reportsLoading === false && reports !== undefined) {
+      setTestResultsLoading(true);
+      collectTestResults();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bastilian, line 54: Why is this needed? What dependency is causing issues?

src/SmartComponents/SystemDetails/useTestResults.js
bastilian: Ah, I understand why.

src/PresentationalComponents/ComplianceTable/serialisers.js
gkarat: Reverted

src/SmartComponents/SystemDetails/useTestResults.js
gkarat: There is a problem adding one missing dependency _fetchTestResults_: the useEffect starts to loop. _fetchTestResults_ is not properly wrapped in useCallback somewhere in _useQuery_ most probably.

src/PresentationalComponents/RulesTable/RulesTableRest.js
```diff
@@ -116,28 +124,28 @@ const RulesTable = ({
         ...COMPLIANCE_TABLE_DEFAULTS,
         ...options,
         showViewToggle: true,
-        defaultTableView: 'tree',
+        defaultTableView,
         ...(ruleTree ? { tableTree: ruleTree } : {}),
-        identifier: itemIdentifier,
-        ...(onSelect ? { onSelect: onSelect || setSelectedRules } : {}),
+        ...(onSelect ? { onSelect: setSelectedRules } : {}),
```
bastilian, line 129: We should still pass on the `onSelect` if it is a function.

src/PresentationalComponents/RulesTable/RulesTableRest.js
```diff
@@ -103,10 +111,10 @@ const RulesTable = ({
           ansibleSupportFilter,
         }),
         ...(hidePassed && {
-          activeFilters: (currentActiveFilters) => ({
+          activeFilters: (currentActiveFilters = {}) => ({
             ...currentActiveFilters,
-            rulestate: currentActiveFilters.rulestate
-              ? currentActiveFilters.rulestate
+            ['rule-state']: currentActiveFilters['rule-state']
```
bastilian, line 116: It seems this default filter isn't showing up when the page is initially loaded.

src/SmartComponents/SystemDetails/useTestResults.js
bastilian: The fetch functions in `useQuery` are all wrapped in `useCallback`, and we are using them in some places as dependencies where they behave normally. There may be something else that

bastilian (approved):
LGTM! It is functional. The only things I noticed are that the rule names are not bold and coloured red as they are on the GQL version and that the "failed rules" filter doesn't seem to be applied properly by default.

However, I consider both these as non critical and we can fix those in a follow up, unless you want to fix them now in this PR.

PR #2262, opened 2024-11-11
fix(RHINENG-14293): fix PolicySystems export

LightOfHeaven1994:
Export systems without selection fails on Policy details - systems tab as we get undefined policy id - `
https://console.stage.redhat.com/api/compliance/v2/policies/undefined/systems?limit=50&offset=0`

Instead of using policy id from fetch arguments, we can provide it as a constant parameter.

To test PR - navigate to PolicyDetails page, open systems tab and try to export Systems

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


mkholjuraev:
@LightOfHeaven1994 with the provided testing instructions, I am not able to reproduce the issue. I am opening a policy detail page and navigating to systems tab. Without anything 'selected', I am trying to export into CSV or Json. Both cases, the export happens with available data. Let me know further details.

LightOfHeaven1994:
@mkholjuraev ensure you have APIv2 feature flag enabled

[Screencast from 2024-11-11 12-53-56.webm](https://github.com/user-attachments/assets/73a9bcac-6c0d-4495-8371-b0a5db298f94)

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
```diff
@@ -7,27 +7,9 @@ import * as Columns from '../SystemsTable/Columns';
...
-      offset,
-      null,
-      fetchArguments.sortBy,
-      fetchArguments.filter
-    )
-    .then(({ data: { data = [], meta = {} } = {} } = {}) => ({
-      data: dataSerialiser(data, systemsDataMapper),
-      meta,
-    }));
+import useFetchPolicySystems from './hooks/useFetchPolicySystems';
```
mkholjuraev, line 12: I intentionally avoided having a hook with some states to make an API call inside a component that uses InventoryTable component. My assumption was that those React states may cost rendering cycles. But, it is up to you to choose.

mkholjuraev (approved):
LGTM! This does solve the issue. However, it would be even better if we find out what `fetchArguments` does not have policy id. Because, not only export but also there may be other functions that depend on `policy.id` from `fetchArguments`.

src/SmartComponents/PolicyDetails/PolicySystemsTab.js
LightOfHeaven1994: @mkholjuraev I spent some time looking into it, and seems on main systems table request we update some ref values here https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/SystemsTableRest.js#L126-L130 like tags and filter that will affect export later. But we do not define here PolicyId ( or ReportId ), as the result we get it undefined when we try to call it here https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/hooks.js#L275

LightOfHeaven1994: @mkholjuraev I found a proper way to fix it and avoid hook usage, can you take a look?

mkholjuraev (commented):
small suggestion.

mkholjuraev:
The tests are failing due to commit lint. However, I think we are safe to merge as the change is minimum enough that does not affect tests.

LightOfHeaven1994:
> The tests are failing due to commit lint. However, I think we are safe to merge as the change is minimum enough that does not affect tests.

thanks a lot for the review and great suggestion :)

PR #2264, opened 2024-11-12
fix(RHINENG-14334): System detail shows empty GridItem

johnsonm325:
There is the case when you have your system assigned to multiple policies but it's not reported against each policy - we will still try to show a card (you can see an GridItem - we show there an empty component). This is happening because we have multiple reports returned but some of them have 0 test results ( as system is never reported)

To repro:
- Go to Compliance -> Systems
- Find a system with multiple policies but at least one policy with 0   test results (data.meta.total === 0)

Before, the GridItem would render, but with no card.

image

Now, the GridItem has been moved into the child component and will only display the GridItem if the meta.total > 0.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemPolicyCards/SystemPolicyCards.js
+++ b/src/SmartComponents/SystemPolicyCards/SystemPolicyCards.js
@@ -24,15 +24,19 @@ const SystemPolicyCard = ({ policy }) => {
     ],
   });
 
-  return data === undefined ? (
-    <LoadingPolicyCards count={1} />
-  ) : data.meta.total === 0 ? (
-    <React.Fragment />
-  ) : (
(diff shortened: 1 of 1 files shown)


LightOfHeaven1994 (approved):
Issue is fixed, thanks a lot!

AsToNlele (approved):
one from me as well 😄 

the `||` is quite unusual in rendering but makes sense

PR #2265, opened 2024-11-13
feat(RHINENG-14339): Unite asynctables & v2 FF

LightOfHeaven1994:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/ComplianceTable/ComplianceTable.js
+++ b/src/PresentationalComponents/ComplianceTable/ComplianceTable.js
@@ -1,8 +1,10 @@
 import React from 'react';
+import { Spinner, Bullseye } from '@patternfly/react-core';
+
 import propTypes from 'prop-types';
 import AsyncTableToolsTable from '@/Frameworks/AsyncTableTools/components/AsyncTableToolsTable';
 import { TableToolsTable } from 'Utilities/hooks/useTableTools';
-import { ENABLE_ASYNC_TABLE_HOOKS } from '@/constants';
+import useAPIV2FeatureFlag from '@/Utilities/hooks/useAPIV2FeatureFlag';
 import {
   paginationSerialiser,
   filtersSerialiser,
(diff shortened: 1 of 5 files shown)


AsToNlele (commented):
this needs to wait for the policies async table migration, since enabling the async table would break the non async policies table

LightOfHeaven1994:
> this needs to wait for the policies async table migration, since enabling the async table would break the non async policies table

sure, we have 1 more table that is not migrated - policy types table (create policy wizard first step)

LightOfHeaven1994:
@AsToNlele since your async Policies PR has been merged, I believe this one is also ready to go. We will implement async tale for policy types (wizard 1st step) later as it's not high priority for now

src/PresentationalComponents/ComplianceTable/ComplianceTable.js
```diff
@@ -31,13 +31,13 @@ const ComplianceTable = (props) => {
   const tableState = useRawTableState();
 
   useDeepCompareEffectNoCheck(() => {
-    if (ENABLE_ASYNC_TABLE_HOOKS) {
+    if (useAPIV2FeatureFlag) {
```
gkarat, line 34: We have to call the hook and use the returned value as a feature flag value.

```suggestion
   const apiV2Enabled = useAPIV2FeatureFalg();
   ...
   
    if (apiV2Enabled) {
```

src/PresentationalComponents/ComplianceTable/ComplianceTable.js
```diff
@@ -31,13 +31,13 @@ const ComplianceTable = (props) => {
...
       console.log('Async Table enabled');
       console.log('TableState', tableState, serialisedTableState);
     }
   }, [tableState, serialisedTableState]);
 
-  return ENABLE_ASYNC_TABLE_HOOKS ? (
+  return useAPIV2FeatureFlag ? (
```
gkarat, line 40: Since we have replaced the constant with the feature flag hook call here, let's also cover the undefined case so that we don't unnecessarily render TableToolsTable.

Like this

```suggestion
  return apiV2Enabled === undefined ? render bullseye + spinner : apiV2Enabled === true ? render async table : render TableToolsTable
```

LightOfHeaven1994: @gkarat yep, sounds good, fixed

LightOfHeaven1994:
I was looking into test failures, some of them are valid and they were not failing because async tables feature was probably disabled when we were running tests. But some of these things might take time to fix

AsToNlele (approved):
Went through the app, other API v2 tables don't use the non async table anymore so all good.

Thank you! 😄

PR #2268, opened 2024-11-13
fix(rhineng-14291): fix source for view policy button id

Fewwy:
https://issues.redhat.com/browse/RHINENG-14291


LightOfHeaven1994 (approved):
LGTM, checked with disabled & enabled feature flags

PR #2287, opened 2024-11-22
fix(rulesTable): Allow rules table to open multiple rule

adonispuente:
Rules Table was unable to open more than 2-4 rows without crashing. This just adds little checks and prevents the table from crashing. 

To test:
navigate to any rules table, rows view and open up multiple.
Rules table in policy details 
Rules table in default_ruleset reached from creatPolicy->rules step-> default rules link


LightOfHeaven1994 (approved):
LGTM, thanks!!

PR #2301, opened 2024-12-02
fix(RHINENG-14527): Enable value editing when creating a SCAP policy

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-14527.

This fixes the create policy wizard: when overriding at least one rule values, the wizard should call PATCH request additionally to update the value overrides.

## How to test

Try to create a policy and find the rules which has some value to override. Set any value and follow up to the finish step. When submitting the policy creation, make sure the policy creation is complete. Follow to the policy details page and verify that the rules has all the value overrides you set in the wizard.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTableRest.js
+++ b/src/PresentationalComponents/RulesTable/RulesTableRest.js
@@ -56,7 +56,7 @@ const RulesTable = ({
   );
 
   const DetailsRow = useMemo(() => {
-    const Row = (props) => {
+    function Row(props) {
       // eslint-disable-next-line react/prop-types
       const { itemId, valueDefinitions } = props?.item || {};
 
@@ -91,7 +91,7 @@ const RulesTable = ({
           item={item}
         />
       );
-    };
+    }
 
     return Row;
   }, [
--- a/src/PresentationalComponents/RulesTable/RulesTableRest.js
+++ b/src/PresentationalComponents/RulesTable/RulesTableRest.js
@@ -6,6 +6,7 @@ import { ComplianceTable } from 'PresentationalComponents';
 import RuleDetailsRow from './RuleDetailsRow';
 import buildFilterConfig from './Filters';
 import defaultColumns from './Columns';
+import { List } from 'react-content-loader';
 
 const RulesTable = ({
   system,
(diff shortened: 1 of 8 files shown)


bastilian:
@gkarat It looks like there is a small issue when trying to save the value. When I try to edit the "SSH Max authentication attempts" in the "Set SSH authentication attempt limit" rule clicking on the save button raises an error in the `setUpdatedPolicy` function with "prev is null".

bastilian:
I patched up the `setUpdatedPolicy` to fix the null error, but it also seems that the values are not yet saved properly when the save of the edit modal is pressed.

bastilian:
🤦‍♂️ I tested the edit policy even though the title clearly says "Create Policy". :D

bastilian:
I now tested the "Create Policy" and there seems to be an issue with the value editing. I tried editing the same rule in the policy wizard and when clicking the ✔️, the spinner appears, but nothing seems to happen.

LightOfHeaven1994:
We are still not showing default values

image

Also some values can't be set on create policy wizard. I got this spinner and console error.
Screenshot 2024-12-05 at 9 59 47

AsToNlele:
Default value is fixed here: https://github.com/RedHatInsights/compliance-frontend/pull/2307

src/SmartComponents/CreatePolicy/FinishedCreatePolicy/FinishedCreatePolicyRest.js
```diff
@@ -95,9 +99,35 @@ export const useUpdatePolicy = () => {
         dispatchProgress();
       });
 
+      tailorings.forEach(async (tailoring) => {
```
bastilian, line 102: Watch out: "forEach() expects a synchronous function — it does not wait for promises. Make sure you are aware of the implications while using promises (or async functions) as forEach callbacks."[0]

It would be better to prefer a proper for loop here or a `Promise.all`/`Promise.allSettled`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/PresentationalComponents/RulesTable/RulesTableRest.js
```diff
@@ -54,54 +55,63 @@ const RulesTable = ({
     />
   );
 
-  const DetailsRow = useMemo(
-    () =>
-      function Row(props) {
```
bastilian, line 59: The way this was written with `function` is on purpose in order to not have an anonymous function.

bastilian: To elaborate a bit more. The function component function will end up in the options for the table hook(s). While `JSON.stringify` removes any such entry, when we use `useDeepCompareEffect`, it will somewhat/somehow use the function name to assess whether or not the value is different, it won't be able to do that with an arrow/anonymous function. This means change it might cause in some situations that a deep compare effect to fire unexpectedly.

bastilian (approved):
LGTM! Nice work! Thank you @gkarat!

PR #2314, opened 2024-12-18
fix(Tailorings): RHINENG-14520 - Fix export for wizard

bastilian:
This fixes the export for the Rules step in the "Create Policy" wizard.

**How to test:**

1) Open the "Create Policy" wizard and advance to the Rules step
2) Try exporting the rules as JSON or CSV
3) Verify the export has been downloaded and the items inside equal the rules in the table
4) Test exporting with and without filters applied

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/Frameworks/AsyncTableTools/utils/withExport/withExport.js
+++ b/src/Frameworks/AsyncTableTools/utils/withExport/withExport.js
@@ -42,6 +42,7 @@ const withExport = ({
       downloadItems(exportColumns, items, format);
       onComplete?.(items);
     } catch (error) {
+      console.log(error);
       onError?.(error);
     }
   };
--- a/src/Frameworks/AsyncTableTools/utils/withExport/withExport.js
+++ b/src/Frameworks/AsyncTableTools/utils/withExport/withExport.js
@@ -42,7 +42,7 @@ const withExport = ({
       downloadItems(exportColumns, items, format);
       onComplete?.(items);
     } catch (error) {
-      console.log(error);
+      console.error(error);
       onError?.(error);
     }
   };
--- a/src/PresentationalComponents/Tailorings/components/TailoringTab.js
+++ b/src/PresentationalComponents/Tailorings/components/TailoringTab.js
@@ -86,9 +86,11 @@ const TailoringTab = ({
         )
       : undefined;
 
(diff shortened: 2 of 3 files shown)


src/Frameworks/AsyncTableTools/utils/withExport/withExport.js
```diff
@@ -42,6 +42,7 @@ const withExport = ({
       downloadItems(exportColumns, items, format);
       onComplete?.(items);
     } catch (error) {
+      console.log(error);
```
gkarat, line 45:
```suggestion
      console.error(error);
```

gkarat (commented):
> Verify the export has been downloaded and the items inside equal the rules in the table

I had 74 rules selected in the beginning and tried to show all of the available rules (not only the selected ones), and did the export as CSV. The end result contained only those that are selected which goes slightly against the expected results (to export all the items in the table). I have tried to select one more item (having 75 rules selected in total), but the exported CSV again contained only 74 initial rules.

gkarat:
Adding filters (like Name) also doesn't change the exported data.

LightOfHeaven1994 (commented):
@bastilian might be a bit outside of context but I think we might fix it here. Export still fails on PolicyDetails rules tab & Edit Policy modal with:
```
TypeError: (intermediate value)(intermediate value)(intermediate value).flatMap is not a function
    at exporter (TailoringTab.js:93:7)
    at async exportWithFormat (withExport.js:41:21)
```

bastilian:
@LightOfHeaven1994 That should be fixed now too.

I added note, because having to wrap the results is a bit odd, but we can look into cleaning that up later.

LightOfHeaven1994 (approved):
LGTM, thank you!

LightOfHeaven1994:
pr check failure is not related to FE, I fixed it once on our side with workaround but Compliance still behaves weirdly with Cyndi. so ignore it

PR #2330, opened 2025-01-09
fix(Create/EditPolicy/PolicyDetails): Add missing system number badges

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-14518.

This returns the systems count to the create policy, edit policy and policy details views.

## Screenshots

Screenshot 2025-01-10 at 11 02 08

Screenshot 2025-01-10 at 11 02 41

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/PresentationalComponents/Tailorings/Tailorings.js
```diff
@@ -199,12 +201,23 @@ import NoTailorings from './NoTailorings';
...
+                        profile={{
+                          osMajorVersion: tab.os_major_version,
+                          osMinorVersion: tab.os_minor_version,
+                        }}
+                      />
+                    </FlexItem>
+                    {selectedVersionCounts && (
+                      <FlexItem>
+                        <Badge isRead>
+                          {selectedVersionCounts[tab.os_minor_version]}
```
LightOfHeaven1994, line 216: I think we need to default here to `0` if no value provided. Otherwise we will see an empty space:
image

src/Utilities/hooks/api/usePolicyOsVersionCounts.js
```diff
@@ -0,0 +1,48 @@
...
+      try {
+        const newCounts = {};
+        const versions = await fetchPolicySystemsOS([policyId], false);
+
+        for (const version of versions) {
+          const minor = version.split('.')[1];
+          const policySystemsResponse = await fetchPolicySystems(
+            {
+              policyId,
+              offset: 1,
```
LightOfHeaven1994, line 25: I think it should be `limit: 1` instead of `offset: 1`

src/Utilities/hooks/api/usePolicySystems.js
```diff
@@ -0,0 +1,25 @@
...
+      sortBy,
+      filter,
+    ];
+  }
+};
+
+const usePolicySystems = (options) =>
+  useComplianceQuery('policySystems', { ...options, convertToArray });
+
+export default usePolicySystems;
```
LightOfHeaven1994, line 25: It might be a bit outside of this PR, but we can use this helper here instead of https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/PolicyDetails/PolicySystemsTab.js#L15-L30

LightOfHeaven1994 (commented):
@gkarat I have found some things to fix, but other than that looks good

src/Utilities/hooks/api/usePolicySystems.js
gkarat: This is something we actually should replace with the usePolicySystems hook as direct apiInstance call implements the network fetch function independently from useComplianceQuery. This helps to stay consistent with other /hooks/api hooks.

PR #2356, opened 2025-01-24
fix(RHINENG-15120): Fix rule patching in policies

adonispuente:
There is an issue when altering the value of a rule, where 1, it does not save, and 2, when updating one and opening another, the first gets reset.

To reproduce: 

- Navigate to Create policy wizard
- Go to Rules step
- Find any rule that has value (search for "password" or "ssh" and most of these rules have values)
- Expand rule and patch value
- Expand rule that goes after patched one
- **Old Behavior** Observe that value of patched rule has been changed to default one (we still store patched values and will apply them, so bug is only visual)
- **New Behavior** Observe that the value is saved. Note that when alter rules, many times the valueCheckIds for the rules are the same, so alter one may alter all the following rules.

Observe everywhere that the value is correct for the rule.
The policy needs to be monitored in 3 places. Create Policy Wizard, Policy Details in the rules Tab, and edit policy in system details.

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTableRest.js
+++ b/src/PresentationalComponents/RulesTable/RulesTableRest.js
@@ -35,6 +35,7 @@ const RulesTable = ({
   onSelect,
   defaultTableView = 'tree',
   reportTestResult,
+  valueOverrides,
   ...rulesTableProps
 }) => {
   const internalSelectedState = useState([]);
@@ -77,6 +78,7 @@ const RulesTable = ({
         ...rule,
         itemId,
         valueDefinitions: ruleValueDefinitions,
+        valueOverrides,
         profile: { id: policyId, name: policyName },
         ruleValues: ruleRuleValues,
       };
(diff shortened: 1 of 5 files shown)


src/PresentationalComponents/RulesTable/RulesTableRest.js
```diff
@@ -98,23 +100,54 @@ const RulesTable = ({
...
   ]);
 
   const itemsWithValueDefinitions = useMemo(
-    () => rules?.map((rule) => ({ ...rule, valueDefinitions })),
-    [rules, valueDefinitions]
+    () =>
+      rules?.map((rule) => {
+        const updatedRule = { ...rule };
+
+        const updatedValueDefinitions = valueDefinitions?.data?.map(
```
bastilian, line 118: This is maybe not right here, but we can revisit this later.

I think we are handling "updated values" in a couple of different ways, but we should only provided them to the table within the rules themselves, not as an extra object. The RulesTable shouldn't need to "prepare" the rules to show, but only concerned with showing them.

src/PresentationalComponents/RulesTable/RulesTableRest.js
```diff
@@ -98,23 +100,54 @@ const RulesTable = ({
...
     return Row;
   }, [
-    policyId,
-    policyName,
-    onRuleValueReset,
     rules,
     ruleValues,
+    policyId,
+    policyName,
+    skipValueDefinitions,
```
bastilian, line 107: `skipValueDefinitions` is not used, is it?

bastilian (approved):
LGTM! Thank you for fixing this @adonispuente!

PR #2363, opened 2025-01-28
fix(sentry): Update sentry flow

adonispuente:
This also removes the issue where the plugin is initialized locally and sends up debugIDs for local env on every build
**Workflow Behavior:**

**Locally (without variables):**
The Sentry plugin won't build as SENTRY_AUTH_TOKEN is not defined.

**GitHub push to master (automatic trigger):**
The plugin won't include the authToken because commit_hash is not provided in push events.

**Manually triggered with commit_hash:**
The plugin includes the authToken and tracks the build with Sentry.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,12 +1,15 @@
-name: sentryConfig
+name: sentryInit
 
 on:
+  push:
+    branches:
+      - master
   workflow_dispatch:
     inputs:
       commit_hash:
         description: 'The commit hash (or branch/tag) to build'
-        required: true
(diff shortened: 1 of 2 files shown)


bastilian:
@adonispuente Something we should maybe note for the future, when we integrate this elsewhere to also make things like `dsn`, `project` `org` etc. also configurable via environment variables and not have them hard coded or expose them directly on GitHub. 

While I don't think that these pose any risk when exposed, better safe than sorry. :D

bastilian (approved):
LGTM! Thank you @adonispuente!

PR #2364, opened 2025-01-29
fix(RHINENG-15635): Set system minVer correctly on search&select

LightOfHeaven1994:
I was looking into Sentry issues and found out that it's possible that when system is selected on Edit policy modal - minor version is `undefined` and as result we get non-working rules tab. I also found minor issue with count labels.

To test:
1. Have Policy with minor version X assigned (let's say RHEL8 policy with system 8.6)
2. Go to edit policy modal and open systems tab
3. Search system by name with minor version Y (let's say you have system test.foo 8.4)
4. Select system
5. Open Rules tab and observe that there are 2 tabs: 8.6 & 8.4 and labels are set correctly

Before:
we see error in console `TypeError: Cannot read properties of undefined (reading 'id')` and v2 request is sent with undefined minor version:
`/api/compliance/v2/security_guides?limit=1&ids_only=true&sort_by=version:desc&filter=os_major_version=8 AND supported_profile=xccdf_org.ssgproject.content_profile_cis_server_l1:undefined`

https://github.com/user-attachments/assets/a2439223-7ad0-46c4-872e-e0bddbf5ccfc

After: 
everything works as it should 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicyFormRest.js
+++ b/src/SmartComponents/EditPolicy/EditPolicyFormRest.js
@@ -42,7 +42,11 @@ const EditPolicyForm = ({
   const handleSystemSelect = useCallback(
     (newSelectedSystems) => {
       const newOsMinorVersions = [
-        ...new Set(newSelectedSystems.map((system) => system.osMinorVersion)), // get unique values
+        ...new Set(
+          newSelectedSystems.map(
+            (system) => system.os_minor_version ?? system.osMinorVersion
+          )
+        ), // get unique values
       ];
 
       const hasNewOsMinorVersions =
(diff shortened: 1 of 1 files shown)


src/SmartComponents/EditPolicy/EditPolicyFormRest.js
```diff
@@ -42,7 +42,11 @@ const EditPolicyForm = ({
   const handleSystemSelect = useCallback(
     (newSelectedSystems) => {
       const newOsMinorVersions = [
-        ...new Set(newSelectedSystems.map((system) => system.osMinorVersion)), // get unique values
+        ...new Set(
+          newSelectedSystems.map(
+            (system) => system.os_minor_version ?? system.osMinorVersion
```
bastilian, line 47: I think we can now start to unify all attribute names on what comes from API v2, or? Meaning we can just remove/replace old GQL attribute names with the underscore versions of attribute names.

LightOfHeaven1994: @bastilian I agree, so we don't have these 2 types. But at the same time it's going to be a huge effort to refactor all of pieces. So maybe we can have this small patch so far, since it affects customers, wdyt?

PR #2405, opened 2025-03-03
fix(RHINENG-16323): Make value update work EditPolicy

LightOfHeaven1994:
This PR fixes rule values update on EditPolicy modal.

**Before the fix:**
If you try to update value - page reloads and on second try page is going to crash

**After the fix:**
Value changes are respected and on Save button click PATCH request is sent for tailoring.

**How to test:**

1. Have freshly created policy with 1 assigned host with minor version X
2. Navigate to Edit policy rules tab
3. Find any rule that has value to edit (hint: search for "selinux" or "passord")
4. Use inline edit to update a value to something that you will remember (like classic "foo")
5. Do not save changes, open systems tab and add 1 more system but with minor version Y (different from first one)
6. Open rules tab back and observe 2 minor version tabs
7. Check if tab with X still has value that you set
8. Try to change a value for tab Y for any rule
9. Save policy
10. Navigate to policy details and check if each tab has rule with patched value

**Since I refactored a bit other calls ensure value update also works from:**

1. Create policy wizard
11. Policy details Rules tab

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -75,7 +75,7 @@ const EditPolicy = ({ route }) => {
     setUpdatedPolicy((prev) => ({
       ...prev,
       value: {
-        ...prev.value,
+        ...prev?.value,
         [tailoring.id]: {
           ...tailoring.value_overrides,
           [valueDefinition.id]: newValue,
--- a/src/SmartComponents/EditPolicy/EditPolicy.js
+++ b/src/SmartComponents/EditPolicy/EditPolicy.js
@@ -72,16 +72,18 @@ const EditPolicy = ({ route }) => {
     newValue,
     closeInlineEdit
   ) => {
-    setUpdatedPolicy((prev) => ({
-      ...prev,
-      value: {
-        ...prev?.value,
-        [tailoring.id]: {
-          ...tailoring.value_overrides,
-          [valueDefinition.id]: newValue,
+    setUpdatedPolicy((prev) => {
+      return {
+        ...prev,
+        tailoringValueOverrides: {
+          ...prev?.tailoringValueOverrides,
+          [tailoring.id]: {
+            ...tailoring.value_overrides,
+            [valueDefinition.id]: newValue,
+          },
         },
-      },
-    }));
+      };
+    });
 
     closeInlineEdit();
   };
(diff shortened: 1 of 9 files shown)


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -75,7 +75,7 @@ const EditPolicy = ({ route }) => {
     setUpdatedPolicy((prev) => ({
       ...prev,
       value: {
-        ...prev.value,
+        ...prev?.value,
```
LightOfHeaven1994, line 78: Stage is failing exactly at this place with `TypeError: Cannot read properties of null (reading 'value')` as if there are no prev value, we can't call it like this

bastilian: I'm wondering why/when the previous value would be `null`. Even if the value wasn't changed before, shouldn't each value at least have its default initially?

LightOfHeaven1994: @bastilian each value is defined under tailoring on line 80 here and I believe prev is used when something changes after modal has been open

src/SmartComponents/EditPolicy/EditPolicyRulesTab.js
```diff
@@ -63,31 +66,54 @@ export const EditPolicyRulesTab = ({
...
-            },
-          }));
+  useDeepCompareEffect(() => {
+    setUpdatedPolicy((prev) => {
+      return {
+        ...prev,
+        tailoringRules: profilesRuleIds?.reduce((prevRuleIds, profile) => {
+          return {
+            ...prevRuleIds,
+            [Number(profile.osMinorVersion)]: profile.ruleIds,
```
LightOfHeaven1994, line 83: I just found out that we need the same thing here as we did for values - as you might configure profile for minor ver X and jump on systems tab and add minor version Y, if you get back to rules tab - your selection for X will disappear. I will add it tomorrow

LightOfHeaven1994: I spent couple hours with it and it requires some time to properly investigate and fix it. It's not critical issue and it was here before my PR. Issue is that selection might not be respected if you select/unselect rules and switch between tabs. I will create a separate JIRA and want to keep it outside of this PR

bastilian (approved):
Awesome! Thank you very much @LightOfHeaven1994!

PR #2413, opened 2025-03-07
fix(SystemsTable): RHINENG-14946 - Properly fetch systems on selection

bastilian:
This PR implements fetching system details properly when systems are selected in the Edit Policy Systems tab.

**How to test:**

1) Open the Edit Policy modal for a policy
2) Navigate to the Systems tab
3) Filter systems by specific OS versions (make sure to exceed the "perPage")
4) Use the bulk select dropdown and "Select All (x)"
5) Verify that the selection count is increased properly

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/hooks.js
+++ b/src/SmartComponents/SystemsTable/hooks.js
@@ -1,5 +1,6 @@
 import React, { useEffect, useLayoutEffect, useCallback } from 'react';
 import { useDispatch } from 'react-redux';
+import pAll from 'p-all';
 import { Spinner } from '@patternfly/react-core';
 import debounce from '@redhat-cloud-services/frontend-components-utilities/debounce';
 import {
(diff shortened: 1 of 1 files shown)


src/SmartComponents/SystemsTable/hooks.js
```diff
@@ -275,10 +295,15 @@ export const useSystemBulkSelect = ({
...
+    const systemsToFetch = selectedIds.filter((id) =>
+      loadedItems.map(({ id }) => id).includes(id)
     );
-    onSelect && onSelect(systemsSelection);
+
+    const fetchedSystems = (await fetchSystemsById(systemsToFetch)).map(
+      ({ data }) => data
+    );
+
+    onSelect && onSelect([...loadedItems, ...fetchedSystems]);
```
gkarat, line 306: This makes all items on the page to be selected if you check just one system. I believe we need to rewrite it to
```suggestion
    onSelect && onSelect(fetchedSystems);
```

gkarat (changes requested):

src/SmartComponents/SystemsTable/hooks.js
bastilian: Should be fixed now. Only taking the fetchedSystems would then in other conditions cause the items on the current page not being properly added. We do need the `lodedItems`, but we need to filter them be the ids that have actually been selected.

bastilian: Honestly, I'm not even sure why we do it this complicated. As far as I can tell the batch request seems to have the same data that a separate request for each system returns. And the way this is implemented doesn't seem right. The return and the call to a state update are a 🚩.

gkarat (approved):
I have tested same scenarios, under different filters and other params and now it works as intended. Thanks @bastilian !

PR #2435, opened 2025-03-18
fix(RHINENG-16777): Fix policy inline edit

LightOfHeaven1994:
#### This PR deals with a couple of things:
1. Error when updating policy data (because refetch wasn't passed correctly)
2. No more re-render and api request when edit is cancelled
3. Styling fix for textArea (now edit description area is a bit bigger)
4. Test coverage

### How to test:
1. Navigate to policy Details tab
2. Try to update threshold, business obj & description ( see that it works )
3. Try to start editing and then cancel edit (no re-rendering happens, no api request made)

### DIFF
--- a/src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
+++ b/src/PresentationalComponents/PolicyDetailsDescription/PolicyDetailsDescription.js
@@ -68,7 +68,7 @@ const PolicyDetailsDescription = ({ policy, refetch }) => {
               Component={TextArea}
               text={descriptionText}
               variant="description"
-              inlineClosedText={businessText}
+              inlineClosedText={descriptionText}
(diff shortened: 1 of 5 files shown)


bastilian:
@LightOfHeaven1994 Looks good and updating seems to work, but we will need to pass through the `refetch` function from the `useQuery` hook in the `useComplianceQuery` hook to make the updating of the policy details, that uses the `usePolicy` hook actually refetch.

LightOfHeaven1994:
> @LightOfHeaven1994 Looks good and updating seems to work, but we will need to pass through the `refetch` function from the `useQuery` hook in the `useComplianceQuery` hook to make the updating of the policy details, that uses the `usePolicy` hook actually refetch.

Good note, fixed 👍

PR #2436, opened 2025-03-18
feat(RHIENG-16658): Change Remediate button label to Plan remediation

adonispuente:
https://issues.redhat.com/browse/RHINENG-16658
In accordance with the above ticket, this changes the text on the remediations button to say 'plan remediation' instead of 'remediate'

In compliance the button can be found in report details page


LightOfHeaven1994 (changes requested):
@adonispuente We need to change text on FallbackButton too [here](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/ComplianceRemediationButton/components/FallBackButton.js#L5), otherwise it will look a bit odd when Button is loading (it firstly shows `Remediate` but when it's loaded it switches to `Plan remediation`

adonispuente:
good catch @LightOfHeaven1994 !

LightOfHeaven1994 (approved):

PR #2441, opened 2025-03-26
fix(RHINENG-16897): Allow saving only when data loaded

LightOfHeaven1994:
So to fix it, we need to wait for all data being loaded initially + handle specifically the case when bulk select is used and we are waiting on systems data. We need to disable save button. Since there might be a lot of systems and waiting time can be ~10 seconds or so, I added Save loading condition so it will be more visible that something is happening.

### How to test:

1. Navigate to Edit policy modal
2. Observe that when modal is open - Save button is disabled until we have data
3. Open systems tab
4. Use bulk select to select all systems (it's better to use account that has over 100 systems)
5. Observe that when data are loading, button keeps being disabled (+ spinner on button is shown)

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/EditPolicy/EditPolicy.js
```diff
@@ -52,8 +52,15 @@ const EditPolicy = ({ route }) => {
   );
 
   const [updatedPolicy, setUpdatedPolicy] = useState(null);
+  const [systemsDataLoading, setIsSystemsDataLoading] = useState(false);
 
-  const saveEnabled = !updatedPolicy;
+  const saveDisabled =
```
bastilian, line 57: Nice touch, with flipping the meaning to make it easier to comprehent! I like. :)

bastilian (approved):
Nice! Thank you for fixing this @LightOfHeaven1994!

PR #2451, opened 2025-03-31
fix(useBulkSelect): Fix issue with changed selection

bastilian:
Fix issues when a rules selection in the edit policy was changed, and later overwritten by loading additional profile rules.

**How to test:**

1) Open the Edit Policy Modal
2) Change the selection of rules for a tailoring and remember the changes
3) Add additional systems with a different minor OS version
4) Verify on the Rules tab that the changes made previously are still preserved. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/PresentationalComponents/Tailorings/components/TailoringTab.js
```diff
@@ -190,7 +190,7 @@ const TailoringTab = ({
 
   const onSelectRule = (...ruleParams) =>
     onSelect?.(
-      tailoring || { ...securityGuide.data, os_minor_version: osMinorVersion },
+      tailoring || { ...securityGuide?.data, os_minor_version: osMinorVersion },
```
bastilian, line 193: Still not sure about this boi. It doesn't seem to break anything, but it also doesn't feel correct...

LightOfHeaven1994:
@bastilian change looks great and fixed the problem. Test failures are valid, I checked report details page and it currently fails with 
```
TypeError: selectedRules is undefined
    remediationAction RulesTable.js:87
```

bastilian:
@LightOfHeaven1994 Thank you for testing! The test failures should now be fixed as well.

src/PresentationalComponents/Tailorings/components/TailoringTab.js
gkarat: If tailoring is not available, then it's supposed to use security guide data after the SG ID is provided. I feel this is not entirely good to have this optional chaining. Though, this would be bad only for one scenario: the SG data is still not loaded but user is attempting to select some rule. If this can't happen, then we can leave it IMO

gkarat (approved):
Tested, works, code-wise looks good. Just one comment. Thanks @bastilian

bastilian:
@LightOfHeaven1994 @gkarat While this fixes the issue, I'm still trying to figure out that one change with the SSG and why it is undefined. I believe there may be an issue lurking that we may not see yet.

bastilian:
@LightOfHeaven1994 @gkarat I did some investigating and the reason that security guide is `undefined` is because the components get unmounted and rendered again, which then has an initial state with the SSG `undefined`. 

I will still merge this, but we should think about refactoring the "Edit Policy", there are a few things that are kept in multiple states and it's difficult to keep all of them properly in sync. The refactor should focus on ensuring that there is only one source of truth for the policy data. There is also inconsistent naming of things. There are "assignedRules", "selectedRules", "selectedRuleRefIds", etc. all basically referring to the same thing.

PR #2823, opened 2026-05-07
feat(RHINENG-24154): change Review step labels to links

opacut:
Based on the UX review we want to change the Review Step labels to links in Create SCAP Policy wizard.

How to test:

1. Have insights-chrome and insert into webpack config
```
'/apps/compliance': {
  host: 'http://localhost:8004/',
},
'/apps/inventory': {
  host: 'http://localhost:8005/',
}, 
```
2. Run insights-chrome with `npm run dev`
3. Run insights-inventory-frontend with `npm run static -- -p=8005`
4. Checkout to this PR and run `npm run static -- -p=8004`
5. Navigate to Compliance -> Policies -> Create new policy
6. choose a RHEL version and a policy
7. click through to the Systems step
8. choose a system
9. click through to the Review step
10. check that the information reflect the mocks

image

Closes https://redhat.atlassian.net/browse/RHINENG-24154

## Summary by Sourcery

Update the SCAP policy creation review step to make the systems count label act as a navigational link back to the Systems step.

New Features:
- Allow users to click the systems count in the Review step to jump directly back to the Systems step of the wizard.

Enhancements:
- Adjust the systems count label styling and pluralization to reflect the selected systems more clearly.


LightOfHeaven1994 (commented):
@opacut I feel we miss some good UX input here. I see we have it in mocks [here](https://www.figma.com/design/r97CXZ3uUrw06Lhr0WitHl/%F0%9F%92%BBCompliance-Mockups-v2?node-id=1229-2363&t=p6ATjtQl6QBajZlP-0) but it looks a bit odd in mocks. Especially if card says it should be links.

I'm good with what we have currently, as it's looking fine and clear (comparing to currentl mocks).
With your change - it's not really intuitive that label can be clicked and it's a link (I know it might be to different to each of us). But it's against patternfly guidance I believe. Can you confirm with UX how this need to look like?

src/SmartComponents/CreatePolicy/components/ReviewStep.js
```diff
@@ -9,6 +9,10 @@ const ReviewStep = () => {
   const profile = values.profile;
   const osMajorVersion = profile?.os_major_version;
   const osMinorVersionCounts = values.osMinorVersionCounts || [];
+  const totalSystemsCount = osMinorVersionCounts.reduce(
```
LightOfHeaven1994: nitpick: we could use systems from `values` like:
`const totalSystemsCount = (values.systems || []).length;` and then we won't need to check below if totalSystemsCount > 0 on line 48
values.systems are coming from systems step in `onSelect` https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/CreatePolicy/components/SystemsStep.js#L88

LightOfHeaven1994 (approved):
@opacut looks good, I left 1 comment

LightOfHeaven1994 (approved):

PR #2849, opened 2026-05-29
feat(RHINENG-24161): align PolicyTypeTable cells vertically

opacut:
Aligns table cells vertically to the middle in the PolicyTypeTable component by wrapping the table in a div with an inline style block.

- wrapped `TableToolsTable` in a `<div className="policy-type-table">`
- added inline `<style>` block to vertically align radio button cells
- added `tableBodyProps={{ style: { verticalAlign: 'middle' } }}` to vertically align text cells

## How to test
1. Set up local environment according to readme
2. Navigate to Create new policy modal
3. Select a OS version
4. Table cells (text and radio buttons) should be vertically aligned to the middle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Enhancements:
- Wrap PolicyTypeTable in a container that applies middle vertical alignment to table body cells and check/radio column.

### DIFF
--- a/src/PresentationalComponents/PolicyTypesTable/PolicyTypeTable.js
+++ b/src/PresentationalComponents/PolicyTypesTable/PolicyTypeTable.js
@@ -59,6 +59,7 @@ const PolicyTypeTable = ({
         emptyRows: emptyRows('policy types', columns.length),
       }}
       variant="compact"
+      tableBodyProps={{ style: { verticalAlign: 'middle' } }}
     />
   );
 };


bastilian:
@opacut In general we should avoid inline style tags, this can mess very badly with other things. We should also not attempt to bodge-fix this and first figure out why there is a misalignment in the TableToolsTable and fix it there, so others benefit from it as well.

opacut:
Decided to solve instead with a pr to patternfly-react: https://github.com/patternfly/patternfly-react/pull/12451

PR #2864, opened 2026-06-17
fix(RHINENG-24133): improve save button disabled state visibility

opacut:
When the compliance threshold field has a validation error, the save button (checkmark icon) is functionally disabled but doesn't look visually disabled. This causes confusion as users try to click it but nothing happens.

How to test:
1. set up local environment according to readme in https://github.com/RedHatInsights/compliance-frontend
2. navigate to Policy details page
3. click the edit (pencil) icon on the Compliance threshold field
4. enter an invalid value (e.g., "0999" or any text)
5. the save button (checkmark icon) should now visibly appear disabled with reduced opacity

Closes https://redhat.atlassian.net/browse/RHINENG-24133

Before:
_Light_:
image
image

_Dark_:
image
image

After:
_Light_:
image
image

_Dark_:
image
image

## Summary by Claude
Bug Fixes:
- Make the disabled state of the save button more visually obvious by using opacity instead of subtle color changes when the compliance threshold field has validation errors.

## Summary by Sourcery

Bug Fixes:
- Make the inline save button visibly appear disabled via reduced opacity when the compliance threshold has an invalid value.

### DIFF
--- a/src/SmartComponents/EditPolicyDetails/EditPolicyDetailsInline.js
+++ b/src/SmartComponents/EditPolicyDetails/EditPolicyDetailsInline.js
@@ -146,9 +146,7 @@ const EditPolicyDetailsInline = ({
                   onClick={() => onSave()}
                   style={{
                     marginLeft: '5px',
-                    color: validThreshold
-                      ? 'var(--pf-t--global--background--color--primary--default)'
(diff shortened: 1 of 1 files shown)


LightOfHeaven1994 (approved):
@opacut lgtm! thank you!

PR #2898, opened 2026-07-21
fix(SystemsTable): fix sorting & cleanup

LightOfHeaven1994:
### What PR does:

1. Correct sorting on other system tables
2. Cleanup system columns
3. Unify column usage
4. `OS` column consistency, rename `Operating systems` column to `OS` so views will have the same column name
5. Remove dead code
6. Fix column indexing on tables that do not have select (Policy details -> systems tab, Report details ->  never reported systems)

### How to test:
**Views to check:** Compliance->Systems, Policy details -> systems tab, EditPolicy -> Systems tab, Report details -> Reporting/Never reported tabs
1. Check every system table in the app and ensure that on initial navigation sorting is set to default `name: asc`
2. Check if other column sorting works

## Secure Coding Practices Checklist GitHub Link
- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

## Summary by Sourcery

Unify systems table column definitions and selection behavior while fixing default sorting and checkbox handling across compliance views.

Bug Fixes:
- Ensure Inventory systems tables default to sorting by name ascending on initial load.
- Correct checkbox visibility so selection is only enabled when bulk selection or row selection is configured.
- Fix column setup for report details and policy systems views to prevent misaligned indexing and incorrect sorting.

Enhancements:
- Refactor systems column definitions to use consistent plain object configs and shared OS/workspace/tag columns across views.
- Rename and standardize time-related inventory column to a Last seen variant used consistently in systems tables.
- Remove unused OS helper utilities and associated tests to simplify the helpers module and report details constants.
- Simplify PolicySystemsTab props by dropping unused compliance threshold wiring.
- Adjust tests to cover updated column configuration, Inventory sort behavior, and selection state on systems tables.

PR #2918, opened 2026-08-03
fix(SystemsTable): correct table loading

LightOfHeaven1994:
Few improvements to systems table (follow up for https://github.com/RedHatInsights/compliance-frontend/pull/2898):

1. Avoid calling dispatch `RESET PAGE` on initial load since inventory table takes care of filters initially
2. Only refresh Inventory when filters actually change
3. Seed default sortBy into Redux via INVENTORY_INIT so the first Inventory fetch already uses the correct sort.

### How to test:

1. Navigate to any page that has systems table (Compliance->Systems, for example)
2. Check if table loads well initially
3. Open network tab and navigate to any other page
4. Now navigate back to systems page and check you see only 1 api requests for systems endpoint

**Before** any system table would send 2 /systems API requests + table will show loading state twice

**After** only 1 systems API request sent and table loads only once. 

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Optimize systems table loading to prevent redundant refreshes and align initial inventory state with the configured sort and pagination.

Bug Fixes:
- Prevent the systems table from triggering an extra reset and duplicate loading on initial render.
- Refresh the inventory table only when its filters actually change to avoid unnecessary API calls.

Enhancements:
- Initialize the inventory store with both per-page and sort configuration so the first data fetch uses the correct sort order.

### DIFF
--- a/src/SmartComponents/SystemsTable/SystemsTable.js
+++ b/src/SmartComponents/SystemsTable/SystemsTable.js
@@ -126,7 +126,7 @@ export const SystemsTable = ({
             tags: false,
             hostGroupFilter: !showGroupsFilter,
           }}
-          onLoad={defaultOnLoad(columns, { perPage })}
+          onLoad={defaultOnLoad(columns, { perPage, sortBy })}
           sortBy={sortBy}
           hasCheckbox={selectionEnabled}
           tableProps={{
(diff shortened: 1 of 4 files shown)


hugohezel (approved):
LGTM, works as expected - I can see only one API call.

